PR #1, opened 2019-12-06
refactor the project structure

ts-nitish-a-jain:
This PR only aims at refactoring the project structure to create base for foundation work in the subsequent PRs (which will target the code infrastructure)


build.gradle
```diff
@@ -46,12 +46,12 @@ plugins {
 
 allprojects {
   repositories {
-    artifactoryRelease()
-    artifactorySnapshot()
-    mobileLibs()
+//    artifactoryRelease()
+//    artifactorySnapshot()
```
corycaywood: We can remove this and `artifactoryRelease`

build.gradle
```diff
@@ -46,12 +46,12 @@ plugins {
...
+//    jitpack()
     google()
     jcenter()
-    jitpack()
+//    mobileLibs()
```
corycaywood: This one can be removed too.

build.gradle
ts-nitish-a-jain: Got it. Thanks. Will remove in the upcoming one.

build.gradle
ts-nitish-a-jain: Got it. Thanks. Will remove in the upcoming one.

PR #2, opened 2019-12-11
add code infrastructure to the project

ts-nitish-a-jain:
This PR include changes to create the code infrastructure (static code analysis, documentation, CI, setup for releasing versions).

the publishing part and fixing detekt reported issues is to follow in the next PR

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,15 +23,14 @@ buildscript {
   apply from: 'config/index.gradle'
 
   dependencies {
-    repositories {
-//    jcenter()
-//    jitpack()
-//    gradle()
-      google()
-    }
-    classpath 'com.android.tools.build:gradle:3.5.2'
+    classpath 'com.android.tools.build:gradle:3.5.3'
     classpath "digital.wup:android-maven-publish:3.6.2"
     classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+  }
+
+  repositories {
+    google()
+    jcenter()
   }
 }
 
@@ -42,12 +41,8 @@ plugins {
 
 allprojects {
   repositories {
-//    artifactoryRelease()
-//    artifactorySnapshot()
-//    jitpack()
     google()
     jcenter()
-//    mobileLibs()
   }
 
   gradle.projectsEvaluated {
--- a/build.gradle
+++ b/build.gradle
@@ -35,8 +35,8 @@ buildscript {
 }
 
 plugins {
+  id 'io.gitlab.arturbosch.detekt' version '1.1.1'
   id 'net.ltgt.errorprone' version '0.7'
-  id 'org.jmailen.kotlinter' version "2.1.1"
 }
 
 allprojects {
@@ -51,22 +51,6 @@ allprojects {
       options.compilerArgs << '-Xlint:unchecked' << '-Xlint:deprecation'
       options.compilerArgs << "-Xmaxerrs" << "500"
     }
-  }
-}
-
-subprojects {
-  // Kotlin linter tool. Source code in '/src/java' can take advantage of gradle task:
-  // formatKotlin, but '/src/kotlin' can't.
-  apply plugin: 'org.jmailen.kotlinter'
-
-  kotlinter {
-    ignoreFailures = false
-    indentSize = 2
-    continuationIndentSize = 2
-    reporters = ['checkstyle', 'plain']
-    experimentalRules = false
-    disabledRules = []
-    fileBatchSize = 30
   }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -9,6 +9,7 @@ buildscript {
   ext.androidx_test_ext = '1.1.2-alpha02'
   ext.androidx_testRunner = '1.3.0-alpha02'
   ext.androidx_work = '2.3.0-alpha01'
+  ext.dokka_version = '0.10.0'
   ext.dagger = '2.24'
   ext.jsr250 = '1.0'
   ext.junit = '4.12'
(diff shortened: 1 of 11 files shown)


build.gradle
```diff
@@ -90,3 +68,5 @@ tasks.withType(JavaCompile) {
   options.fork = true
   options.incremental = true
 }
+
+project.version = scmVersion.version
```
corycaywood: Is this working? I don't see the axion-release plugin anywhere. The plugin allows you to derive the version name from git tags. See here: https://github.com/rakutentech/android-remote-config/blob/master/build.gradle#L36

gradle.properties
```diff
@@ -15,18 +15,15 @@ org.gradle.parallel.threads=8
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 org.gradle.parallel=true
-org.gradle.java.home=/Applications/Android Studio.app/Contents/jre/jdk/Contents/Home
 # ---------------------------------- MiniApp SDK flags ------------------------------------
 # Artifactory information needed.
 group=com.rakuten.tech.mobile.miniapp
 name=miniappplatform
```
corycaywood: I don't think you need to define `name`. If you leave it out then `name` will just get set to the module name (`miniapp`).

gradle.properties
```diff
@@ -15,18 +15,15 @@ org.gradle.parallel.threads=8
...
-#"-" indicates aar file will be uploaded to snapshot.
-version=0.0.1-SNAPSHOT
-versionCode=1
 url=https://rakutentech.github.io/android-miniapps
 # MiniApp SDK info for Android.
 application_id=com.rakuten.mobile.miniappsplatform
 description=Mini App platform can download mini apps and display them.
+MINI_APP_BASE_URL=www.example.com
```
corycaywood: Maybe the default should be defined inline in `build.gradle`? But I guess it doesn't really matter for now because we need to change this so that the base url is defined by the App which is implementing the SDK.

miniapp/build.gradle
```diff
@@ -96,12 +71,23 @@ dependencies {
...
+
 //publishing {
 //  publications {
 //    miniapps(MavenPublication, androidArtifact(artifactId: 'miniapp-platform'))
 //  }
-//}
\ No newline at end of file
+//}
+// Android plugin - For publishing to artifactory.
+//apply from: "$CONFIG.internalConfigDir/publish/android.gradle"
```
corycaywood: You can remove this one. The publishing setup can come from the normal config submodule, like this: https://github.com/rakutentech/android-sdkutils/blob/master/sdk-utils/build.gradle#L49

gradle.properties
```diff
@@ -15,18 +15,15 @@ org.gradle.parallel.threads=8
...
 # ---------------------------------- MiniApp SDK flags ------------------------------------
 # Artifactory information needed.
 group=com.rakuten.tech.mobile.miniapp
 name=miniappplatform
-#"-" indicates aar file will be uploaded to snapshot.
-version=0.0.1-SNAPSHOT
-versionCode=1
 url=https://rakutentech.github.io/android-miniapps
 # MiniApp SDK info for Android.
 application_id=com.rakuten.mobile.miniappsplatform
```
corycaywood: Do you know why `application_id` is being set here? I think we don't need it.

gradle.properties
```diff
@@ -15,18 +15,15 @@ org.gradle.parallel.threads=8
 # This option should only be used with decoupled projects. More details, visit
```
corycaywood: Because this is being published as open source, you should also define `licenseName`, `licenseUrl`, and `scmUrl` (it's used when publishing to Bintray). See here: https://github.com/rakutentech/android-remote-config/blob/master/gradle.properties#L4

build.gradle
ts-nitish-a-jain: Yup, it displays a nice `0.1.0-SNAPSHOT` at this time, the addition is at `build.gradle:42`.

gradle.properties
ts-nitish-a-jain: Right, this will be out from here once we start doing the code architecture refactoring

gradle.properties
ts-nitish-a-jain: Thanks for sharing. It has been added now.

gradle.properties
ts-nitish-a-jain: The publishing part is to follow in the next PR. I didn't touch much there yet.
By the way, I was thinking to define a `rootProject.name` in settings.gradle instead.

miniapp/build.gradle
ts-nitish-a-jain: The publishing part is to follow in the next PR. I didn't touch much there yet. 
I have made a note of it. Thanks.

gradle.properties
ts-nitish-a-jain: It's something coming over from the previous structure. It is not defined in the `build.gradle` and the previous intent seems to be setting it up from the gradle.properties instead.
I also think, we can remove this going forward, along with the changes for publishing.

build.gradle
corycaywood: Ah, okay I see. But the it still needs to have the configuration for the plugin so that it uses tags prefixed by `v` i.e. `v1.0.0`.

gradle.properties
corycaywood: That's probably okay. I don't think `rootProject.name` is used for anything in the publishing config.

build.gradle
ts-nitish-a-jain: Noted. Thanks.

PR #5, opened 2019-12-13
fix detekt reported issues

ts-nitish-a-jain:


miniapp/detekt-baseline.xml
```diff
@@ -3,26 +3,15 @@
   <Blacklist></Blacklist>
   <Whitelist>
     <ID>ComplexMethod:MiniAppFileWriter.kt$MiniAppFileWriter$ fun writeResponseBodyToDisk(response: ResponseBody, appId: String, fileUrl: String)</ID>
```
corycaywood: Maybe it's better to move some of these suppressions inline in each file? Then it's easier to see where there are issues being suppressed and remember to fix them. i.e. you could add `@Suppress("ComplexMethod")` to the `MiniAppFileWriter#writeResponseBodyToDisk` method.

Probably the suppressions related to comments can be left here, but the ones related to code should be moved inline.

PR #6, opened 2019-12-16
add bintray publishing support

ts-nitish-a-jain:


gradle/bintray.gradle
```diff
@@ -0,0 +1,35 @@
+apply plugin: 'com.jfrog.bintray'
```
corycaywood: This file is already in the shared build config, so you should be able to use `../config/publish/bintray.gradle` instead.

ts-nitish-a-jain: Thanks for sharing. I will update it out and rebase.

PR #14, opened 2019-12-21
refactor code architecture

ts-nitish-a-jain:
This PR focuses on:
- migrating the previous design and implementation to legacy
- creating a skeletal structure based out of architecture discussion
- creating dependency chains between the SDK components
- basic implementation for auto-initialization of the SDK
- few updates to build.gradle which were otherwise causing failure to `check`


miniapp/src/main/java/com/rakuten/mobile/miniapp/MiniappSdkInitializer.kt
```diff
@@ -0,0 +1,69 @@
...
+        /**
+         * Host app version for the mini app backend.
+         **/
+        @MetaData(key = "com.rakuten.mobile.miniapp.HostAppVersion")
+        fun hostAppVersion(): String
+
+        /**
+         * App Id assigned to host App.
+         **/
+        @MetaData(key = "com.rakuten.mobile.ras.AppId")
```
corycaywood: The key for App ID has to be `com.rakuten.tech.mobile.ras.AppId` and the key for Subscription Key has to be `com.rakuten.tech.mobile.ras.ProjectSubscriptionKey` so that they are the same as our other SDKs.

Actually, the package name for this SDK should be `com.rakuten.tech.mobile` and not `com.rakuten.mobile` so maybe this is a good chance to change it. The package for Bintray is already `com.rakuten.tech.mobile.miniapp` so it should be okay to change.

- [ ] Change manifest key values to `com.rakuten.tech.mobile`
- [ ] Change package name to `com.rakuten.tech.mobile.miniapp` and sample app to ` `com.rakuten.tech.mobile.miniapp.testapp`

miniapp/src/main/java/com/rakuten/mobile/miniapp/QrCodeParser.kt
```diff
@@ -0,0 +1,3 @@
+package com.rakuten.mobile.miniapp
+
+internal class QrCodeParser
```
corycaywood: I'm not really sure if we should add all of these empty classes. It can be kind of confusing seeing all of these empty implementations. Maybe we should just keep the ones that we'll be using during this first development cycle? i.e. miniapp, api, display, storage, etc.

- [ ] Remove classes that won't be used yet

testapp/build.gradle
```diff
@@ -9,6 +9,13 @@ android {
         versionCode versionCode
         versionName versionName
 
+        manifestPlaceholders = [
+                baseUrl: property("MINIAPP_SERVER_BASE_URL") ?: "https://www.example.com/",
```
corycaywood: Seems this implementation isn't working on CI because Gradle throws an exception when the property doesn't exist. Need to check if the property exists first.

- [ ] Check if gradle property exists before using

miniapp/src/main/java/com/rakuten/mobile/miniapp/RealMiniApp.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    override suspend fun create(
+        appId: String,
+        success: (MiniAppView) -> Unit,
+        error: (Exception) -> Unit
+    ) {
+        TODO("not implemented")
+    }
+
+    override suspend fun createFromQrCode(
```
corycaywood: Similar comment on this one: maybe we shouldn't put this in yet because we still won't be implementing this for a while.

- [ ] Remove `createFromQrCode` api

miniapp/src/main/java/com/rakuten/mobile/miniapp/miniapp/MiniAppDownloader.kt
```diff
@@ -0,0 +1,11 @@
+package com.rakuten.mobile.miniapp.miniapp
+
+import com.rakuten.mobile.miniapp.api.ApiClient
+import com.rakuten.mobile.miniapp.storage.MiniAppStorage
+
+@SuppressWarnings("UseDataClass")
+internal class MiniAppDownloader(
```
corycaywood: I was thinking maybe this can just be named `Downloader`? It's already in a package named `miniapp` so the name `MiniAppDownloader` seems redundant. And same with the other classes in this package. What do you think?

- [ ] Consider renaming MiniAppDownloader, MiniAppLister, and MiniAppStatus to Downloader, Lister, Status

miniapp/src/main/java/com/rakuten/mobile/miniapp/display/Displayer.kt
```diff
@@ -0,0 +1,4 @@
+package com.rakuten.mobile.miniapp.display
+
+@SuppressWarnings("UseDataClass")
+internal class Displayer(val realMiniAppView: RealMiniAppView)
```
corycaywood: I'm not sure if this should be passed in the constructor? At the very least, it should be passing the interface, not the implementation.

- [ ] Fix Displayer constructor

miniapp/src/main/java/com/rakuten/mobile/miniapp/QrCodeParser.kt
ts-nitish-a-jain: I think only removing the public API method would have sufficed as it would somewhat reflect the architecture in place and yet not exposing any public API to consumers. 
Nevertheless, It's still okay, as we spoke.

miniapp/src/main/java/com/rakuten/mobile/miniapp/miniapp/MiniAppDownloader.kt
ts-nitish-a-jain: It is a tricky one actually. In Java, it used to be a single public class per file. In Kotlin, we can have multiple of those in one file. File/Class names used to be descriptive and contextual of what they tend to do. Package names in the namespace augmented to provide the context for a file's purpose and grouping.

For faster intuitivity, I think it is better to have a file/class that depicts its purpose by name and package's information context as a secondary supporting artifact. 

It's difficult to understand the purpose of Lister and Status just by its name (let say who's status or what type of lister). For `MiniAppDownloader`, I thought to rely on the secondary context, `legacy`. Further, It might make use of a separate downloader and update checker class.
I think we can rename the `miniapp` package to `base` or alike. What do you think?

corycaywood: Yeah, makes sense I think. But if we name them `MiniAppDownloader`, etc. then I think they don't need to be in a sub-package. Maybe they could just be in the root package at `com.rakuten.tech.mobile.miniapp.MiniAppDownloader`. These classes will have most of the core business rules, so maybe it makes more sense to be at the root?

PR #16, opened 2020-01-09
add implementation for manifest api

ts-nitish-a-jain:


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ManifestApi.kt
```diff
@@ -1,5 +1,18 @@
...
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Path
+import java.util.UUID
 
-internal class ManifestResponse
+internal interface ManifestApi {
+    @GET("miniapp/{miniappId}/version/{versionId}/manifest")
+    fun fetchFileListFromManifest(
+        @Path("miniappId") miniAppId: UUID,
```
corycaywood: Why use `UUID` instead of just `String`? It seems like using `UUID` will tie us to using a specific `UUID` variant so this will break if the API changes the implementation. Is there any benefit to using it instead of `String`?

ts-nitish-a-jain: The idea to use UUID was to filter out the possible String-based invalid inputs and save the overhead of server-side validation, explicit checks within code to verify the input params. By the way, do we have a spec around the UUID variant?

corycaywood: Nope, we don't have a spec for the UUID variant. The problem is that even if we have a spec for the UUID, the API could still change the format of it at some point and then old versions of the SDK will be broken. So that's why I don't think that validating the UUID should be the responsibility of the SDK.  It would be safer to just use a `String` and assume that the server will validate the UUID.

PR #19, opened 2020-01-17
ci: Deploy Test App to App Center on merge to master branch

corycaywood:
Also changed the versionCode and versionName for the test app because they were not getting set correctly. It might be nice if we could automatically increment the versionCode on CI, but it is not necessary at this time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,15 +49,16 @@ jobs:
       - store_artifacts:
           path: miniapp/build/outputs/aar
           destination: aar/
-      - store_artifacts:
-          path: sample/build/outputs/apk
-          destination: sample/
+      - persist_to_workspace:
+          root: testapp/build/outputs/apk/debug
+          paths:
+            - testapp-debug.apk
       - persist_to_workspace:
           root: ~/code
           paths:
             - miniapp/build/
 
-  publish:
+  publish-sdk:
     docker:
       - image: circleci/android:api-28
     working_directory: ~/code
@@ -66,6 +67,8 @@ jobs:
     steps:
       - checkout
       - run: git submodule update --init
+      - attach_workspace:
+          at: ./
       - run:
           name: Current Version
           command: ./gradlew cV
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -49,15 +49,12 @@ jobs:
...
           destination: aar/
-      - store_artifacts:
-          path: sample/build/outputs/apk
-          destination: sample/
       - persist_to_workspace:
-          root: ~/code
+          root: testapp/build/outputs/apk/debug
           paths:
-            - miniapp/build/
+            - testapp-debug.apk
```
ts-nitish-a-jain: Because the build job is common for the `publish-sdk` workflow, should persisting the AAR of SDK be required?

corycaywood:
> Looks OK.
> `deploy-test-app` workflow can make the sample app go ahead of the SDK's published version. I think it should firstly publish SDK and then the sample app. What do you think?

Ah, good point. I intended for `deploy-test-app` to run after `publish-sdk`, but it will actually run in parallel to `publish-sdk`. 

But I can't seem to find a way to change this. If I make `deploy-test-app` require `publish-sdk`, then it will only deploy the test app when we are publishing the SDK. I think it should deploy the test app any time that commits are merged to master so that we can have a binary for QA. 

Maybe it's okay if they run parallel? Because anytime that we publish the SDK, the test app that's deployed should be the same as the previous one that was deployed.

.circleci/config.yml
corycaywood: It was persisting the `miniapp/build` folder before, but it wasn't working. But maybe it's just because it was missing the `attach_workspace` step in the `publish-sdk` job. I have added `attach_workspace` now and added back in the `persist_to_workspace` for the build folder.

ts-nitish-a-jain:
> > Looks OK.
> > `deploy-test-app` workflow can make the sample app go ahead of the SDK's published version. I think it should firstly publish SDK and then the sample app. What do you think?
> 
> Ah, good point. I intended for `deploy-test-app` to run after `publish-sdk`, but it will actually run in parallel to `publish-sdk`.
> 
> But I can't seem to find a way to change this. If I make `deploy-test-app` require `publish-sdk`, then it will only deploy the test app when we are publishing the SDK. I think it should deploy the test app any time that commits are merged to master so that we can have a binary for QA.
> 
> Maybe it's okay if they run parallel? Because anytime that we publish the SDK, the test app that's deployed should be the same as the previous one that was deployed.

Let's revisit this later then. I think it will do the job for now.

PR #21, opened 2020-01-20
Update the exception facing the host app and handle network request T…

ts-nitish-a-jain:
…ype T scenario

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
@@ -2,6 +2,8 @@ package com.rakuten.tech.mobile.miniapp.api
 
 import androidx.annotation.VisibleForTesting
 import com.rakuten.tech.mobile.miniapp.MiniAppInfo
+import com.rakuten.tech.mobile.miniapp.MiniAppSdkException
+import okhttp3.ResponseBody
 import retrofit2.Call
 import retrofit2.HttpException
 import retrofit2.Response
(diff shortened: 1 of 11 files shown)


miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/api/ApiClientSpec.kt
```diff
@@ -126,14 +127,28 @@ open class RetrofitRequestExecutorNormalSpec : RetrofitRequestExecutorSpec() {
...
+    @Test
+    fun `should throw exception when the response returned by server is not of type T`() =
+        runBlockingTest {
+            mockServer.enqueue(createInvalidTestApiResponse())
+            try {
+                createRequestExecutor()
+                    .executeRequest(createApi().fetch())
+
+                TestCase.fail("Should have thrown MiniAppSdkException.")
+            } catch (exception: MiniAppSdkException) {
```
corycaywood: You should be able to use the `expected` parameter instead of try/catch for this test case: `@Test(expected = MiniAppSdkException::class)`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -50,18 +52,31 @@ internal class RetrofitRequestExecutor(
...
-            return response.body()!! // Body can't be null if request was successful
-        } else {
-            val error = response.errorBody()!! // Error body can't be null if request wasn't successful
-            throw MiniAppHttpException(
+    private fun sdkExceptionFromHttpException(
+        response: Response<in Nothing>,
+        error: ResponseBody
+    ): MiniAppSdkException {
+        return MiniAppSdkException(
+            MiniAppHttpException(
```
corycaywood: I forgot about this: `MiniAppHttpException` should be `internal` now.

PR #25, opened 2020-02-03
base implementation for displayer module

ts-nitish-a-jain:


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppView.kt
```diff
@@ -1,7 +1,42 @@
...
+    override suspend fun obtainView(context: Context): WebView =
+        WebView(context).apply {
+            setLayoutParams(
+                FrameLayout.LayoutParams(
+                    ViewGroup.LayoutParams.MATCH_PARENT,
+                    ViewGroup.LayoutParams.MATCH_PARENT
+                )
+            )
+            settings.javaScriptEnabled = true
+            setWebViewClient(object : WebViewClient() {
```
corycaywood: An anonymous class is okay for now, but probably we'll want to move it out into a separate class when there's more functionality.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppView.kt
```diff
@@ -1,7 +1,42 @@
...
+                override fun shouldOverrideUrlLoading(view: WebView, url: String): Boolean {
+                    view.loadUrl(url)
+                    return true
+                }
+            })
+            loadUrl("$basePath/index.html")
+        }
+
+    override fun destroyView(webView: WebView) {
```
corycaywood: I wonder if there's a way we can implement this so the host app doesn't have to call `obtainView` and `destroyView`. We can think about that later.

PR #28, opened 2020-02-04
Poc

minh-rakuten:
Exemplar of displaying MiniApp with SDK

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,3 +21,8 @@ kotlin.code.style=official
 licenseName="MIT License"
 licenseUrl=https://opensource.org/licenses/MIT
 scmUrl=https://github.com/rakutentech/android-miniapp.git
+
+MINIAPP_SERVER_BASE_URL=https://miniapp-platform-japaneast-dev.azurewebsites.net/api/v1/
+HOST_APP_VERSION=1.0.0
+HOST_APP_ID=1
+HOST_APP_SUBSCRIPTION_KEY=1
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,9 +21,3 @@ kotlin.code.style=official
 licenseName="MIT License"
 licenseUrl=https://opensource.org/licenses/MIT
 scmUrl=https://github.com/rakutentech/android-miniapp.git
-
-#temp for POC Demo
-MINIAPP_SERVER_BASE_URL=https://miniapp-platform-japaneast-dev.azurewebsites.net/api/v1/
-HOST_APP_VERSION=1.0.0
-HOST_APP_ID=1
-HOST_APP_SUBSCRIPTION_KEY=1
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
@@ -7,12 +7,14 @@ import java.io.InputStream
 
 internal class FileWriter {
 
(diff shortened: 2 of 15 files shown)


gradle.properties
```diff
@@ -21,3 +21,9 @@ kotlin.code.style=official
 licenseName="MIT License"
 licenseUrl=https://opensource.org/licenses/MIT
 scmUrl=https://github.com/rakutentech/android-miniapp.git
+
+#temp for POC Demo
+MINIAPP_SERVER_BASE_URL=https://miniapp-platform-japaneast-dev.azurewebsites.net/api/v1/
```
corycaywood: Values for the base url, app version, id, and subscription key should not be committed, so please remove this section of `gradle.properties`.

We can set these values as environment variables on CircleCI. If you need it for testing locally, you can set it in your global gradle properites at `~/.gradle/gradle.properties`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
```diff
@@ -7,16 +7,14 @@ import java.io.InputStream
...
-    ) {
+    suspend fun write(inputStream: InputStream, path: String) {
         withContext(Dispatchers.IO) {
-            File(parentPath).mkdirs()
-            val file = File(parentPath, fileName)
+            val file = File(path)
+            file.parentFile?.mkdirs()
             inputStream.use { it.copyTo(file.outputStream()) }
-            file.outputStream().close()
+            // hotfix for the save issue
```
corycaywood: Let's just remove these lines completely - we will still have the git history.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
```diff
@@ -19,13 +19,14 @@ internal class MiniAppStorage(
...
     }
 
-    fun getParentPath(
+    fun getAbsoluteWritePath(
         basePath: String,
-        filePath: String
-    ) = "$basePath$filePath"
+        filePath: String,
+        fileName: String
+    ) = "${basePath.removeSuffix("/")}$filePath$fileName"
```
corycaywood: This works for now, but in the future I think we will need to make sure that the `/` isn't removed if it needs to be there - i.e. if `filePath` doesn't start with a `/`.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -0,0 +1,38 @@
...
+        viewModel = ViewModelProviders.of(this)
+            .get(MiniAppDwnlViewModel::class.java).apply {
+                miniAppView.observe(this@MiniAppDisplayActivity, Observer {
+                    //action: display webview
+                    setContentView(it)
+                })
+            }
+
+        val miniAppInfo = MiniAppInfo(
+            id = "c028be14-4ded-4734-acc6-2d35d9a67630",
```
corycaywood: We can do this in another PR, but we need to change `MiniApp#create` to accept `id` and `versionId` instead of `MiniAppInfo` so they don't need to construct a `MiniAppinfo`.

ts-nitish-a-jain:
@corycaywood @minh-rakuten As we talked about it, these changes will be targeted to POC branch. Thereafter, we will reinstate the required things to master. After updating the BE API implementation in the master, the public interface will also receive updates.

PR #30, opened 2020-02-05
Migrate useful stuff from poc branch

ts-nitish-a-jain:
- remove invocation to obtain view in the sample app
- update the path scheme for file saving and fix file saving issue
- update the WebView loading based on the updated scheme
- update the public API interface

Please note that this PR is mostly oriented on migrating the implementation.

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
@@ -19,7 +19,9 @@ abstract class MiniApp internal constructor() {
     abstract suspend fun listMiniApp(): List<MiniAppInfo>
 
     /**
-     * Creates a mini app from the metadata [MiniAppInfo] object.
+     * Creates a mini app.
+     * @param appId application ID of the mini app
+     * @param versionId a version ID associated with the mini app
      * The mini app is downloaded, saved and provides a [MiniAppView] when successful
      * @throws MiniAppSdkException when there is some issue during fetching,
      * downloading or creating the view.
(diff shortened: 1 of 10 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
```diff
@@ -7,16 +7,11 @@ import java.io.InputStream
...
-        parentPath: String,
-        fileName: String
-    ) {
+    suspend fun write(inputStream: InputStream, path: String) {
         withContext(Dispatchers.IO) {
-            File(parentPath).mkdirs()
-            val file = File(parentPath, fileName)
+            val file = File(path)
+            file.parentFile?.mkdirs()
             inputStream.use { it.copyTo(file.outputStream()) }
```
minh-rakuten: If `file.parentFile` is null then the copy execution will cause crash. No further execution when the file path is invalid.
The implementation might be like this
```suggestion
            if (file.parentFile != null) {
                file.parentFile.mkdirs()
                inputStream.use { it.copyTo(file.outputStream()) }
            }
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -25,7 +25,8 @@ abstract class MiniApp internal constructor() {
      * downloading or creating the view.
      */
     abstract suspend fun create(
-        miniAppInfo: MiniAppInfo
```
corycaywood: Need to update the KDoc comment.

PR #31, opened 2020-02-06
Minh/master

minh-rakuten:
Sample mini app view.
MINI-178:

When a Mini App is tapped in the Mini App listing, a new Activity should be launched which will download and display the Mini App using `MiniApp#create`. 

A loading icon can be displayed in the Activity while the Mini App is downloading.

After the success callback is invoked, the provided WebView should be displayed in the Activity.

Display the message error.


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppInfo.kt
```diff
@@ -25,6 +25,6 @@ data class MiniAppInfo(
      */
     val icon: String,
 
-    internal val versionId: String,
```
ts-nitish-a-jain: This is okay for now. We would want to change it back once we update the backend API sepcs.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppView.kt
```diff
@@ -24,6 +24,10 @@ internal class RealMiniAppView(
                 )
             )
             settings.javaScriptEnabled = true
+            settings.allowFileAccess = true
```
ts-nitish-a-jain: I think this should also be fine, considering that it would just enhance the experience and save us from some debugging sessions back and forth when we receive new mini apps for testing.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/adapter/MiniAppListAdapter.kt
```diff
@@ -14,7 +14,7 @@ import com.rakuten.tech.mobile.miniapp.MiniAppInfo
 import com.rakuten.tech.mobile.miniapp.testapp.databinding.ItemListMiniappBinding
 
 
-class MiniAppListAdapter(var miniapps: List<MiniAppInfo>) :
+class MiniAppListAdapter(var miniapps: List<MiniAppInfo>, val iMiniAppList: IMiniAppList) :
```
ts-nitish-a-jain: That looks like Hungarian Notation. I think in the current code base we don't have much of it. I think it is better to make all code look the same style.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/miniapplist/MiniAppListViewModel.kt
```diff
@@ -17,7 +15,7 @@ class MiniAppListViewModel constructor(
     constructor() : this(MiniApp.instance())
 
     private val _miniAppListData =
-        MutableLiveData<List<MiniAppInfo>>().apply { value = emptyList() }
+        MutableLiveData<List<MiniAppInfo>>()
```
ts-nitish-a-jain: I think it is better to keep it initialized with an emptyList() because the object has been created already. When it is an empty list, the adapter wouldn't do much for the list.

testapp/src/main/res/values/strings.xml
```diff
@@ -1,5 +1,5 @@
 <resources>
-    <string name="app_name">Mini App Test</string>
+    <string name="app_name">Mini App Sample App</string>
```
ts-nitish-a-jain: Maybe just Sample App seems better. @corycaywood What do you think?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/legacy/MainActivity.kt
```diff
@@ -49,7 +49,7 @@ class MainActivity : AppCompatActivity() {
     }
 
     companion object {
-        const val APP_ID = "b9247b09-c534-48c5-bcc5-0644907d22f8"
-        const val VERSION_ID = "9761a1dc-a1cb-4730-92ac-e9aa09a87531"
+        const val APP_ID = "d005410f-4055-4d95-a0c9-899a60c5936a"
```
ts-nitish-a-jain: This is in legacy code. This should not be referred to and shouldn't require any modification.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppInfo.kt
corycaywood: Yeah, the API spec has changed it bit. There is now a `versionId` and `versionTag`. `versionTag` should be public, but `versionId` should be private.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppView.kt
corycaywood: I think we will need to revisit this at some point. I think there could be some security concerns here. i.e. the Mini App WebView will be able to access anything in the App's `files` directory (including other Mini App data or other data that the host App has saved in `files`). We can revisit this when we address separation of data.

testapp/src/main/res/values/strings.xml
corycaywood: I think "Mini App Sample App" is okay, or maybe just "Mini App Sample"? "Mini App Demo"? Something like that...

minh-rakuten: Mini App Sample sounds okay too.

minh-rakuten:
> LGTM 👍 Be sure to squash when merging.
> 
> By the way, @minh-rakuten: Usually we push branches to our forked repos instead of the main repo and create the PR from the fork. It just helps to keep the main repo cleaner.

@corycaywood Yeah 👍. I'm doing it from my own fork.

PR #32, opened 2020-02-06
remove legacy code from the sample application

ts-nitish-a-jain:


ts-nitish-a-jain:
> Nice 👍 I like PRs that only have deletions 😆

I am all into the next one in SDK 🤤

PR #33, opened 2020-02-06
fix: Change init order of SdkUtils

corycaywood:
It needs to be initialized before the Mini App SDK

Cherry-picking this fix from the `poc` branch.


corycaywood:
> Just want to comment that document of [android-sdkutils](https://github.com/rakutentech/android-sdkutils) should have it 😆

Yep, good point 👍

PR #39, opened 2020-02-13
refactor display implementation

ts-nitish-a-jain:
Refactored display module's implementation and added unit tests.

IF RealMiniAppDisplay becomes bigger, we would like to revisit the separation of UI and the actions on the UI 

# Checklist


miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -0,0 +1,70 @@
...
+
+    @Before
+    fun setup() {
+        context = getApplicationContext()
+    }
+
+    @Test
+    fun `for a given basePath, obtainView returns corresponding view to the caller`() =
+        runBlockingTest {
+            val display = Displayer(context).createMiniAppDisplay(basePath = TEST_BASE_PATH)
```
corycaywood: This should be constructed manually instead of via `Displayer`. i.e. `RealMiniAppDisplay(context, basePath)`

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -0,0 +1,70 @@
...
+            miniAppView shouldNotHaveTheSameClassAs WebView(context)
+            if (miniAppView is WebView) {
+                miniAppView.url shouldContain TEST_BASE_PATH
+            }
+        }
+
+    @Test
+    fun `for a given basePath, obtainView should not return WebView to the caller`() =
+        runBlockingTest {
+            val display = Displayer(context).createMiniAppDisplay(basePath = TEST_BASE_PATH)
```
corycaywood: Same here as above: this should not use `Displayer` for construction.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -0,0 +1,45 @@
...
+        webViewClient = MiniAppWebViewClient()
+        loadUrl(getLoadUrl())
+    }
+
+    override fun setWebViewClient(client: WebViewClient?) {
+        super.setWebViewClient(client ?: MiniAppWebViewClient())
+    }
+
+    @VisibleForTesting
+    fun getLoadUrl() = "file://$basePath/index.html"
```
corycaywood: Looks like this function could be internal. 

Or can't we test the load url by directly calling `WebView#url`? And then this function can be private.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -0,0 +1,70 @@
...
+            val miniAppView = display.getMiniAppView()
+            miniAppView shouldNotHaveTheSameClassAs WebView::class
+        }
+
+    @Test
+    fun `when getLoadUrl is called then a formed path is returned from basePath for loading`() {
+        val miniAppWindow = RealMiniAppDisplay(context, basePath = TEST_BASE_PATH)
+        miniAppWindow.getLoadUrl() shouldBeEqualTo "file://dummy/index.html"
```
corycaywood: As I mentioned, I think we should test this by calling `WebView#url` instead of exposing `getLoadUrl`. Checking this exposed method doesn't guarantee that the webview is actually using that method.

I know that we are supposed to be returning a generic `View` and not a `WebView`, but I think it's okay to treat it as a `WebView` in the tests for `RealMiniAppDisplay`.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        val miniAppWindow = RealMiniAppDisplay(context, basePath = TEST_BASE_PATH)
+        miniAppWindow.layoutParams.width shouldEqualTo ViewGroup.LayoutParams.MATCH_PARENT
+        miniAppWindow.layoutParams.height shouldEqualTo ViewGroup.LayoutParams.MATCH_PARENT
+    }
+
+    @Test
+    fun `when MiniAppDisplay is created then required websettings should be enabled`() {
+        val miniAppWindow = RealMiniAppDisplay(context, basePath = TEST_BASE_PATH)
+        assertTrue { miniAppWindow.settings.javaScriptEnabled }
+        assertTrue { miniAppWindow.settings.allowUniversalAccessFromFileURLs }
```
corycaywood: Should we separate these into two tests? Something like `javascript should be enabled` and `should allow universal access from file urls`.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Test
+    fun `when MiniAppDisplay is created then then WebViewClient is set to MiniAppWebViewClient`() {
```
corycaywood: Typo in the description: two `then`'s

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    override fun getMiniAppView(): View = this
+}
+
+internal class MiniAppWebViewClient : WebViewClient() {
```
corycaywood: Do you think this should be private?

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -0,0 +1,70 @@
...
+        val miniAppWindow = RealMiniAppDisplay(context, basePath = TEST_BASE_PATH)
+        (miniAppWindow as WebView).webViewClient shouldBeInstanceOf MiniAppWebViewClient::class
```
corycaywood: Actually it seems like this test isn't really needed. It's not actually testing any functionality. Maybe we should remove this test and add functionality based tests once the WebViewClient actually does something?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -0,0 +1,45 @@
...
+import android.widget.FrameLayout
+import androidx.annotation.VisibleForTesting
+import com.rakuten.tech.mobile.miniapp.MiniAppDisplay
+
+@SuppressLint("SetJavaScriptEnabled")
+internal class RealMiniAppDisplay(context: Context, val basePath: String) : MiniAppDisplay,
+    WebView(context) {
+
+    init {
+        setLayoutParams(
```
corycaywood: Could use `layoutParams = FrameLayout.LayoutParams(`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
ts-nitish-a-jain: Actually, there are two ways to do it.
1) Functional way: Try to break down the units into the smallest pieces. And then, try to test the smaller units.
2) Without the explicit method: The string expression can directly be embedded into loadUrl. This can be unit tested by WebView#url.

Currently, I am trying to follow the first approach. Making it internal is not bringing any benefit as such because the class is already internal and can not be used for typecasting. 💭 🤔 💭

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
ts-nitish-a-jain: This is only attempting to check if the path is correctly formed when the `basePath` is supplied into the constructor.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
ts-nitish-a-jain: Initially, I created two separate ones, but then as both belong to WebSettings, I thought to keep them into one.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
ts-nitish-a-jain: It's attempting to test if the correct implementing class is being used. e.g. in the near future, if there are multiple implementers of the contract, then this part of the code should use the required implementer and not any other.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
ts-nitish-a-jain: One of the unit tests is checking this part. Just added annotation to manage the case.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
ts-nitish-a-jain: Yea, it got a bit mixed up after the second level refactor.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/Displayer.kt
```diff
@@ -1,8 +1,14 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
-internal class Displayer {
+internal class Displayer(val context: Context) {
 
-    suspend fun getViewForMiniApp(basePath: String): MiniAppView = RealMiniAppView(basePath)
+    suspend fun createMiniAppDisplay(basePath: String): MiniAppDisplay =
+        withContext(Dispatchers.Main) {
+            RealMiniAppDisplay(context, basePath)
```
minhlunso1: You may want to have a context null validation here or exception handling.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
corycaywood: Yeah, I agree that things can be tested in a functional way, but it doesn't make sense to do so when the function is exposed only so that it can be tested. In that case, it's the same thing as testing internal implementation details.

My point is that the reason for testing this is to make sure the WebView is loading the correct URL, but just testing an exposed `getLoadUrl` function gives us no guarantee that the WebView is actually using that function and loading the correct URL.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
corycaywood: True, but they are actually two different functionalities: the ability to use JavaScript and the ability to access file URLs. Maybe you can create a separate test class in this file for `WebSettings`? So you could have something like:
```
class RealMiniAppDisplayTest {
...
}

class RealMiniAppDisplayWebSettingsTest {
....
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
ts-nitish-a-jain: Thanks for sharing your thoughts.

Actually, the intent is to just test if the URL forming logic is correct or not (which can be an easy source of error otherwise). I had marked it `VisibleForTesting`. 
To be in parity with `RetrofitCreatorUtils`, I have updated the modifiers.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/Displayer.kt
minh-rakuten: You may want to have a context null validation here or exception handling.

ts-nitish-a-jain: Thanks. Done.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
ts-nitish-a-jain: It's a bit tricky contextually. Anyhow, I have separated the tests to move forward.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
corycaywood: You're right, we do need to test that the URL was formed correctly, but I don't it should be tested by exposing a function. The reason we need to test if the URL was formed correctly is to check if the WebView is displaying the correct URL. But if we were to change the call for `WebView#loadUrl` and are no longer calling the `getLoadUrl` function, then the test for `getLoadUrl` would be meaningless.

Actually, I think it would be better if `"file://$basePath/index.html"` were moved inline to `WebView#loadUrl` instead of being returned by a separate function. `getLoadUrl` is only being used by the WebView and it doesn't need to be known by anything else.

Sorry I keep coming back to this issue, but I think it's important that we avoid this kind of pattern where some functionality is exposed only so that it can be used in unit tests. It can make the code more difficult to read and can be a potential source of bugs later on.

ts-nitish-a-jain: Your description has two different contexts, I will discuss this with you directly. Unit testing a block of logic code is different from unit testing its usage by its consumer.

PR #41, opened 2020-02-18
feat: update implementation for new manifest endpoint

ts-nitish-a-jain:
This PR updates the implementation for `manifest` API endpoint. 
Since the structure is somewhat the same, and only the URL parsing scheme has changed, the legacy implementation has been updated.

All changes for updating endpoints implementation are being pushed to a separate branch at this moment (`update-impl-api-endpoints`). The adapted URL scheme lengthens up the URL in comparison to the previous one (as such the receiving end, SDK in this case, only requires content and the intelligence in URL for the directory structure in the current system implementation. This was discussed with BE team yesterday and has been taken up by them to see if there's some quick way to address it). IF they are able to find some quick fix for it, the implementation could again see some minor change.

In refactoring exercise, dependency on legacy parser will be removed (along with code/build/tests etc.) (i.e., the next [ 🤞] step after updating endpoints) 

# Checklist

### DIFF
--- a/miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/legacy/core/BaseTest.kt
+++ b/miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/legacy/core/BaseTest.kt
@@ -20,8 +20,10 @@ open class BaseTest {
         const val INVALID_MANIFEST_ENDPOINT =
             "/78d85043-d04f-486a-8212-bf2601cb63a2/17bccee1-17f0-44fa-8cb8-2da89eb49905/manifest/"
 
-        const val VALID_FILE_URL_PATH = "https://version/78d85043-d04f-486a-8212-bf2601cb63a2/js" +
-                "/index.html"
+        const val VALID_FILE_URL_PATH =
+            "https://miniappsplatformdev.blob.core.windows.net/"
+                .plus("map-published/min-872f9172-804f-44e2-addd-ed612170dac9/")
+                .plus("ver-6181004c-a6aa-4eda-b145-a5ff73fc4ad0/a/b/index.html")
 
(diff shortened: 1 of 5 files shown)


miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/legacy/core/BaseTest.kt
```diff
@@ -20,8 +20,10 @@ open class BaseTest {
         const val INVALID_MANIFEST_ENDPOINT =
             "/78d85043-d04f-486a-8212-bf2601cb63a2/17bccee1-17f0-44fa-8cb8-2da89eb49905/manifest/"
 
-        const val VALID_FILE_URL_PATH = "https://version/78d85043-d04f-486a-8212-bf2601cb63a2/js" +
-                "/index.html"
+        const val VALID_FILE_URL_PATH =
+            "https://miniappsplatformdev.blob.core.windows.net/"
```
corycaywood: This should not be a real API URL. i.e. use `https://www.example.com/` instead.

corycaywood:
> LGTM!
> However, I think we should only merge it when those new api endpoints are actually ready on the backend side.

@minh-rakuten He's merging it onto a new branch `feature/update-impl-api-endpoints`, so I think it's okay to merge there for now. And then we can merge the `feature/update-impl-api-endpoints` branch to master only after the API is ready.

ts-nitish-a-jain:
> Looks good. But I think that `ManifestApi` is supposed to have a `hostVersion` query param added, isn't it?

@corycaywood yes, you're right. I also realized while updating `list` endpoint's implementation. I have updated this part in that PR. Sorry about the mix.

PR #43, opened 2020-02-18
feat: Check already downloaded

minh-rakuten:
# Description
- Local miniapp loading when it's already downloaded
- Bring back destroy flow for MiniAppDisplay
- Temporarily implement destroy call in sample app

## Links
[MINI-366](https://jira.rakuten-it.com/jira/browse/MINI-366)

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSharedPreferences.kt
```diff
@@ -0,0 +1,16 @@
...
+import android.content.Context
+import android.content.SharedPreferences
+
+internal class MiniAppSharedPreferences(context: Context) {
+    private val prefs: SharedPreferences = context.getSharedPreferences(
+        "com.rakuten.tech.mobile.miniapp.storage", Context.MODE_PRIVATE
+    )
+
+    fun setAppExisted(appId: String, versionId: String, value: Boolean) =
+        prefs.edit().putBoolean(appId + versionId, value).apply()
```
corycaywood: In the future, we will also need to know what version was downloaded because we will need to delete the previous version when a new version is downloaded for the specified App ID. So could we cache the versionId by AppId here, i.e. `prefs.edit().putString(appId, versionId)`. 

What do you think? Were you thinking of another way of handling deletion?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
```diff
@@ -42,4 +42,6 @@ internal class MiniAppStorage(
 
     fun getSavePathForApp(appId: String, versionId: String) =
         "${basePath.path}/$SUB_DIR_MINIAPP/$appId/$versionId"
+
+    fun filePathExisted(baseSavePath: String): Boolean = File(baseSavePath).exists()
```
corycaywood: `filePathExisted` -> `filePathExists`

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloaderSpec.kt
```diff
@@ -78,10 +74,71 @@ class MiniAppDownloaderSpec {
         assertFalse { getMockedDownloader().isManifestValid(manifestEntity) }
     }
 
+    @Test
+    fun `when no existing app in local storage, run download execution`() {
+        runBlockingTest {
+            val apiClient: ApiClient = mock()
```
corycaywood: You could setup these mocks at the top of this class and then you don't have to define them in every test.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSharedPreferences.kt
minh-rakuten: I don't think that we need to store the specific appId and versionId here.
The remove of previous version occurs when the download of new version completes -> We know the appId and the new versionId so we can remove all older version folders.

Actually, I have already implemented that feature but not in this PR.

```MiniAppStorage.kt
private fun getParentPathApp(appId: String) = "${basePath.path}/$SUB_DIR_MINIAPP/$appId/"
```
```MiniAppDownloader.kt
fun removeOutdatedVersionApp(appId: String, versionId: String) {
    val parentFile = File(getParentPathApp(appId))

    if (parentFile.isDirectory && parentFile.listFiles() != null) {
        parentFile.listFiles().forEach {file ->
            if (!file.absolutePath.endsWith(versionId))
                file.deleteRecursively()
        }
    }
}
```

corycaywood: Ah, okay nice solution! We just need to make sure that the old version isn't removed until after the new version has completed downloading (in case the new version download fails).

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -14,4 +14,9 @@ interface MiniAppDisplay { // This contract keeps the display of a mini app impl
      * the parent's dimensions.
      */
     fun getMiniAppView(): View
+
+    /**
+     * Destroy necessary view state and any services registered with.
```
ts-nitish-a-jain: It will be a good idea to describe when this should be called.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -3,18 +3,31 @@ package com.rakuten.tech.mobile.miniapp
...
     val storage: MiniAppStorage,
-    val apiClient: ApiClient
+    val apiClient: ApiClient,
+    val miniAppStatus: MiniAppStatus
 ) {
 
+    suspend fun getMiniApp(appId: String, versionId: String): String {
+        if (miniAppStatus.isVersionDownloaded(appId, versionId)) {
+            val baseSavePath = storage.getSavePathForApp(appId, versionId)
+            if (storage.filePathExists(baseSavePath))
```
ts-nitish-a-jain: What is the scenario when version is downloaded but the path doesn't exist?

minh-rakuten: When someone deletes data folder in local storage.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStatusTest.kt
```diff
@@ -0,0 +1,31 @@
...
+        miniAppStatus = MiniAppStatus(ApplicationProvider.getApplicationContext())
+    }
+
+    @Test
+    fun `isAppExisted should be false if no app has been saved`() {
+        miniAppStatus.isVersionDownloaded(TEST_ID_MINIAPP, TEST_ID_MINIAPP_VERSION) shouldBe false
+    }
+
+    @Test
+    fun `setAppExsited and isAppExisted should correspond with each other`() {
```
ts-nitish-a-jain: The description needs to be changed with the latest I believe.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
```diff
@@ -42,4 +42,6 @@ internal class MiniAppStorage(
 
     fun getSavePathForApp(appId: String, versionId: String) =
         "${basePath.path}/$SUB_DIR_MINIAPP/$appId/$versionId"
+
```
ts-nitish-a-jain: Unit tests for this too please

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
ts-nitish-a-jain: I think shared preferences are stored in the same place. Isn't it?

minh-rakuten: The SharedPreference just saves the state of the miniapp so need to check if the data source code of mini app whether it's still exsited.

PR #45, opened 2020-02-21
feat: add api to fetch single mini app info

ts-nitish-a-jain:
This PR is about adding another method to the public API interface to fetch metadata of a single mini app by it's appId.

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -36,10 +36,15 @@ internal class ApiClient @VisibleForTesting constructor(
...
 
     suspend fun list(): List<MiniAppInfo> {
-        val request = listingApi.list(hostAppId, hostAppVersionId)
+        val request = appInfoApi.list(hostAppId, hostAppVersionId)
         return requestExecutor.executeRequest(request)
     }
 
+    suspend fun fetchInfo(appId: String): MiniAppInfo {
+        val request = appInfoApi.fetchInfo(hostAppId, hostAppVersionId, appId)
+        return requestExecutor.executeRequest(request).first()
```
minh-rakuten: Instead of returning a single object, the backend returns a list -> it looks painful to me 😂
1/ `List<T>.first()` throws an exception for empty list. May you want to handle it with our sdk exception?

2/ We assume that there is only one item in the response list but let's create a unit test for the case: `should only get the first item from mini app list response`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -31,6 +31,13 @@ abstract class MiniApp internal constructor() {
         versionId: String
     ): MiniAppDisplay
 
+    /**
+     * Fetches meta data information of a mini app.
+     * @return [MiniAppInfo] for the provided appId of a mini app
+     * @throws [MiniAppSdkException] when fetching fails from the BE server for any reason.
+     */
+    abstract suspend fun fetchMiniAppInfo(appId: String): MiniAppInfo
```
corycaywood: Maybe this can be `fetchInfo`? `fetchMiniAppInfo` sounds redundant since the class is already named `MiniApp`.

ts-nitish-a-jain: ☯️

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
ts-nitish-a-jain: The first part I agree with, but from backend, for now, we have to face the list.
The other two have been updated :) 👍

PR #49, opened 2020-02-26
docs: Add note in userguide about host app version

corycaywood:
# Description
If the user puts a number such as 1 or 1.0 as the value for the host app version, then the app will crash

# Checklist


corycaywood:
> I think it might be a good idea to add a note in general.
> All `MetaData` values are expected to be of type String in the Key-Value Map (considering the SDK implementation).

I have updated the message to reflect this.

PR #50, opened 2020-02-28
Feature/sample-app-switch

minh-rakuten:
# Description
The sample app with Display List and Input:
- Switching between Display List and Input
- Single tap event on the sample app

## Links
[MINI-581](https://jira.rakuten-it.com/jira/browse/MINI-581)

### DIFF
--- a/testapp/src/main/AndroidManifest.xml
+++ b/testapp/src/main/AndroidManifest.xml
@@ -12,24 +12,19 @@
         android:theme="@style/AppTheme"
         tools:ignore="GoogleAppIndexingWarning">
 
-        <activity android:name="com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity">
+        <activity
+            android:name="com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
-
         <activity android:name="com.rakuten.tech.mobile.testapp.ui.base.BaseActivity"/>
         <activity android:name="com.rakuten.tech.mobile.testapp.ui.display.MiniAppDisplayActivity"/>
-        <activity android:name="com.rakuten.tech.mobile.testapp.ui.input.MiniAppInputActivity"
-            android:label="Input MiniApp">
-            <intent-filter>
-                <action android:name="android.intent.action.MAIN" />
-
-                <category android:name="android.intent.category.LAUNCHER" />
-            </intent-filter>
-        </activity>
+        <activity
+            android:name="com.rakuten.tech.mobile.testapp.ui.input.MiniAppInputActivity"
+            android:label="Input MiniApp"/>
 
         <meta-data
             android:name="com.rakuten.tech.mobile.miniapp.BaseUrl"
--- a/testapp/src/main/AndroidManifest.xml
+++ b/testapp/src/main/AndroidManifest.xml
@@ -13,7 +13,8 @@
         tools:ignore="GoogleAppIndexingWarning">
 
         <activity
-            android:name="com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity">
+            android:name="com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity"
+            android:label="@string/action_display_list">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
@@ -24,7 +25,7 @@
         <activity android:name="com.rakuten.tech.mobile.testapp.ui.display.MiniAppDisplayActivity"/>
         <activity
             android:name="com.rakuten.tech.mobile.testapp.ui.input.MiniAppInputActivity"
-            android:label="Input MiniApp"/>
+            android:label="@string/action_go_input"/>
 
         <meta-data
             android:name="com.rakuten.tech.mobile.miniapp.BaseUrl"
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/RaceExecution.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/RaceExecution.kt
@@ -5,7 +5,7 @@ import android.os.SystemClock
 /**
  * The simple controller which ensures only one method executed at a time.
  */
-class SingleExecution {
+class RaceExecution {
     private var interval:Int = 800
     private var lastTimeClicked:Long = 0
 
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
@@ -1,0 +1,18 @@
+package com.rakuten.tech.mobile.testapp.helper
+
+import android.os.SystemClock
+
+/**
+ * The simple controller which ensures only one method executed at a time.
+ */
+class SingleExecution {
+    private var interval:Int = 800
+    private var lastTimeClicked:Long = 0
+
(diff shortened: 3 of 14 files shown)


testapp/src/main/AndroidManifest.xml
```diff
@@ -12,24 +12,19 @@
         android:theme="@style/AppTheme"
         tools:ignore="GoogleAppIndexingWarning">
 
-        <activity android:name="com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity">
+        <activity
+            android:name="com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity">
```
ts-nitish-a-jain: Let's define a label here too?

testapp/src/main/AndroidManifest.xml
```diff
@@ -12,24 +12,19 @@
...
-            android:label="Input MiniApp">
-            <intent-filter>
-                <action android:name="android.intent.action.MAIN" />
-
-                <category android:name="android.intent.category.LAUNCHER" />
-            </intent-filter>
-        </activity>
+        <activity
+            android:name="com.rakuten.tech.mobile.testapp.ui.input.MiniAppInputActivity"
+            android:label="Input MiniApp"/>
```
ts-nitish-a-jain: How about `Display MiniApp From ID`?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
```diff
@@ -0,0 +1,18 @@
+package com.rakuten.tech.mobile.testapp.helper
+
+import android.os.SystemClock
+
+/**
+ * The simple controller which ensures only one method executed at a time.
+ */
+class SingleExecution {
```
ts-nitish-a-jain: I'm not sure. Will the default `run` from Kotlin suffice? 
It has something like:
`contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }`

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/input/MiniAppInputActivity.kt
```diff
@@ -1,23 +1,34 @@
...
+import com.rakuten.tech.mobile.testapp.ui.base.BaseActivity
 import com.rakuten.tech.mobile.testapp.ui.display.MiniAppDisplayActivity
+import com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListActivity
 import kotlinx.android.synthetic.main.mini_app_input_activity.*
 
-class MiniAppInputActivity : AppCompatActivity() {
+class MiniAppInputActivity : BaseActivity() {
+
+    companion object {
+        fun start(context: Context) {
```
ts-nitish-a-jain: Any specific reason for placing the launch-logic within?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/miniapplist/MiniAppListFragment.kt
```diff
@@ -47,14 +48,19 @@ class MiniAppListFragment : BaseFragment(), MiniAppList {
 
     override fun onActivityCreated(savedInstanceState: Bundle?) {
         super.onActivityCreated(savedInstanceState)
+
+        binding.btnInput.setOnClickListener {
```
ts-nitish-a-jain: How about extracting this into the binding itself?

testapp/src/main/res/values/strings.xml
```diff
@@ -5,4 +5,6 @@
     <string name="lb_app_id">App ID</string>
     <string name="action_display">Display</string>
     <string name="error_empty">Required</string>
+    <string name="action_go_input">Input MiniApp</string>
```
ts-nitish-a-jain: All entities might require updates to naming.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
minh-rakuten: ```contract { callsInPlace(block, InvocationKind.EXACTLY_ONCE) }```
-> could you please give me the resource document of this?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/input/MiniAppInputActivity.kt
minh-rakuten: Yeah. There is a restriction here when launching this activity: `addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP or Intent.FLAG_ACTIVITY_NEW_TASK)`

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
ts-nitish-a-jain: Check out `Standard.kt`

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/input/MiniAppInputActivity.kt
ts-nitish-a-jain: I noticed another one in the list activity. How about extracting them into a unit like `ActivityLauncher.kt`. What do you think?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
corycaywood: `contract` looks pretty cool. It's still experimental, but that should be okay since this is for the Demo App.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/input/MiniAppInputActivity.kt
minh-rakuten: I think just keep it separately as it is now. As specific activity should have its own launching method.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/SingleExecution.kt
minh-rakuten: The SingleExecution implementation here ensures only one type of method is going be executed at the specific time. i.e tap two different buttons on screen at the same time.
Using `InvocationKind.EXACTLY_ONCE` does not match the situation here. 

Perhaps, can you suggest the better name instead of `SingleExecution`?

ts-nitish-a-jain: Got it. Thanks..!
I was looking at the kdoc in the file. How about `RaceController`. 
It is like a race between the functions. The one who reaches first wins the race.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/RaceExecution.kt
```diff
@@ -0,0 +1,18 @@
+package com.rakuten.tech.mobile.testapp.helper
+
+import android.os.SystemClock
+
+/**
+ * The simple controller which ensures only one method executed at a time.
```
ts-nitish-a-jain:
```suggestion
 * A simple controller that ensures only one method is executed at a time where there is a race between two. 
```

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/RaceExecution.kt
```diff
@@ -0,0 +1,18 @@
...
+ */
+class RaceExecution {
```
ts-nitish-a-jain:
```suggestion
class RaceExecutor {
```

PR #51, opened 2020-03-02
Change App Id and Key at runtime

corycaywood:
# Description
This change is required so that we can give a build to customers in which they can easily change the App ID and Subscription Key for testing a Mini App using a test project. 

This required me to make the `MiniApp#init` public so that the SDK can be re-initialized. I guess I could add a `@RestrcitTo` annotation so that it can only be used from the Sample App. But it seems like developers might also want to be able to use that feature for QA purposes. What do you guys think?

## Links
MINI-427

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -42,16 +47,37 @@ abstract class MiniApp internal constructor() {
...
-            miniAppLister: MiniAppLister
-        ) {
+        /**
+         * Initializes the SDK with the specified settings.
+         * Note that you are not required to manually call this method. The SDK is automatically
+         * initialized using the settings in AndroidManifest.xml. Only call this method if you
+         * wish to re-initialize the SDK with custom settings.
+         */
+        @Suppress("LongMethod")
+        fun init(
```
minh-rakuten: `App ID` and `Key` only involve with `ApiClient` so we could have the feature for updating the `ApiClient` only. The `context` from the content provider should be keep unchanged.

What we're doing at first is passing the `Context` from `ContentProvider` which ensures no corruption while the client app using Miniapp SDK because the `context` is still alive as long as client app is still working.

If the client developers freely pass the `context` from any place they want, our sdk cannot control it.

corycaywood:
> Thanks for the attempt..!
> From the face of it, this can cause a potential object-littering for the complete SDK interfacing and associated object-creation chains.
> 
> * The moment one wants to customize, there are two SDK interfacing objects with the ability to +1 (count) with every change of setting.
> 
> In principle, changing settings at runtime:
> 
> * needs to provide/ensure a graceful exit to existing/running operation (on old values)
> * needs to avoid object-littering as much as possible (this may encounter trade-offs)
> 
> If this is business critical, it can be taken into a separate branch for further work/review.

@ts-nitish-a-jain You are right, this solution really isn't ideal. But I think it will require significant refactoring and add a lot of complexity to the code if we do this in the proper way by adding some kind of `setAppId` and `setSubscriptionKey` methods.

This is a feature that we really only need for the Demo App and we weren't originally intending to have something like this for developers.  And re-initializing the SDK like this should be fine as long as the functionality is always accessed directly via `MiniApp#instance` and we don't store the instance object in the Demo App anywhere. So maybe we can make this an unsupported functionality by using the `@RestrictTo` annotation, hiding the method from javadocs, and adding a comment that it is an internal function.

What do you think? Do you have any other ideas?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
corycaywood: That's a good point.  I was thinking maybe I can move the `ApiClient`, `MiniAppDownloader`, and `MiniAppLister` instantiation into `RealMiniApp` and then we can avoid a lot of these issues.

I'll update this PR with some changes.

ts-nitish-a-jain:
> > Thanks for the attempt..!
> > From the face of it, this can cause a potential object-littering for the complete SDK interfacing and associated object-creation chains.
> > 
> > * The moment one wants to customize, there are two SDK interfacing objects with the ability to +1 (count) with every change of setting.
> > 
> > In principle, changing settings at runtime:
> > 
> > * needs to provide/ensure a graceful exit to existing/running operation (on old values)
> > * needs to avoid object-littering as much as possible (this may encounter trade-offs)
> > 
> > If this is business critical, it can be taken into a separate branch for further work/review.
> 
> @ts-nitish-a-jain You are right, this solution really isn't ideal. But I think it will require significant refactoring and add a lot of complexity to the code if we do this in the proper way by adding some kind of `setAppId` and `setSubscriptionKey` methods.
> 
> This is a feature that we really only need for the Demo App and we weren't originally intending to have something like this for developers. And re-initializing the SDK like this should be fine as long as the functionality is always accessed directly via `MiniApp#instance` and we don't store the instance object in the Demo App anywhere. So maybe we can make this an unsupported functionality by using the `@RestrictTo` annotation, hiding the method from javadocs, and adding a comment that it is an internal function.
> 
> What do you think? Do you have any other ideas?

I agree that it will take some effort. An ideal plan might take some time. There are some usable parts while decoupling as mentioned by @minh-rakuten.

A way to approach would be to create a config using the builder and using it like `MiniApp.instance(withConfig: SdkConfig)`, this will help to ensure SDK creates it for the app and ensures the reference update in MiniApp.kt and yet boxed within SDK scope. 🤔
Only re-initializing the updated parts.

The choice was made for auto-initialization. And now, I think it is better to transit and transit well, to keep it in a separate branch until it does and maybe release in case of urgency. 🤔

corycaywood:
> I agree that it will take some effort. An ideal plan might take some time. There are some usable parts while decoupling as mentioned by @minh-rakuten.
> 
> A way to approach would be to create a config using the builder and using it like `MiniApp.instance(withConfig: SdkConfig)`, this will help to ensure SDK creates it for the app and ensures the reference update in `MiniApp.kt` and yet boxed within SDK scope. 🤔
> Only re-initializing the updated parts.
> 
> The choice was made for auto-initialization. And now, I think it is better to transit and transit well, to keep it in a separate branch until it does and maybe release in case of urgency. 🤔

I thought about this and it seems to make sense. Providing alternative settings to `instance` is similar to what Firebase does. I looked into their SDK to see how they're doing it, and they are storing a map of instances and then add a new instance to the map whenever a user instantiates a new one.

I tried making these updates using a map that stores the instances by a key which is created from the provided settings. The disadvantage is that we have to store the Context as a static field, but maybe this is okay because it comes from the ContentProvider rather than an Activity. @ts-nitish-a-jain @minh-rakuten I have pushed these changes. What do you think?

I also thought about keeping the `MiniApp#settings` function, moving `ApiClient` instantiation into `RealMiniApp` and then exposing the `apiClient` property from `MiniAppDownloader` and `MiniAppLister`. Then we could just re-create the `ApiClient` whenever the settings change and update the `apiClient` instance where it's used. So that's also another option.

ts-nitish-a-jain:
@corycaywood can you share the link to Firebase implementation?

minh-rakuten:
> @corycaywood can you share the link to Firebase implementation?

It might be this one [FirebaseApp](https://firebase.google.com/docs/reference/android/com/google/firebase/FirebaseApp)

corycaywood:
> > @corycaywood can you share the link to Firebase implementation?
> 
> It might be this one [FirebaseApp](https://firebase.google.com/docs/reference/android/com/google/firebase/FirebaseApp)

Yes, that's the SDK. Here's the specific section in the source code:
https://github.com/firebase/firebase-android-sdk/blob/10c8922651362d647417aa40115541f652b41927/firebase-common/src/main/java/com/google/firebase/FirebaseApp.java#L102

Though the approach is a little bit different on Firebase. You can manually initialize another `FirebaseApp` and set an `app name` key. So something like this:

```
// Retrieve the automatically initialized instance
val autoInstance = FirebaseApp.getInstance()

// Create and retrieve customized Instance
val options = FirebaseOptions.Builder()
            .setApplicationId(context.getPackageName())
            .setApiKey("apikey")
            .build();
FirebaseApp.initializeApp(context, options, "custom_instance");
val customInstance = FirebaseApp.getInstance("custom_instance");
```

ts-nitish-a-jain:
@corycaywood Upcoming things to be modified:
- Preferences 
- Storage

minh-rakuten:
> Then we could just re-create the ApiClient whenever the settings change and update the apiClient instance where it's used.

I still prefer our original implementation approach. Single instance and the implementation changes here for updating ApiClient only.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -31,6 +36,8 @@ abstract class MiniApp internal constructor() {
         versionId: String
     ): MiniAppDisplay
 
+    protected abstract fun updateConfiguration(settings: MiniAppSdkConfig)
```
ts-nitish-a-jain: Let's make this `internal`.
And, place it after all the public API methods in the file (helps with the reading and lookup). What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -40,25 +47,59 @@ abstract class MiniApp internal constructor() {
 
     companion object {
         private lateinit var instance: MiniApp
+        private lateinit var defaultSettings: MiniAppSdkConfig
```
ts-nitish-a-jain: I think `defaultConfig` looks like a better fit. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -40,25 +47,59 @@ abstract class MiniApp internal constructor() {
...
+         * Instance of [MiniApp] with custom [MiniAppSdkConfig].
+         * This function should only be used if you wish to use the SDK with custom settings
+         * which are changed at runtime for QA purposes, etc.
+         * Note that the default [instance] without [settings] uses the config settings from AndroidManifest.xml.
          *
          * @return [MiniApp] instance
          */
         @JvmStatic
-        fun instance(): MiniApp = instance
+        fun instance(settings: MiniAppSdkConfig = defaultSettings): MiniApp =
```
ts-nitish-a-jain: Need to update the documentation for the method.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -40,25 +47,59 @@ abstract class MiniApp internal constructor() {
...
-            miniAppInfoFetcher: MiniAppInfoFetcher
+            context: Context,
+            baseUrl: String,
+            rasAppId: String,
+            subscriptionKey: String,
+            hostAppVersionId: String
         ) {
+            val miniAppStatus = MiniAppStatus(context)
+            val storage = MiniAppStorage(FileWriter(), context.filesDir)
+            val apiClient = ApiClient(
```
ts-nitish-a-jain: This `apiClient` client needs to be registered with other instances.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -40,25 +47,59 @@ abstract class MiniApp internal constructor() {
...
             )
         }
     }
 }
+
+/**
+ * Config for the Mini App SDK.
+ * Contains settings which are used when sending requests to the Mini App API.
+ */
+data class MiniAppSdkConfig(
```
ts-nitish-a-jain: Let's move this to a separate class and add doc for the properties as well.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -8,10 +8,14 @@ import com.rakuten.tech.mobile.miniapp.storage.MiniAppStorage
 
 internal class MiniAppDownloader(
     private val storage: MiniAppStorage,
-    private val apiClient: ApiClient,
+    private var apiClient: ApiClient,
     private val miniAppStatus: MiniAppStatus
 ) {
 
+    fun setApiClient(apiClient: ApiClient) {
```
ts-nitish-a-jain: Let's place the update methods at the bottom of the class?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -28,4 +29,15 @@ internal class RealMiniApp(
             displayer.createMiniAppDisplay(basePath, appId)
         }
     }
+
+    override fun updateConfiguration(settings: MiniAppSdkConfig) {
+        miniAppDownloader.setApiClient(
+            ApiClient(
```
ts-nitish-a-jain: Both `MiniAppDownloader` and `MiniAppInfoFetcher` would also need to update the ApiClient.
It will be good to store the ApiClient instances like it was happening for the SDK instances previously. What do you think?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuBaseActivity.kt
```diff
@@ -0,0 +1,61 @@
...
+        val settingsDialog = LayoutInflater.from(this)
+            .inflate(R.layout.app_settings_dialog, null, false)
+
+        val appId = settingsDialog.findViewById(R.id.app_id) as EditText
+        val subscriptionKey = settingsDialog.findViewById(R.id.subscription_key) as EditText
+        appId.setText(settings.appId)
+        subscriptionKey.setText(settings.subscriptionKey)
+
+        AlertDialog.Builder(this)
+            .setTitle("App Settings")
```
ts-nitish-a-jain: Let's also extract the strings ✌️

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
ts-nitish-a-jain: Also, we will need to add validations for the properties of `MiniAppSdkConfig` and throw some error when invalid settings are provided for.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
minh-rakuten: Actually, I already edited it so could you read it again or suggest the content 🙇

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
minh-rakuten: I intended to create a interface for any component required to update `ApiClient`

```
/**
 * The interface for MiniApp components which require api interaction.
*/
internal interface MiniAppApiImpl {
    /** Is executed when there is a change of SDK configuration at runtime **/
    fun updateApiClient(apiClient: ApiClient)
}
```

What do you think?

ts-nitish-a-jain: It is a good approach too. 👍 

In the current state, because RealMiniApp already is aware of `MiniAppDownloader` & `MiniAppInfoFetcher`, it might not bring any added advantage (i.e., if you want to do something like- for all implementers -> update ApiClient, as this will make you maintain a collection of all the implementers).

But, on the other hand, it can help the new classes which intend to use ApiClient. As they will be aware that the ApiClient can get updated by virtue of this contract.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/MiniAppApiImpl.kt
```diff
@@ -0,0 +1,9 @@
+package com.rakuten.tech.mobile.miniapp.api
+
+/**
+ * The interface for MiniApp components which require api interaction.
+ */
+internal interface MiniAppApiImpl {
```
corycaywood: I also think this name doesn't explain the purpose very well, but I'm not sure what a good name would be. Maybe `UpdatableApiClient`? But then we might also add additional functionality other than just updating the client...

testapp/build.gradle
```diff
@@ -14,9 +14,9 @@ android {
 
         manifestPlaceholders = [
                 baseUrl        : property("MINIAPP_SERVER_BASE_URL") ?: "https://www.example.com/",
-                hostAppVersion : property("HOST_APP_VERSION") ?: "",
-                appId          : property("HOST_APP_ID") ?: "",
-                subscriptionKey: property("HOST_APP_SUBSCRIPTION_KEY") ?: ""
+                hostAppVersion : property("HOST_APP_VERSION") ?: "a6a57b4a-17eb-48ed-b307-95f9ae1e1e26",
+                appId          : property("HOST_APP_ID") ?: "e03b8516-b230-4d0a-a1bd-8a61b0ad46a2",
+                subscriptionKey: property("HOST_APP_SUBSCRIPTION_KEY") ?: "0fd1d40c72d143918cddc902b1c031dd"
```
corycaywood: Maybe we can make these be obviously fake values such as `test-subscription-key`, etc.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/MiniAppApiImpl.kt
minh-rakuten: Yeah let's pick `UpdatableApiClient`. I never good at naming.

minh-rakuten:
> Also, I think we've got this functionality refined pretty well, so we should add some unit tests now.

@corycaywood Done. Everything looks okay now.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/MiniAppSpec.kt
```diff
@@ -0,0 +1,28 @@
...
+
+class MiniAppSpec {
+
+    @Test
+    fun `should update configuration when get instance of MiniApp`() {
+        val miniAppSdkConfig: MiniAppSdkConfig = mock()
+        val miniAppCompanion: MiniApp.Companion = MiniApp.Companion
+        val miniApp: MiniApp = mock()
+
+        val instance = MiniApp.Companion::class.memberProperties.first { it.name == "instance" }
```
corycaywood: We typically avoid using reflection in tests. Is there another way to test this?

I guess you could make `instance` be `internal` with `@VisibleForTesting`? It's still not ideal, but it's better than using reflection.

minh-rakuten: Unfortunately, I don't find another way to solve it so I applied your suggestion.

PR #52, opened 2020-03-04
Feature/update error response handling

ts-nitish-a-jain:
# Description
This PR include:
- Modifications for handling 401/403 error responses (which are in a different format than the rest) as received from the backend.
- Minor refactor and fix for detekt warnings

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
@@ -91,7 +91,7 @@ internal class RetrofitRequestExecutor(
         val errorData = response.errorBody() ?: throw sdkExceptionForInternalServerError()
         when (response.code()) {
             401, 403 -> throw MiniAppSdkException(
-                convertApiCatalogErrorToMsg(
+                convertAuthErrorToMsg(
                     response, errorData, createErrorConvertor(retrofit)
                 )
             )
@@ -103,10 +103,10 @@ internal class RetrofitRequestExecutor(
         }
     }
 
-    private fun convertApiCatalogErrorToMsg(
+    private fun convertAuthErrorToMsg(
         response: Response<in Nothing>,
         error: ResponseBody,
-        converter: Converter<ResponseBody, ApiCatalogErrorResponse>
+        converter: Converter<ResponseBody, AuthErrorResponse>
     ) = errorMsgFromHttpException(response, converter.convert(error)?.message)
 
     private fun convertStandardHttpErrorToMsg(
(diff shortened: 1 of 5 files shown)


ts-nitish-a-jain:
Local shows me 4% ↑ image. What am I doing differently?

image

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -63,52 +64,79 @@ internal class ApiClient @VisibleForTesting constructor(
...
+    ).message()
 }
 
-internal data class ErrorResponse(
+internal data class HttpErrorResponse(
     val code: Int,
+    override val message: String
+) : ErrorResponse
+
+internal data class ApiCatalogErrorResponse(
```
corycaywood: I think this name is exposing details about the API that we don't need to know as the SDK. How about `AuthErrorResponse` instead?

Climbatize: On iOS we called it UnauthorizedData, as opposed to ErrorData we already had in place

ts-nitish-a-jain: The choice I made was by considering future reference and readability. I would definitely forget why there are two types from the backend. The SDK is managing the deviation of response types from the backend. So, it is aware indeed.

I had two options, mark that domain-effect into the names of behavior or record it as a comment. I choose the former.

Unless, there is a strong opinion on renaming it, I would like to keep it the same.

corycaywood: I think something that refers to it being related to Authorization is more explanatory. It's not the SDK's responsibility to document why certain choices were made by the API side. We just have to document that there is a different response for different response codes.

Also, keep in mind that this is a public SDK, so outside developers will have no idea what API Catalog is.

ts-nitish-a-jain: Thanks @corycaywood, I think that's reasonable. 👍

corycaywood:
> Local shows me 4% ↑ image. What am I doing differently?
> 
> image

CodeCov works in mysterious ways. Sometimes it will show a different percentage when there were only changes to build files and not to source files.

ts-nitish-a-jain:
> > Local shows me 4% ↑ image. What am I doing differently?
> > image
> 
> CodeCov works in mysterious ways. Sometimes it will show a different percentage when there were only changes to build files and not to source files.

That's some interesting stuff..! 🤣

PR #53, opened 2020-03-04
Feature/update impl

minh-rakuten:
# Description
Specification requirement:
> The Mobile Demo App should fail to render a webview for Mini Apps that are not currently published.
Including previously cached versions.

Adding the latest miniapp version checking. Only latest version of miniapp can be rendered.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -13,6 +13,7 @@ internal class MiniAppDownloader(
 ) {
 
     suspend fun getMiniApp(appId: String, versionId: String): String = when {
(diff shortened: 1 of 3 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -12,7 +12,9 @@ internal class MiniAppDownloader(
     private val miniAppStatus: MiniAppStatus
 ) {
 
+    /** Only run the latest version of specified MiniApp. **/
     suspend fun getMiniApp(appId: String, versionId: String): String = when {
+        apiClient.fetchInfo(appId).version.versionId != versionId -> throw sdkExceptionForInvalidVersion()
```
ts-nitish-a-jain: How about extracting this into`IsValidVersion`?

PR #56, opened 2020-03-10
feat: auto clear up WebView resources for usual lifecycle scenarios

ts-nitish-a-jain:
# Description
Adds an alternative for the automatic clearing of WebView in the display unit for the usual lifecycle oriented scenarios.

## Links
[MINI-499](https://jira.rakuten-it.com/jira/browse/MINI-499)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
@@ -9,7 +9,6 @@ import androidx.lifecycle.LifecycleObserver
  * This contract complies to Android's [LifecycleObserver] contract, and when made to observe
  * the lifecycle, it automatically clears up the view state and any services registered with.
  */
-// This contract keeps the display of a mini app implementation agnostic
 interface MiniAppDisplay: LifecycleObserver {
 
     /**
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
@@ -1,12 +1,16 @@
 package com.rakuten.tech.mobile.miniapp
 
 import android.view.View
+import androidx.lifecycle.LifecycleObserver
 
 /**
  * This represents the contract by which the host app can interact with the
  * display unit of the mini app.
+ * This contract complies to Android's [LifecycleObserver] contract, and when made to observe
+ * the lifecycle, it automatically clears up the view state and any services registered with.
  */
-interface MiniAppDisplay { // This contract keeps the display of a mini app implementation agnostic
+// This contract keeps the display of a mini app implementation agnostic
+interface MiniAppDisplay: LifecycleObserver {
 
     /**
      * Provides the view associated with the mini app to the caller for showing the mini app.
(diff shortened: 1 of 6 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -45,18 +47,23 @@ internal class RealMiniAppDisplay(
 
     override fun getMiniAppView(): View = this
 
-    override fun destroyView() {
+    override fun destroyView() = clearUp()
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
+    fun clearUp() {
```
corycaywood: The function needs to be public in order to use the `@OnLifecycleEvent` annotation? Could you just add the annotation directly to `destroyView` and avoid having two methods?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
```diff
@@ -30,6 +33,7 @@ class MiniAppDisplayViewModel constructor(
         try {
             _isLoading.postValue(true)
             miniAppDisplay = miniapp.create(appId, versionId)
+            hostLifeCycle?.addObserver(miniAppDisplay)
             _miniAppView.postValue(miniAppDisplay.getMiniAppView())
```
minh-rakuten: I was thinking about passing the `hostLifeCycle` into `miniAppDisplay.getMiniAppView(hostLifeCycle)` to put a constraint to developer.

`RealMiniAppDisplay` could have this function:
 
```
fun registerLifecycle(lifecycle: Lifecycle) {
    lifecycle.addObserver(this)
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -1,12 +1,16 @@
 package com.rakuten.tech.mobile.miniapp
 
 import android.view.View
+import androidx.lifecycle.LifecycleObserver
 
 /**
  * This represents the contract by which the host app can interact with the
  * display unit of the mini app.
+ * This contract complies to Android's [LifecycleObserver] contract, and when made to observe
+ * the lifecycle, it automatically clears up the view state and any services registered with.
```
corycaywood: Can you add a short code example to the User Guide of how to use this? Maybe it can go under an "Advanced" or "FAQ" section?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -1,12 +1,16 @@
...
  */
-interface MiniAppDisplay { // This contract keeps the display of a mini app implementation agnostic
+// This contract keeps the display of a mini app implementation agnostic
```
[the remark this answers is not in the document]
ts-nitish-a-jain: It had some history. Removed it out.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
ts-nitish-a-jain: I think merging is better as well. Updated it.
(Developer Documentation had no scopes, though works with private as well).

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
ts-nitish-a-jain: Sure, just added one.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
ts-nitish-a-jain: As we discussed it out, there were a few reasons behind the approach I followed:
- I was intending to not to force the behavior and be neutral for a variety of host app developer needs (the advanced ones could utilize lifecycle and the traditional ones could just do it manually). It may even happen that some of the old apps might not even have a lifecycle in there.

- At this moment, `RealMiniAppDisplay` has no dependency on its operations which is fulfilled via lifecycle and is mandated. it would have definitely made sense to add it to the contract if there was more operational dependency. I think in the near future we might just add it once there is an increase in dependency.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
ts-nitish-a-jain: Thanks @corycaywood..! I think it is good to update the guide.

PR #59, opened 2020-03-13
build: update gradle dependency version

ts-nitish-a-jain:
# Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -30,7 +30,7 @@ buildscript {
     CONFIG.versions.android.sdk.min = 21
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.6.0'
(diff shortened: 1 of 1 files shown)


corycaywood:
Strange that Codecov is still reporting only 61%, even after updating Jacoco. When I run Jacoco locally, it says 69% for instructions and 78% for branches 🤔 Maybe we need to reset the Codecov data?

ts-nitish-a-jain:
> Strange that Codecov is still reporting only 61%, even after updating Jacoco. When I run Jacoco locally, it says 69% for instructions and 78% for branches 🤔 Maybe we need to reset the Codecov data?

Yea, let's try that, might fix Codecov strangeness 👍

PR #60, opened 2020-03-13
feature: Delete old versions

minh-rakuten:
# Description
After a new version of a Mini App has finished downloading, delete the files excluding the new version in that Mini App package.

## Links
[MINI-613](https://jira.rakuten-it.com/jira/browse/MINI-613)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -17,7 +17,7 @@ internal class MiniAppDownloader(
     suspend fun getMiniApp(appId: String, versionId: String): String = when {
         !isLatestVersion(appId, versionId) -> throw sdkExceptionForInvalidVersion()
         miniAppStatus
-            .isVersionDownloaded(appId, versionId) -> storage.getSavePathForApp(appId, versionId)
+            .isVersionDownloaded(appId, versionId) -> storage.getPathMiniAppVersion(appId, versionId)
         else -> startDownload(appId, versionId)
     }
 
@@ -48,7 +48,7 @@ internal class MiniAppDownloader(
         versionId: String,
         manifest: ManifestEntity
     ): String {
-        val baseSavePath = storage.getSavePathForApp(appId, versionId)
+        val baseSavePath = storage.getPathMiniAppVersion(appId, versionId)
         when {
             isManifestValid(manifest) -> {
                 for (file in manifest.files) {
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -6,8 +6,6 @@ import com.rakuten.tech.mobile.miniapp.api.ManifestEntity
 import com.rakuten.tech.mobile.miniapp.api.UpdatableApiClient
 import com.rakuten.tech.mobile.miniapp.storage.MiniAppStatus
 import com.rakuten.tech.mobile.miniapp.storage.MiniAppStorage
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.withContext
 
 internal class MiniAppDownloader(
     private val storage: MiniAppStorage,
@@ -58,7 +56,7 @@ internal class MiniAppDownloader(
                     storage.saveFile(file, baseSavePath, response.byteStream())
                 }
                 miniAppStatus.setVersionDownloaded(appId, versionId, true)
-                cleanUpOldData(appId, versionId)
+                storage.removeOutdatedVersionApp(appId, versionId)
                 return baseSavePath
             }
             // If backend functions correctly, this should never happen
@@ -73,9 +71,5 @@ internal class MiniAppDownloader(
 
     override fun updateApiClient(apiClient: ApiClient) {
         this.apiClient = apiClient
-    }
-
-    private suspend fun cleanUpOldData(appId: String, versionId: String) = withContext(Dispatchers.IO) {
-        storage.removeOutdatedVersionApp(appId, versionId)
     }
 }
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -17,7 +17,7 @@ internal class MiniAppDownloader(
     suspend fun getMiniApp(appId: String, versionId: String): String = when {
         !isLatestVersion(appId, versionId) -> throw sdkExceptionForInvalidVersion()
         miniAppStatus
-            .isVersionDownloaded(appId, versionId) -> storage.getPathMiniAppVersion(appId, versionId)
+            .isVersionDownloaded(appId, versionId) -> storage.getMiniAppVersionPath(appId, versionId)
         else -> startDownload(appId, versionId)
     }
 
@@ -48,7 +48,7 @@ internal class MiniAppDownloader(
         versionId: String,
         manifest: ManifestEntity
     ): String {
-        val baseSavePath = storage.getPathMiniAppVersion(appId, versionId)
+        val baseSavePath = storage.getMiniAppVersionPath(appId, versionId)
         when {
             isManifestValid(manifest) -> {
                 for (file in manifest.files) {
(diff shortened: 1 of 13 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -56,6 +56,7 @@ internal class MiniAppDownloader(
                     storage.saveFile(file, baseSavePath, response.byteStream())
                 }
                 miniAppStatus.setVersionDownloaded(appId, versionId, true)
+                storage.removeOutdatedVersionApp(appId, versionId)
```
ts-nitish-a-jain: The display doesn't need to be blocked/waiting for deletion. This should happen in parallel. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
```diff
@@ -43,6 +47,25 @@ internal class MiniAppStorage(
     @VisibleForTesting
     fun getFileName(file: String) = urlToFileInfoParser.getFileName(file)
 
-    fun getSavePathForApp(appId: String, versionId: String) =
-        "${basePath.path}/$SUB_DIR_MINIAPP/$appId/$versionId"
+    @VisibleForTesting
+    internal fun getParentPathApp(appId: String) = "${basePath.path}/$SUB_DIR_MINIAPP/$appId/"
+
+    fun getSavePathForApp(appId: String, versionId: String) = "${getParentPathApp(appId)}$versionId"
```
ts-nitish-a-jain: I think let's break the paths down and use them up, they will tend to look untidy soon.

```
basePathHostApp : basePath.path
rootPathForMiniApp : basePathHostApp/SUB_DIR_MINIAPP/
pathForMiniApp : rootPathForMiniApp/appId/
pathForMiniAppVersionDownload : pathForMiniApp/versionId

File: miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt [hunk 3]
```diff
@@ -43,6 +47,25 @@ internal class MiniAppStorage(
...
+
+    suspend fun removeOutdatedVersionApp(
+        appId: String,
+        versionId: String,
```
ts-nitish-a-jain: `versionId` -> `latestVersionId`, might be better for readability.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
minh-rakuten: Good point. Should run in parallel for deletion because it not affects the main flow usage.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
```diff
@@ -43,6 +49,33 @@ internal class MiniAppStorage(
...
+            withContext(Dispatchers.IO) {
+                launch(Job()) {
+                    val parentFile = File(appPath)
+                    if (parentFile.isDirectory && parentFile.listFiles() != null) {
+                        flow {
+                            parentFile.listFiles()?.forEach { file ->
+                                if (!file.absolutePath.endsWith(latestVersionId))
+                                    emit(file)
+                            }
+                        }.collect { file -> file.deleteRecursively() }
```
corycaywood: Is it possible for the delete to fail? Should we add some error handling here?

On iOS, it just logs a debug message when deletion fails.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
```diff
@@ -43,6 +49,33 @@ internal class MiniAppStorage(
...
+    @VisibleForTesting
+    internal fun getBasePathHostApp() = basePath.path
+
+    @VisibleForTesting
+    internal fun getRootPathMiniApp() = "${getBasePathHostApp()}/$SUB_DIR_MINIAPP/"
+
+    @VisibleForTesting
+    internal fun getPathMiniApp(appId: String) = "${getRootPathMiniApp()}/$appId/"
+
+    fun getPathMiniAppVersion(appId: String, versionId: String) = "${getPathMiniApp(appId)}$versionId"
```
ts-nitish-a-jain: As we spoke, all the functions aren't required to be marked for testing. 
If you only test the consumer function, that should suffice. See if the following looks better.

```
    // Documentation goes here
    private val hostAppBasePath = basePath.path

    // Documentation goes here
    private val miniAppBasePath
        get() = "$hostAppBasePath/$SUB_DIR_MINIAPP/"

    // Documentation goes here
    private val miniAppPath = { appId: String -> "${miniAppBasePath}$appId/" }

    @VisibleForTesting
    // Documentation goes here
    fun getMiniAppVersionPath(appId: String, versionId: String) =
        "${miniAppPath(appId)}$versionId"

File: miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorageTest.kt [hunk 6]
```diff
@@ -44,7 +45,37 @@ class MiniAppStorageTest {
...
+        val oldFile2 = tempFolder.newFile()
+        val latestPackage = tempFolder.newFolder(TEST_ID_MINIAPP_VERSION)
+
+        runBlocking {
+            miniAppStorage.removeOutdatedVersionApp(
+                TEST_ID_MINIAPP,
+                TEST_ID_MINIAPP_VERSION,
+                tempFolder.root.path)
+            // Deletion run in parallel so it requires time to be completed.
+            delay(1000)
```
ts-nitish-a-jain: I'm not sure about this. @corycaywood What do you think?

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorageTest.kt
```diff
@@ -44,7 +45,37 @@ class MiniAppStorageTest {
         verify(localUrlParser, times(1)).getFileName(TEST_URL_FILE)
     }
 
-    private fun getMockedMiniAppStorage() = MiniAppStorage(mock(), mock(), mock())
+    @Test
+    fun `should get consistent path when get path for mini app version`() {
+        val storage = MiniAppStorage(FileWriter(), File(TEST_BASE_PATH))
+
+        storage.getPathMiniApp(TEST_ID_MINIAPP) shouldStartWith storage.getRootPathMiniApp()
```
ts-nitish-a-jain: As we spoke, this should just need to check the formed path like before.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
minh-rakuten: There is an exception `SecurityException` which denies the access of file. I think I'll implement the general exception handling just like iOS.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorage.kt
minh-rakuten: Thanks!
I still keep `miniAppPath` as function for testing purpose and lamda expression functionality is not different from the function itself.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStorageTest.kt
Climbatize: I agree @ts-nitish-a-jain, this test is not safe

corycaywood: Yeah, let's find another way of doing this. Actually, in my opinion we should make `MiniAppStorage#removeOutdatedVersionApp` be synchronous (it's already a suspend function), and move the coroutine launch into `RealMiniApp`. What do. you think?

minh-rakuten: it should be launched inside `MiniAppDownloader` not `RealMiniApp`.

PR #61, opened 2020-03-17
fix: Update ManifestApi response to match spec

corycaywood:
# Description
The API specs have changed the `files` key to `manifest`.

# Checklist


corycaywood:
> 👍 👍
> I'm just wondering passing the `dispatcher` here is the good practice of Coroutine usage or it's a workaround for Coroutine testing? The `runBlockingTest` not working looks like a bug of Kotlin Coroutine to me.

From what I've seen, there is currently no way to use hard-coded, default dispatchers in unit tests (except for `Dispatchers.Main` via `Dispatchers#setMain`): see https://github.com/Kotlin/kotlinx.coroutines/issues/1365

So it seems that the recommended solution is to inject Dispatchers: https://github.com/Kotlin/kotlinx.coroutines/tree/master/kotlinx-coroutines-test#using-time-control-with-withcontext

PR #63, opened 2020-03-26
Feature/interface webview

minh-rakuten:
# Description
Create interface for passing unique id to webview.

I have tested with the TypeScript SDK which I'm working on MINI-618 and it runs well.

## Links
[MINI-608](https://jira.rakuten-it.com/jira/browse/MINI-608)
[MINI-618](https://jira.rakuten-it.com/jira/browse/MINI-618)

# Checklist


miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/RealMiniAppSpec.kt
```diff
@@ -23,6 +23,7 @@ class RealMiniAppSpec {
     private val realMiniApp = RealMiniApp(apiClientRepository, miniAppDownloader, displayer, miniAppInfoFetcher)
     private val miniAppSdkConfig: MiniAppSdkConfig = mock()
     private var apiClient: ApiClient = mock()
+    private val masInf: MiniAppMessageInterface = mock()
```
ts-nitish-a-jain: I think `miniAppMessageInterface` which has been used in `src` is descriptive.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -18,6 +19,7 @@ class RealMiniAppDisplayTest {
     private lateinit var context: Context
     private lateinit var basePath: String
     private lateinit var realDisplay: RealMiniAppDisplay
+    private val masInf: MiniAppMessageInterface = com.nhaarman.mockitokotlin2.mock()
```
ts-nitish-a-jain: How about using import for `com.nhaarman.mockitokotlin2.mock()` and just using mock() like everywhere else I think?

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/RealMiniAppSpec.kt
ts-nitish-a-jain: It will be good to use the same descriptive name in `test`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -75,4 +89,17 @@ internal class MiniAppWebViewClient(private val loader: WebViewAssetLoader) : We
         view: WebView,
         request: WebResourceRequest
     ): WebResourceResponse? = loader.shouldInterceptRequest(request.url)
+
+    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    override fun onPageFinished(webView: WebView, url: String?) {
+        super.onPageFinished(webView, url)
+        try {
+            val inputStream = webView.context.assets.open("inject.js")
```
ts-nitish-a-jain, line 97: @minh-rakuten We discussed this yesterday, let's move the file reading part to an earlier stage. Maybe the init block of `MiniAppWebViewClient`. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageInterface.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.rakuten.tech.mobile.miniapp.MiniAppDisplay
+
+/** Interface for communicating with mini app. **/
+interface MiniAppMessageInterface {
+    /** Get provided id of mini app for any purpose. **/
+    @JavascriptInterface
+    fun getUniqueId(jsonStr: String)
+
+    /** Return a value to mini app. **/
+    fun postValue(jsonStr: String, view: MiniAppDisplay, value: String) {
```
ts-nitish-a-jain, line 14: I think `view` should be either the first or last param in the signature considering it is public API.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -29,4 +30,14 @@ interface MiniAppDisplay : LifecycleObserver {
      * within the same state of parent's lifecycle.
      */
     fun destroyView()
+
+    /**
+     * Provided the interface implementation for javascript interaction.
+     */
+    fun injectJSInterface(miniAppMessageInterface: MiniAppMessageInterface)
```
ts-nitish-a-jain, line 37: Let's add the doc for the param as well.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -29,4 +30,14 @@ interface MiniAppDisplay : LifecycleObserver {
...
+
+    /**
+     * Post a value to mini app with callback.
+     */
+    fun runJsAsyncCallback(callbackId: String, value: String)
```
ts-nitish-a-jain, line 42: Same for this.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -61,9 +73,7 @@ class MiniAppDisplayActivity : BaseActivity() {
                     })
                 }
 
-            launch {
-                viewModel.obtainMiniAppView(appId, versionId)
-            }
+            launch { viewModel.obtainMiniAppView(appId, versionId) }
```
ts-nitish-a-jain, line 76: I think you might have to pull in changes from `master` for this part.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
```diff
@@ -7,6 +7,7 @@ import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
 import com.rakuten.tech.mobile.miniapp.MiniApp
 import com.rakuten.tech.mobile.miniapp.MiniAppDisplay
+import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageInterface
```
ts-nitish-a-jain, line 10: It looks like an unused import.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
minh-rakuten: Sure. However, we still need to determine when to do an injection here so I'll update it later.

@corycaywood As I know for now, I can only injected js after the page is loaded. Could you give a support looking for a solution 🙏 ?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -29,4 +30,18 @@ interface MiniAppDisplay : LifecycleObserver {
...
+     * between native and miniapp.
+     */
+    fun injectJSInterface(miniAppMessageInterface: MiniAppMessageInterface)
+
+    /**
+     * Post a value to mini app with callback.
+     * @property callbackId The id of callback execution storing in miniapp.
+     * @property value The value which is transmitted to miniapp.
+     */
+    fun runJsAsyncCallback(callbackId: String, value: String)
```
corycaywood, line 46: This function should be `internal`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -29,4 +30,18 @@ interface MiniAppDisplay : LifecycleObserver {
      * within the same state of parent's lifecycle.
      */
     fun destroyView()
+
+    /**
+     * Provided the interface implementation for javascript interaction.
+     * @property miniAppMessageInterface The interface implementation which transmits data
+     * between native and miniapp.
+     */
+    fun injectJSInterface(miniAppMessageInterface: MiniAppMessageInterface)
```
corycaywood, line 39: I'm not sure about this because this doesn't give us a guarantee that the native App has defined the interface. What will happen if the native app doesn't define the interface? I think the Mini App will have errors, right?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -75,4 +89,17 @@ internal class MiniAppWebViewClient(private val loader: WebViewAssetLoader) : We
...
+            inputStream.bufferedReader().use(BufferedReader::readText)
+        } catch (e: Exception) {
+            null
+        }?.let {
+            webView.evaluateJavascript("javascript:(function (){$it})()") {}
```
corycaywood, line 102: I think this doesn't need the `javascript:` prefix or the immediately invoked function. It could be just `webView.evaluateJavascript(it) {}`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
corycaywood: I tested this with `onPageStarted` and it seems to be working. Maybe it's due to the issue I commented on below: you don't need to inject the `javascript:` keyword or use an immediately invoked function.

miniapp/src/main/assets/inject.js
```diff
@@ -0,0 +1,42 @@
+var messageQueue = []
```
ts-nitish-a-jain, line 1: I think there is a new version of the script available.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -69,10 +94,29 @@ internal class RealMiniAppDisplay(
...
+    } catch (e: Exception) {
+        null
+    }
 
     override fun shouldInterceptRequest(
         view: WebView,
         request: WebResourceRequest
     ): WebResourceResponse? = loader.shouldInterceptRequest(request.url)
+
+    override fun onPageStarted(webView: WebView, url: String?, favicon: Bitmap?) {
```
corycaywood, line 114: Did you get this working for Android 9?

minh-rakuten: It's working on my Huawei device but not Android 9 emulator.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -0,0 +1,36 @@
...
+import android.webkit.JavascriptInterface
+import com.google.gson.Gson
+import com.rakuten.tech.mobile.miniapp.display.WebViewListener
+
+/** Bridge interface for communicating with mini app. **/
+abstract class MiniAppMessageBridge {
+    private lateinit var webViewListener: WebViewListener
+
+    /** Get provided id of mini app for any purpose. **/
+    abstract fun getUniqueId(callbackId: String)
```
corycaywood, line 12: I think that the callback ID should be internal to the SDK: the implementing client shouldn't need to care about it.

On iOS, this function signature is like this: `fun getUniqueId(): String`, so I think Android should match.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -59,7 +61,11 @@ class MiniAppDisplayActivity : BaseActivity() {
                 }
 
             launch {
-                viewModel.obtainMiniAppView(appId)
+                viewModel.obtainMiniAppView(appId, object: MiniAppMessageBridge() {
+                    override fun getUniqueId(callbackId: String) {
+                        postValue(callbackId, "example_unique_id")
```
corycaywood, line 66: I think that the call to `postValue` can be move internal to the SDK (probably in `postMessage`?). This will allow the `getUniqueId` function to directly return a string.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
corycaywood: Hmm, if it's not working in the emulator, I think we can't rely on this to be a good solution unfortunately. 

I guess we could use `onPageFinished`, but then we would need to add some kind of callback in the JS SDK to notify when the SDK is ready to use. And I don't really like that solution because then a Mini App will have to wait for some time before they can use any of the SDK functionality.

I was thinking, we could inject a `<script>` tag into the `<head>` of the HTML. But it's kind of hacky.... Something like this:
```
    override fun shouldInterceptRequest(
        view: WebView,
        request: WebResourceRequest
    ): WebResourceResponse? {
        if (request.url.host == domain && request.url.path?.endsWith("html") == true) {
            val response = loader.shouldInterceptRequest(request.url)

            val data = response?.data?.bufferedReader().use {
                it?.readText()
            }

            response?.data = data?.replace("<head>", "<head>\n<script>$bridgeJs</script>\n")?.byteInputStream()
            return response
        }

        return loader.shouldInterceptRequest(request.url)
    }
```

What do you think? It looks like our options are either:
1. Inject the JS after the page has loaded and add some `onMiniAppReady` callback in the JS SDK. 
2. or Modify the HTML to inject JS.

corycaywood: Ah, it looks like it's also possible to use `onLoadResource`: https://stackoverflow.com/a/58932747

But if we use that, it would mean that the Mini App must always load at least one resource or else the JS won't get injected. I think the Mini App should always at least try to load the `favicon.ico`, but it is possible to disable that request from the client side though it's not very common to do it.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
minh-rakuten: @corycaywood Previously, I did exactly like that 👍 . So the `postError` should be internal too and only get called from the sdk, right?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
minh-rakuten: Here is my thought of your ideas:

1/ The callback for notifying the ready of sdk could be done in `OnPageStarted` or `OnPageFinished` depending on a webview. I think we could improve it by having a function to check whether `bridge.js` is injected. 
`OnPageStarted`: do injection -> `OnPageFinished`: check if it's not injected, do injection again.
However, this requires so much work on the js sdk side.

2/ Modify the HTML to inject JS: The whole file `bridge.js` will be visible from the web inspector and yeah, it's so hacky.

3/ It works perfectly. I think we can go with this solution if we can force the miniapps load resource or just simply have it in our document declaring that favicon.ico request is mandatory.

corycaywood: Okay, probably 3 is okay. We just have to make sure Mini Apps are aware of the restriction that they need to make the favicon request.

PR #65, opened 2020-04-02
update the location of display button in input screen

ts-nitish-a-jain:
# Description
Updates the location of the display button on the input screen.
Though, I don't think it is a bug in any way, more of a UX suggestion rather (which is further opinionated). It is an overlaying tooltip after all. 
But just to be of consideration of suggestions, I have moved it to the other side.

## Links
[MINI-629](https://jira.rakuten-it.com/jira/browse/MINI-629)

# Checklist
- [x] wrote/updated tests for new/changed code


ts-nitish-a-jain:
> Agreed with you.
> We should ask the issue reporter if he is okay with this solution, though.

You're right @minh-rakuten. I checked with Ashton and mentioned to him about the movement of the display button a bit.

PR #67, opened 2020-04-08
fix: Ensure error messages from the server are passed along for 'MiniApp.create'

corycaywood:
# Description
Some error messages were getting swallowed because there was a redundant try catch in 'MiniAppInfoFetcher#getInfo'. The exceptions are already handled and wrapped as 'MiniAppSdkException' by the ApiClient, so there is no need to handle exceptions in the other class.

## Links
MINI-827

# Checklist
- [ x] I ran `./gradlew check` without errors

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppInfoFetcher.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppInfoFetcher.kt
@@ -5,17 +5,11 @@ import com.rakuten.tech.mobile.miniapp.api.UpdatableApiClient
 
 internal class MiniAppInfoFetcher(private var apiClient: ApiClient) : UpdatableApiClient {
 
+    @Throws(MiniAppSdkException::class)
     suspend fun fetchMiniAppList() = apiClient.list()
 
-    @Suppress("TooGenericExceptionCaught", "SwallowedException")
-    suspend fun getInfo(appId: String) = run {
-        try {
-            apiClient.fetchInfo(appId)
-        } catch (error: Exception) {
-            // If backend functions correctly, this should never happen
-            throw sdkExceptionForInternalServerError()
-        }
-    }
+    @Throws(MiniAppSdkException::class)
+    suspend fun getInfo(appId: String) = apiClient.fetchInfo(appId)
 
     override fun updateApiClient(apiClient: ApiClient) {
         this.apiClient = apiClient
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
@@ -37,11 +37,13 @@ internal class ApiClient @VisibleForTesting constructor(
         hostAppId = rasAppId
     )
 
+    @Throws(MiniAppSdkException::class)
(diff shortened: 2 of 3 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppInfoFetcher.kt
```diff
@@ -5,17 +5,11 @@ import com.rakuten.tech.mobile.miniapp.api.UpdatableApiClient
...
 internal class MiniAppInfoFetcher(private var apiClient: ApiClient) : UpdatableApiClient {
 
+    @Throws(MiniAppSdkException::class)
     suspend fun fetchMiniAppList() = apiClient.list()
 
-    @Suppress("TooGenericExceptionCaught", "SwallowedException")
-    suspend fun getInfo(appId: String) = run {
-        try {
-            apiClient.fetchInfo(appId)
-        } catch (error: Exception) {
```
ts-nitish-a-jain, line 14: I think catching specific `NoSuchElementException` might suffice. What do you think?

minh-rakuten: Do you mean catching `NoSuchElementException` in `getInfo` or in `apiClient`?

ts-nitish-a-jain: in `getInfo`

corycaywood: Oh, in the case that it can't return `first` item? Hmm, maybe that should be handled in the ApiClient instead - we can check if the list is empty and return a `MiniAppSdkException` if so.

ts-nitish-a-jain: Yeah for the first item. Until now, the handling/intelligence was in `MiniAppInfoFetcher` which was inline to `MiniAppDownloader` where manifest is also checked for. It's a very boundary-types one.

corycaywood: I have updated this. I think in this case it makes more sense to check in `ApiClient` because we can directly check if the list is empty in that class.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -37,14 +37,22 @@ internal class ApiClient @VisibleForTesting constructor(
...
+    @Throws(MiniAppSdkException::class)
     suspend fun fetchInfo(appId: String): MiniAppInfo {
         val request = appInfoApi.fetchInfo(hostAppId, hostAppVersionId, appId)
-        return requestExecutor.executeRequest(request).first()
+        val info = requestExecutor.executeRequest(request)
+
+        if (info.isNotEmpty()) {
+            return info.first()
+        } else {
+            throw MiniAppSdkException("Server returned no info for the Mini App Id: $appId")
```
ts-nitish-a-jain, line 54: Having it return the real error is good.
But in reality, this would only happen if there's some internal serval error.

PR #68, opened 2020-04-08
fix: handling bad id from configuration

minh-rakuten:
# Description
Entering a bad guid should throw an error by calling api and should not make changes to the list. 

## Links
[MINI-820](https://jira.rakuten-it.com/jira/browse/MINI-820)

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuBaseActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuBaseActivity.kt
@@ -3,9 +3,13 @@ package com.rakuten.tech.mobile.testapp.ui.settings
 import android.os.Bundle
 import android.view.*
 import android.widget.EditText
+import android.widget.Toast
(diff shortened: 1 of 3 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuBaseActivity.kt
```diff
@@ -42,25 +49,55 @@ abstract class SettingsMenuBaseActivity: BaseActivity() {
...
+                runOnUiThread {
+                    pb.visibility = View.GONE
+                    recreate()
+                }
+            } catch (error: MiniAppSdkException) {
+                settings.appId = appIdHolder
+                settings.subscriptionKey = subscriptionKeyHolder
+                runOnUiThread {
+                    pb.visibility = View.GONE
+                    Toast.makeText(this@SettingsMenuBaseActivity, error.message, Toast.LENGTH_LONG)
```
corycaywood, line 97: I think the toast message is difficult to see because it displays on top of the keyboard. Maybe you can change the position so it's displayed at the top of the screen?

PR #74, opened 2020-04-24
update: add new bridge js, uuid for unique id in sample.

minh-rakuten:
# Description
- Update bridge js asset due to this [update](https://github.com/rakutentech/ios-miniapp/pull/63) from js sdk.
- Update color accent from ios.
- Generating uuid in sample app.

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
@@ -3,8 +3,8 @@ package com.rakuten.tech.mobile.testapp.ui.display
 import android.content.Context
 import android.content.Intent
 import android.content.pm.ApplicationInfo
-import android.os.Build
 import android.os.Bundle
+import android.provider.Settings
 import android.view.View
 import android.webkit.WebView
 import android.widget.Toast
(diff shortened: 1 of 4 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -62,7 +62,7 @@ class MiniAppDisplayActivity : BaseActivity() {
 
             launch {
                 viewModel.obtainMiniAppView(appId, object: MiniAppMessageBridge() {
-                    override fun getUniqueId() = "example_unique_id"
+                    override fun getUniqueId() = UUID.nameUUIDFromBytes(Build.DEVICE.toByteArray()).toString()
```
Climbatize, line 65: people with the same phone will have the UUID generated from the same array ?

minh-rakuten: That's right. There is no authentication for different users in our sample app at the moment so the sample app uses uuid generated from device.

corycaywood: Maybe we could use the Android ID instead? https://developer.android.com/reference/android/provider/Settings.Secure#ANDROID_ID

minh-rakuten:
> LGTM, except my comment, but it's not that important as it's in the demo app

@Climbatize You're right so I've changed into ANDROID_ID.

PR #76, opened 2020-05-01
Feature/custom scheme

minh-rakuten:
# Description
- Using custom scheme loader for webview.

## Links
[MINI-756](https://jira.rakuten-it.com/jira/browse/MINI-756)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
@@ -1,0 +1,28 @@
+package com.rakuten.tech.mobile.miniapp.display
+
+import android.content.Context
+import android.webkit.WebChromeClient
+import android.webkit.WebView
+import androidx.annotation.VisibleForTesting
+import java.io.BufferedReader
+
+internal class MiniAppWebChromeClient(context: Context) : WebChromeClient() {
+    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    @VisibleForTesting
+    internal val bridgeJs = try {
+        val inputStream = context.assets.open("bridge.js")
+        inputStream.bufferedReader().use(BufferedReader::readText)
+    } catch (e: Exception) {
+        null
+    }
+
+    override fun onReceivedTitle(webView: WebView, title: String?) {
+        doInjection(webView)
+        super.onReceivedTitle(webView, title)
+    }
+
+    @VisibleForTesting
+    internal fun doInjection(webView: WebView) {
+        webView.evaluateJavascript(bridgeJs) {}
+    }
+}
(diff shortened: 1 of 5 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -110,13 +113,22 @@ internal class MiniAppWebViewClient(
     override fun shouldInterceptRequest(
         view: WebView,
         request: WebResourceRequest
-    ): WebResourceResponse? = loader.shouldInterceptRequest(request.url)
+    ): WebResourceResponse? {
+        if (request.url != null && request.url.toString().startsWith(customScheme)) {
+            // Do js injection when index is loaded and at least one resource is requested to trigger this.
+            doInjection(view, request.url.toString())
```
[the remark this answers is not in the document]
minh-rakuten: You're right. Let me try with better solution / your suggestion.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -0,0 +1,28 @@
...
+    @VisibleForTesting
+    internal val bridgeJs = try {
+        val inputStream = context.assets.open("bridge.js")
+        inputStream.bufferedReader().use(BufferedReader::readText)
+    } catch (e: Exception) {
+        null
+    }
+
+    override fun onReceivedTitle(webView: WebView, title: String?) {
+        doInjection(webView)
```
corycaywood, line 20: Nice! Looks like this one worked.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -30,7 +25,9 @@ internal class RealMiniAppDisplay(
     miniAppMessageBridge: MiniAppMessageBridge
 ) : MiniAppDisplay, WebView(context), WebViewListener {
 
-    private val miniAppDomain = "$appId.$ASSET_DOMAIN_SUFFIX"
+    private val miniAppDomain = "mscheme.$appId"
```
corycaywood, line 28: I'm still not really sure about this scheme name because it doesn't really communicate what it's for. Is everyone okay with this?

I'm not sure what would be better though. If we use `miniapp.$appId`, then the URL has `miniapp` twice...

minh-rakuten: @ts-nitish-a-jain @ts-leojoseph-r @Climbatize 
We have discussed but are you really okay with this scheme format: `mscheme.MINI_APP_ID://miniapp/index.html`?
Naming is always the hard decision 😭

Climbatize: I did not like this naming from the beginning, but it's not that bad. If we decide to keep it like this it's not a big issue.

PR #80, opened 2020-05-13
Refactor/ios parity

minh-rakuten:
# Description
Keep the parity of downloading mini app flow between android & ios.

1/ Loading miniapp from MiniAppInfo -> 2/ Call api get Manifest files -> 3/ Load from local or Download from server

Change in android:
Removing unnecessary `isLatestVersion` function because the server already handled this.
Only catch and throw an exception from the server response. This also improves the loading speed when getting miniapp.

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -19,11 +19,11 @@ This SDK supports Android API level 21 (Lollipop) and above.
 ```groovy
 repositories {
     jcenter()
-    maven { url 'http://oss.jfrog.org/artifactory/simple/libs-snapshot/' } // If you want to use snapshot releases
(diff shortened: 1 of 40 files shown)


ts-nitish-a-jain:
@minh-rakuten Could you please link the "back end update discussion" where it has been updated?

minh-rakuten:
> @minh-rakuten Could you please link the "back end update discussion" where it has been updated?

@ts-nitish-a-jain There is no comment discussion from backend but it's good to implement the same flow as iOS.
I also tested with api and the backend already solved this case.

minh-rakuten:
> Let's hook up the discussion or API spec change before we commit this.

Okay. @ts-nitish-a-jain Please take your time on getting confirmation from BE.

minh-rakuten:
iOS will update which is following android. Both platforms accept slower loading time due to the agreement on design flow 😑

PR #82, opened 2020-05-19
Feature/offline loading

minh-rakuten:
# Description

- Loading miniapp in offline mode.
When there is no network but the mini had been already downloaded then run it from local. Otherwise, throwing the usual exception.

- Implement `Parcelable` for `MiniAppInfo`. 

## Links
[MINI-948](https://jira.rakuten-it.com/jira/browse/MINI-948)

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -20,23 +20,31 @@ internal class MiniAppDownloader(
...
-            return apiClient.fetchInfo(appId).version.versionId == versionId
-        } catch (e: Exception) {
-            // If backend functions correctly, this should never happen
-            throw sdkExceptionForInternalServerError()
+            return when {
+                !isLatestVersion(appId, versionId) -> throw sdkExceptionForInvalidVersion()
+                miniAppStatus.isVersionDownloaded(
+                    appId,
+                    versionId,
+                    storage.getMiniAppVersionPath(appId, versionId)
```
corycaywood, line 31: The version path is used several times here. Maybe you can set it in a `val versionPath` before the `try/catch`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -20,23 +20,31 @@ internal class MiniAppDownloader(
...
-    private suspend fun isLatestVersion(appId: String, versionId: String): Boolean {
+    @Suppress("SwallowedException", "LongMethod")
+    suspend fun getMiniApp(appId: String, versionId: String): String {
         try {
-            return apiClient.fetchInfo(appId).version.versionId == versionId
-        } catch (e: Exception) {
-            // If backend functions correctly, this should never happen
-            throw sdkExceptionForInternalServerError()
+            return when {
+                !isLatestVersion(appId, versionId) -> throw sdkExceptionForInvalidVersion()
```
corycaywood, line 27: So every time that the user tries to create a Mini App, we're calling the API to find out what the latest version is. And then if the version from the API doesn't match, we return an error.

Now I'm thinking that because we always fetch the latest version ID anyways, would it make more sense to just always create the Mini App using the version ID which was just fetched? And then we wouldn't need to pass the version Id here anymore.

I'm also thinking about how we will implement the test versions of Mini Apps in the future. It will be possible for there to be multiple version ids published for test versions, so we will still need the ability to fetch a specify version id. So maybe in the future (when we implement test mini app version) we can split this method into two: 
`suspend fun getMiniApp(appId: String, versionId: String): String` 
and `suspend fun getMiniApp(appId: String): String`

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/api/RetrofitRequestExecutorSpec.kt
```diff
@@ -201,6 +198,48 @@ open class RetrofitRequestExecutorErrorSpec : RetrofitRequestExecutorSpec() {
         executor.executeRequest(request)
     }
 
+    @Test(expected = MiniAppNetException::class)
+    fun `should throw exception when there is network error`() = runBlockingTest {
+        val executor = spyRetrofitExecutor()
+        val request: Call<String> = object : Call<String> {
```
corycaywood, line 204: Couldn't you do this mocking instead?

```
val request: Call<String> = mock()
When calling request.execute() itThrows SocketTimeoutException()
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
minh-rakuten: Agree. Actually, I had implemented the one always getting the latest version from this [PR](https://github.com/rakutentech/android-miniapp/pull/80) but this PR was closed due to the design concept from @ts-nitish-a-jain. Checking latest version function is redundant here.

If there are two different signatures `getMiniApp` then it makes sense because it does not break the original concept. We still can use `isLatestVersion(appId, versionId)` for getting test version. I need our consent here about this @ts-nitish-a-jain @corycaywood

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/api/RetrofitRequestExecutorSpec.kt
minh-rakuten: If you do mocking in function to throw exception, you must declare the throwing exception in method. For example, 
```
@Throws(SocketTimeoutException::class)
suspend fun <T> executeRequest(call: Call<T>)
```
but I guess we don't want to change this in our source code.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
ts-nitish-a-jain: @minh-rakuten There will be some changes coming in with the test requirement perspective as well. Let's have a walkthrough on this in the meeting tomorrow. Also, the quoted behavior from Clayton was:

> check via the API if the current version on the device is the latest.
> Immediately open, if it is, or the request fails.
> Download the latest, if it is not, then open.
> Throw an error and close if the mini-app no longer exists - not to be confused with a network error.

minh-rakuten: After considering and checking on BE api, I think we still need to call api to find out the latest version. Because the sdk is getting miniapp from `MiniAppInfo` object so it's a little bit ambiguous when automatically getting the latest version.

If we have a second method as getting miniapp by `miniAppId` then it's a good idea to approach that way. I don't know we should change our method prototype or replace the existing one ( loading by `MiniAppInfo`). What do you think? @corycaywood 
This scenario is out of scope of this PR so let's move it on team channel discussion / another PR.

corycaywood: Sure, okay maybe we can revisit this when we implement the testing features.

miniapp/build.gradle
```diff
@@ -26,6 +26,10 @@ android {
     buildToolsVersion = androidBuildTool
 }
 
+androidExtensions {
+    experimental = true
```
Climbatize, line 30: why do you need experimental extensions ?

minh-rakuten: it's for using some annotations in Kotlin `@Parcelize`. Extending `Parcelable` requires methods implementation but this annotation implicitly handle it -> short code.

PR #86, opened 2020-05-26
feat: add capability for navigation to display unit based on back key

ts-nitish-a-jain:
# Description
Adds the capability to go backward onBackKeyPressed

## Links
[MINI-1094](https://jira.rakuten-it.com/jira/browse/MINI-1094)

# Checklist


ts-nitish-a-jain:
> Looks good 👍
> 
> But let's not merge this until after releasing v1.1.0. I'll mark this as "Request changes" for now so that it doesn't get merged on accident.

@corycaywood Thanks. It makes sense.

ts-nitish-a-jain:
> Because we support navigation on phone back button now so I think of adding `close` button in sample app 🤔

@minh-rakuten for the close button & more functions on navigation, we would need to check with Clayton, I created the PR out of a feasibility-check getting converted into code-implementation.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -1,5 +1,6 @@
 package com.rakuten.tech.mobile.miniapp
 
+import android.util.Log
```
ts-nitish-a-jain, line 3: @minh-rakuten I think the branch that you baselined it with is a different one.

minh-rakuten: Oops, sorry 🙇 . The same baseline but forgot to remove logger codes.
Removed and totally checked.

PR #88, opened 2020-05-27
Fix/miniapp storage

minh-rakuten:
# Description
Fix data storage separation when loading custom scheme by Redirecting to custom http url.

## Links
[MINI-1057](https://jira.rakuten-it.com/jira/browse/MINI-1057)

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -17,11 +17,17 @@ internal class MiniAppWebViewClient(
...
+
+    override fun onReceivedError(
+        view: WebView,
+        request: WebResourceRequest,
+        error: WebResourceError
+    ) {
         if (request.url != null && request.url.toString().startsWith(customScheme)) {
-            val interceptUri = request.url.toString().replace(customScheme, customDomain).toUri()
-            return loader.shouldInterceptRequest(interceptUri)
+            view.loadUrl(request.url.toString().replace(customScheme, customDomain))
```
corycaywood, line 28: I was thinking, in `RealMiniAppDisplay` should we also change it to directly load the `https` url instead of the custom scheme? That way we don't have to hit an error on the initial load. 

I know that then it seems kind of pointless to have the `mscheme` at all, but the main purpose of the custom scheme is for having a consistent callback URI across platforms.

minh-rakuten:
> And I had another thought: should we change back to the old method for injecting the JS using `onLoadResource`? Basically revert this change: [d023dae#diff-74446ea3a44cad4526b6746288e3eac8L115](https://github.com/rakutentech/android-miniapp/commit/d023dae754c0cf084cf5d8f43c666825a1b232b1#diff-74446ea3a44cad4526b6746288e3eac8L115)
> 
@corycaywood `onLoadResource` requires at least one resource to be loaded and our unique id sample apps seem like not loading with this.

How about `onPageStarted`? I just figured out that we cannot do inject on `onPageStarted` on emulator which has no Google Play libs but I believe they're corrupted OS (no Google libs).

minh-rakuten:
I just updated loading with https first and `onPageStarted`. 
Because we add this handling `addPathHandler("/", ...)`, it swallowed the icon loading so `onLoadResource` is not called on our unique id sample apps.

corycaywood:
> How about `onPageStarted`? I just figured out that we cannot do inject on `onPageStarted` on emulator which has no Google Play libs but I believe they're corrupted OS (no Google libs). The normal devices are fine.

Oh, nice! Thanks for finding this. I think `onPageStarted` is the best solution if it's working on normal devices, so let's go with that. I wonder why it's not working on the "Google APIs" emulators 🤔 Maybe the WebView behavior is more restricted on those emulators since it's technically a rooted device...

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -80,10 +79,16 @@ internal class RealMiniAppDisplay(
...
+            "/", WebViewAssetLoader.InternalStoragePathHandler(
+                context,
+                File(basePath)
+            )
+        )
         .build()
 
     @VisibleForTesting
-    internal fun getLoadUrl() = "$customScheme$SUB_DOMAIN_PATH/index.html"
+    internal fun getLoadUrl() = "$customDomain$SUB_DOMAIN_PATH/index.html"
```
corycaywood, line 91: Now we again have this question of what should the HTTPS url be? Right now it would be something like `https://mscheme.MINI_APP_ID/miniapp/index.html`. Probably that's okay for now since we're not intending for users to use the HTTPS URL (they should use the custom scheme for redirecting to the mini app). But we'll need to change this URL in the future when we possibly replace the custom scheme on iOS with HTTPS.

minh-rakuten: Yeah. In the past, we had `https://MINI_APP_ID.miniapps.androidplatform.net/miniapp/index.html` but this `androidplatform.net` is not required so it could be `https://miniapps.MINI_APP_ID/miniapp/index.html` 🤷‍♂️.
How do you think @ts-nitish-a-jain @corycaywood?

corycaywood: Probably it's okay to keep as it is, and then we can update it in a later release (since we haven't yet decided what we're doing on iOS).

ts-nitish-a-jain: Yeah, I think let's keep it `https://mscheme.MINI_APP_ID/miniapp/index.html` for now. Thanks @minh-rakuten @corycaywood

PR #89, opened 2020-05-28
refactor: load view with provided context

minh-rakuten:
# Description
The concept is only creating webview when calling `MiniAppDisplay.getMiniAppView(Context)`
- Load webview with provided context.
- Update the flow of lifecycle registration.

We can also achieve this goal in a simpler way by updating `create` method. Please do the evaluation to see which one is better.

## Links
[MINI-1020](https://jira.rakuten-it.com/jira/browse/MINI-1020)
Another PR by updating`MiniApp.create`:
https://github.com/rakutentech/android-miniapp/pull/91

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppDisplay.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppDisplay.kt
@@ -1,7 +1,12 @@
-package com.rakuten.tech.mobile.miniapp
+package com.rakuten.tech.mobile.miniapp.display
 
+import android.content.Context
 import android.view.View
+import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleObserver
+import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageBridge
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
 
 /**
  * This represents the contract by which the host app can interact with the
@@ -9,14 +14,23 @@ import androidx.lifecycle.LifecycleObserver
  * This contract complies to Android's [LifecycleObserver] contract, and when made to observe
  * the lifecycle, it automatically clears up the view state and any services registered with.
  */
-interface MiniAppDisplay : LifecycleObserver {
+interface MiniAppDisplay: LifecycleObserver {
+    val basePath: String
+    val appId: String
+    val miniAppMessageBridge: MiniAppMessageBridge
 
     /**
      * Provides the view associated with the mini app to the caller for showing the mini app.
+     * Register the lifecycle as optional
      * @return [View] as mini app's view with [LayoutParams] set to match
      * the parent's dimensions.
      */
-    fun getMiniAppView(): View
+    suspend fun getMiniAppView(context: Context, hostLifecycle: Lifecycle? = null): View =
+        withContext(Dispatchers.Main) {
+            val view = RealMiniAppDisplay(context, basePath, appId, miniAppMessageBridge)
+            hostLifecycle?.addObserver(view)
+            view
+        }
 
     /**
      * Upon invocation, destroys necessary view state and any services registered with.
@@ -28,5 +42,5 @@ interface MiniAppDisplay : LifecycleObserver {
      * on the lifecycle of views e.g. removal of the view from the view system, yet
      * within the same state of parent's lifecycle.
      */
-    fun destroyView()
+    fun destroyView() = Unit
 }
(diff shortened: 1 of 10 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppDisplay.kt
```diff
@@ -1,22 +1,35 @@
...
+    val miniAppMessageBridge: MiniAppMessageBridge
 
     /**
      * Provides the view associated with the mini app to the caller for showing the mini app.
+     * [LifecycleObserver] as optional. When made to observe the lifecycle,
+     * it automatically clears up the view state and any services registered with.
      * @return [View] as mini app's view with [LayoutParams] set to match
      * the parent's dimensions.
      */
-    fun getMiniAppView(): View
```
corycaywood, line 19: Is there a way we can keep the old `getMiniAppView` method as well as well and deprecate it? I guess it would keeping a context passed from the Application, so things would get complicated.

Maybe we can release a major version for the next release? Then we don't have to worry as much about minor changes to the interface.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppDisplay.kt
```diff
@@ -1,22 +1,35 @@
...
+    suspend fun getMiniAppView(context: Context, hostLifecycle: Lifecycle? = null): View =
+        withContext(Dispatchers.Main) {
+            val view = RealMiniAppDisplay(context, basePath, appId, miniAppMessageBridge)
```
corycaywood, line 29: This will recreate the view every time that they call `getMiniAppView`. Could this cause problems? For example, wouldn't they expect it to always return the same instance if they call it multiple times? Though I can't think of any examples why they would want to call `getMiniAppView` more than once...

I know that I suggested passing the context at `getMiniAppView`, but now I'm wondering if it would be better to pass the context at `MiniApp#create`? It seems like it would solve some potential issues. What do you think @ts-nitish-a-jain @minh-rakuten ?

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -24,24 +26,26 @@ import org.junit.runner.RunWith
 import org.mockito.Mockito
 import kotlin.test.assertTrue
 
+@ExperimentalCoroutinesApi
 @RunWith(AndroidJUnit4::class)
 class RealMiniAppDisplayTest {
     private lateinit var context: Context
-    private lateinit var basePath: String
+    private lateinit var mBasePath: String
```
corycaywood, line 33: Let's keep this as `basePath`. We aren't using hungarian notation anywhere else, so we need to stay consistent.

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplayTest.kt
```diff
@@ -24,24 +26,26 @@ import org.junit.runner.RunWith
...
+    private lateinit var miniAppDisplay: MiniAppDisplay
     private lateinit var realDisplay: RealMiniAppDisplay
     private lateinit var webResourceRequest: WebResourceRequest
-    private val miniAppMessageBridge: MiniAppMessageBridge = mock()
+    private val mMiniAppMessageBridge: MiniAppMessageBridge = mock()
```
corycaywood, line 37: Let's also keep this one as `miniAppMessageBridge`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppDisplay.kt
```diff
@@ -1,22 +1,35 @@
...
 
     /**
      * Provides the view associated with the mini app to the caller for showing the mini app.
+     * [LifecycleObserver] as optional. When made to observe the lifecycle,
+     * it automatically clears up the view state and any services registered with.
      * @return [View] as mini app's view with [LayoutParams] set to match
      * the parent's dimensions.
      */
-    fun getMiniAppView(): View
+    suspend fun getMiniAppView(context: Context, hostLifecycle: Lifecycle? = null): View =
```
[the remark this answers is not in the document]
minh-rakuten: The argument `context` and `hostLifecycle` are not related here. The old method does not work in this situation because `hostLifecycle` observes the `RealMiniAppDisplay` which is created in this function `getMiniAppView(Context, LicycleObserver`).

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppDisplay.kt
minh-rakuten: I think I'll create another PR which targets at `MiniApp#create` so that we can compare. I agree that using `getMiniAppView` reveal issue such as `MiniAppDisplay` exposes the `basePath`.

minh-rakuten:
This [PR](https://github.com/rakutentech/android-miniapp/pull/90) solution had been merged.

PR #90, opened 2020-05-30
base commit for activity's context injection into the mini app display

ts-nitish-a-jain:
# Description
Base commit for showing the activity's context injection into the mini-app display.

Note: CI would fail at this moment because of code refactoring and breaking of unit tests.

Updated: 
- Retain `ApplicationContext` in `Displayer` to maintain compatability, will omit it in the next release.
- `getMiniAppView()` is labeled as deprecated function.
- Update test cases.

# Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext.androidBuildTool = '29.0.2'
-    ext.androidx_appcompat = '1.1.0'
+    ext.androidx_appcompat = '1.2.0-rc01'
     ext.androidx_constraintLayout = '1.1.3'
     ext.androidx_coreKtx = '1.2.0'
     ext.androidx_lifecycle = '2.2.0'
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,8 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext.androidBuildTool = '29.0.2'
+    // AndroidX Appcompat 1.1.0 results in crash on Android 5
+    // https://issuetracker.google.com/issues/141132133
     ext.androidx_appcompat = '1.2.0-rc01'
     ext.androidx_constraintLayout = '1.1.3'
     ext.androidx_coreKtx = '1.2.0'
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
@@ -18,10 +18,10 @@ interface MiniAppDisplay : LifecycleObserver {
      * the parent's dimensions.
      *
      * This version of retrieval creates the mini app view which doesn't render some of the native
-     * elements of web tech. Using the updated API as recommended.
+     * elements of web tech. We recommend switching to the new API.
      */
     @Deprecated(message = "Please replace with getMiniAppView(Context)")
-    suspend fun getMiniAppView(): View?
+    suspend fun getMiniAppView(): View
 
     /**
      * Provides the view associated with the mini app to the caller for showing the mini app.
@@ -29,6 +29,7 @@ interface MiniAppDisplay : LifecycleObserver {
      * Should be the context of activity to ensure that all standard html components work properly.
      * @return [View] as mini app's view with [LayoutParams] set to match
      * the parent's dimensions.
+     * @throws MiniAppSdkException when a non-matching context is supplied
      */
     suspend fun getMiniAppView(activityContext: Context): View?
 
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
@@ -16,18 +16,21 @@ interface MiniAppDisplay : LifecycleObserver {
      * Provides the view associated with the mini app to the caller for showing the mini app.
      * @return [View] as mini app's view with [LayoutParams] set to match
      * the parent's dimensions.
+     *
+     * This version of retrieval creates the mini app view which doesn't render some of the native
+     * elements of web tech. Using the updated API as recommended.
      */
     @Deprecated(message = "Please replace with getMiniAppView(Context)")
-    suspend fun getMiniAppView(): View
+    suspend fun getMiniAppView(): View?
 
     /**
      * Provides the view associated with the mini app to the caller for showing the mini app.
-     * @param providedContext is used by the view for initializing the internal services.
+     * @param activityContext is used by the view for initializing the internal services.
      * Should be the context of activity to ensure that all standard html components work properly.
      * @return [View] as mini app's view with [LayoutParams] set to match
      * the parent's dimensions.
      */
-    suspend fun getMiniAppView(providedContext: Context): View?
+    suspend fun getMiniAppView(activityContext: Context): View?
 
     /**
      * Upon invocation, destroys necessary view state and any services registered with.
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
@@ -1,5 +1,6 @@
 package com.rakuten.tech.mobile.miniapp
 
+import android.content.Context
 import android.view.View
 import androidx.lifecycle.LifecycleObserver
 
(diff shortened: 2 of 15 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/Displayer.kt
```diff
@@ -1,21 +1,18 @@
...
 
-internal class Displayer(private val context: Context) {
+internal class Displayer {
 
     suspend fun createMiniAppDisplay(
         basePath: String,
         appId: String,
         miniAppMessageBridge: MiniAppMessageBridge
     ): MiniAppDisplay =
         withContext(Dispatchers.Main) {
```
minh-rakuten, line 15: Because not creating webview so does not need to put in this looper anymore. Just only:
```suggestion
        context?.let {
                RealMiniAppDisplay(basePath, appId, miniAppMessageBridge)
            }
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -1,97 +1,37 @@
...
-        settings.domStorageEnabled = true
-        settings.databaseEnabled = true
-        webViewClient = MiniAppWebViewClient(context, getWebViewAssetLoader(), customDomain, customScheme)
-
-        loadUrl(getLoadUrl())
-    }
-
-    override fun getMiniAppView(): View = this
+    override fun getMiniAppView(activityContext: Context): View? =
+        if (activityContext is Activity || activityContext is ActivityCompat) {
```
minh-rakuten, line 23: I guess we should let the hostapps provide any kind of `context` they want. In documentation, we recommend using `ActivityContext` with the reasonable explaination. 

Furthermore, using `ActivityContext` causes crash on Android 5 (maybe even 6) so you need to update the `androidx.appcompat` like [this](https://github.com/rakutentech/android-miniapp/pull/89/commits/7602ff5965a6087b3938f6c8e8a487f974d39aaa).

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -1,97 +1,37 @@
...
-
-        settings.allowUniversalAccessFromFileURLs = true
-        settings.domStorageEnabled = true
-        settings.databaseEnabled = true
-        webViewClient = MiniAppWebViewClient(context, getWebViewAssetLoader(), customDomain, customScheme)
-
-        loadUrl(getLoadUrl())
-    }
-
-    override fun getMiniAppView(): View = this
```
minh-rakuten, line 49: We still need to keep this method before deprecating it. I suggest that we still keep `ApplicationContext` in `RealMiniAppDisplay` for creating `MiniAppWebView`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -1,97 +1,37 @@
...
+    override fun getMiniAppView(activityContext: Context): View? =
+        if (activityContext is Activity || activityContext is ActivityCompat) {
+            (miniAppWebView ?: MiniAppWebView(
```
minh-rakuten, line 24: Creating `WebView` requires to be put in `withContext(Dispatchers.Main) {}`.

build.gradle
```diff
@@ -1,7 +1,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 buildscript {
     ext.androidBuildTool = '29.0.2'
-    ext.androidx_appcompat = '1.1.0'
+    ext.androidx_appcompat = '1.2.0-rc01'
```
ts-nitish-a-jain, line 4: Unless blocking, let's try to use only the stable versions.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -16,7 +17,17 @@ interface MiniAppDisplay : LifecycleObserver {
...
+    suspend fun getMiniAppView(): View
+
+    /**
+     * Provides the view associated with the mini app to the caller for showing the mini app.
+     * @param providedContext is used by the view for initializing the internal services.
+     * Should be the context of activity to ensure that all standard html components work properly.
+     * @return [View] as mini app's view with [LayoutParams] set to match
+     * the parent's dimensions.
+     */
+    suspend fun getMiniAppView(providedContext: Context): View?
```
ts-nitish-a-jain, line 30: Let's keep the name of the param explicitly `activityContext`.

build.gradle
minh-rakuten: The stable version `1.1.0` has a [bug](https://stackoverflow.com/questions/58028821/webview-crash-on-android-5-5-1-api-21-22-resourcesnotfoundexception-string-r) causing crash on android 5 when passing activity context.  We have only option to update the library.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
minh-rakuten: Hmmm I think the hostapp can provide any context they want, though, `ApplicationContext` causes webview not working properly. They may even provide the context from `Service`.

build.gradle
ts-nitish-a-jain: I read your previous comment just now. Thanks @minh-rakuten

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
ts-nitish-a-jain: That's why we want to tell the developers explicitly to only provide `activityContext` at this stage, and in SDK we are validating if it is of type `Activity` or `ActivityCompat` otherwise returning null for the new API.

ts-nitish-a-jain: We could add in the docs:
`This version of retrieval creates the mini app view which doesn't render some of the native elements of web tech. We recommend, using the updated API.`

minh-rakuten: Updated.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -1,97 +1,57 @@
...
-        settings.allowUniversalAccessFromFileURLs = true
-        settings.domStorageEnabled = true
-        settings.databaseEnabled = true
-        webViewClient = MiniAppWebViewClient(context, getWebViewAssetLoader(), customDomain, customScheme)
+    @VisibleForTesting
+    internal var miniAppWebView: MiniAppWebView? = null
 
-        loadUrl(getLoadUrl())
-    }
+    override suspend fun getMiniAppView(): View? = provideMiniAppWebView(context)
```
corycaywood, line 27: Actually, does this need to be a `suspend` function at all? It's just instantiating the `MiniAppWebView` class. `provideMiniAppWebView` is using the `Dispatchers.Main` context anyways and the App should already be calling it on the main thread since it returns a view. Is there any problem with removing the `suspend`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -1,97 +1,57 @@
...
+                withContext(Dispatchers.Main) {
+                    MiniAppWebView(
+                        context = activityContext,
+                        basePath = basePath,
+                        appId = appId,
+                        miniAppMessageBridge = miniAppMessageBridge
+                    )
+                }
+            }
+        } else null
```
corycaywood, line 52: If they provided the wrong context, then this is a case where they've implemented the SDK wrong. So maybe this should throw an error instead of returning null? It would allow `getMiniAppView` to have a non-null return type, and it would give them a helpful error message to see what they've done wrong. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
minh-rakuten: We need the `suspend` for `Dispatchers.Main`. `Dispatchers.Main` is required here for rendering the view because the hostapp can do `launch` with any `CoroutineScope` they want (not in main thread anymore).

For example, our sample app calling `launch {}` means launching with `Dispatchers.Default`. It's safer to specify `Dispatchers.Main`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
minh-rakuten: I think we should keep it nullable here for `destroyView` purpose 🤔

build.gradle
minh-rakuten: Updated on USERGUIDE.md

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
corycaywood: Yeah, good point. The old method should continue to function for now.

PR #94, opened 2020-06-09
Feature/build_version_display & update_get_latest_miniapp

minh-rakuten:
# Description
- Display build information in the setting menu. The build version is generated in `build.gradle` of testapp. There will the appending in apk build Version name`-build-$buildVersion`.

- From discussion & agreement, update sdk as loading miniapp with latest version. `IsLatestVersion()` is omitted.

## Links
[MINI-921](https://jira.rakuten-it.com/jira/browse/MINI-921)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -36,8 +36,7 @@ internal class MiniAppDownloader(
                 val versionPath = storage.getMiniAppVersionPath(appId, versionId)
                 if (miniAppStatus.isVersionDownloaded(appId, versionId, versionPath))
                     return versionPath
-            } else
-                throw netError
+            }
         }
         // cannot load miniapp from server
         throw sdkExceptionForInternalServerError()
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -20,7 +20,7 @@ internal class MiniAppDownloader(
 ) : UpdatableApiClient {
 
     @Suppress("SwallowedException", "LongMethod")
-    suspend fun getMiniApp(appId: String, versionId: String): String {
+    suspend fun getMiniApp(appId: String): String {
         try {
             val miniAppInfo = apiClient.fetchInfo(appId)
             val versionPath = storage.getMiniAppVersionPath(miniAppInfo.id, miniAppInfo.version.versionId)
@@ -31,9 +31,13 @@ internal class MiniAppDownloader(
             }
         } catch (netError: MiniAppNetException) {
             // load local if possible when offline
-            val versionPath = storage.getMiniAppVersionPath(appId, versionId)
-            if (miniAppStatus.isVersionDownloaded(appId, versionId, versionPath))
-                return versionPath
+            val versionId = miniAppStatus.localVersion
+            if (versionId != null) {
+                val versionPath = storage.getMiniAppVersionPath(appId, versionId)
+                if (miniAppStatus.isVersionDownloaded(appId, versionId, versionPath))
+                    return versionPath
+            } else
+                throw netError
         }
         // cannot load miniapp from server
         throw sdkExceptionForInternalServerError()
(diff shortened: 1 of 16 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -21,25 +21,24 @@ internal class MiniAppDownloader(
...
-                !isLatestVersion(appId, versionId) -> throw sdkExceptionForInvalidVersion()
-                miniAppStatus.isVersionDownloaded(appId, versionId, versionPath) -> versionPath
-                else -> startDownload(appId, versionId)
+                miniAppStatus.isVersionDownloaded(
+                    miniAppInfo.id, miniAppInfo.version.versionId, versionPath) -> versionPath
+                else -> startDownload(miniAppInfo.id, miniAppInfo.version.versionId)
             }
         } catch (netError: MiniAppNetException) {
             // load local if possible when offline
+            val versionPath = storage.getMiniAppVersionPath(appId, versionId)
```
corycaywood, line 34: This means that the host app needs to know the last downloaded `versionId` if they want to launch a Mini App while offline. So they would need to cache the `MiniAppInfo` object?

I think there are two possible solutions to this:
1. Add another function: `MiniApp#create(miniAppId: String, miniAppMessageBridge: MiniAppMessageBridge)`. This function would do basically the same thing as the current `create` function, except that when the network is offline it would find the currently downloaded `versionId` for the Mini App and then load that. (We could also possibly deprecate the old `create` function - I think the original purpose of passing `MiniAppInfo` is that we thought there could be multiple versions of a Mini App published at the same time, but this isn't possible).
2. or Change `MiniApp#fetchInfo` so that it will return cached a cached `MiniAppInfo` object while the network is offline. But this would mean we need to keep this cache on the SDK side, so probably this option isn't ideal.

What do you think @minh-rakuten @ts-nitish-a-jain?

testapp/build.gradle
```diff
@@ -45,18 +45,23 @@ android {
             appId          : property("HOST_APP_PROD_ID") ?: "test-host-app-id",
             subscriptionKey: property("HOST_APP_PROD_SUBSCRIPTION_KEY") ?: "test-subs-key"
     ]
+
+    def buildVersion = new Date().format('yyMMddHHmm')
```
corycaywood, line 49: Maybe you could use the Circle CI build number (`CIRCLE_BUILD_NUM` env variable see https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables), and then use a timestamp only if the env variable isn't set? Something like `System.getenv("CIRCLE_BUILD_NUM") ?: new Date().format('yyMMddHHmm')`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
minh-rakuten: I had updated with a solution no.1 👍 and also declare MiniApp#create(MiniAppInfo, MiniAppMessageBridge) as deprecated.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStatus.kt
```diff
@@ -5,11 +5,17 @@ import android.content.SharedPreferences
...
+private const val LOCAL_VERSION = "local_version"
+
 internal class MiniAppStatus(context: Context) {
     private val prefs: SharedPreferences = context.getSharedPreferences(
         "com.rakuten.tech.mobile.miniapp.storage", Context.MODE_PRIVATE
     )
 
+    var localVersion: String?
+        get() = prefs.getString(LOCAL_VERSION, null)
+        set(downloadedVersion) = prefs.edit().putString(LOCAL_VERSION, downloadedVersion).apply()
```
corycaywood, line 17: I think I'm not understanding something here. This is aways saving the local version to the same preferences key, so it will overwrite the cached version regardless of the mini app id, right? So how would this work when multiple mini apps have been downloaded?

minh-rakuten: I'm sorry 🙏 
I have updated code for caching multiple apps.

corycaywood: Ah, okay no problem! I thought I was missing seeing something 😆

PR #97, opened 2020-06-15
fix: stop display error page

minh-rakuten:
# Description
- Fix custom scheme error redirection.
We had implemented the redirection with custom domain before by replacing custom scheme with http on `OnReceivedError`. However, in some webview versions, the behavior of displaying error is still there even the new redirection has been called. The solution is to stop display webview then wait after a while then finally calling redirection with http.

- Display sdk version sample app as a requirement from updated ticket.

## Links
[MINI-1118](https://jira.rakuten-it.com/jira/browse/MINI-1118)
[MINI-921](https://jira.rakuten-it.com/jira/browse/MINI-921)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
@@ -25,9 +25,19 @@ internal class MiniAppWebViewClient(
         error: WebResourceError
     ) {
         if (request.url != null && request.url.toString().startsWith(customScheme)) {
-            view.loadUrl(request.url.toString().replace(customScheme, customDomain))
+            loadWithCustomDomain(view, request.url.toString().replace(customScheme, customDomain))
             return
         }
         super.onReceivedError(view, request, error)
+    }
+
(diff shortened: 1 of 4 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -25,9 +25,19 @@ internal class MiniAppWebViewClient(
...
         super.onReceivedError(view, request, error)
     }
+
+    @Suppress("MagicNumber")
+    @VisibleForTesting
+    internal fun loadWithCustomDomain(view: WebView, requestUrl: String) {
+        view.stopLoading()
+        view.postDelayed(
+            { view.loadUrl(requestUrl) },
+            100
```
corycaywood, line 40: Is `stopLoading` asynchronous? Why was 100ms chosen as the delay? It seems like this still leaves the possibility that `loadUrl` will get called before the view has stopped loading.

I was thinking maybe you could use `onPageFinished` for this. For example, you could call `stopLoading` in `onReceivedError` and then use `onPageFinished` for calling `loadUrl`. Do you think that would work?

PR #98, opened 2020-06-15
Poc/share

minh-rakuten:
# Description
The snapshot release with `share` feature from miniapp. Only for a demo.

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -25,6 +28,11 @@ abstract class MiniAppMessageBridge {
         }
     }
 
+    @JavascriptInterface
+    fun shareTo(content: String) {
```
ts-nitish-a-jain, line 32: `shareToChat` sounds more natural. What do you think?

corycaywood, line 32: This is okay for POC purposes, but for the actual implementation `shareTo` should be passed through `postMessage` and included in the JS SDK.

PR #100, opened 2020-06-22
Feature: testing endpoint

minh-rakuten:
# Description
1/ Switching to test endpoint to see miniapp versions in Testing environment by adding Path on url:
MiniApp api: `.../test`
Manifest api: `.../test/manifest`

The data responses from server side are expected the same as they are now.

Update for the manifest configuration:
`IsTestMode`: Decide to run original api or testing one.

When this PR gets approved, require an update config on CI for this value:

- Staging
`IS_TEST_MODE`: true

- Production
`PROD_IS_TEST_MODE`: should be false. Production does not need testing miniapps.

2/ Update UI sample app
- Add toggle button to switch to Testing on Setting dialog.
- Display miniapp list as Section list. The section is miniapp name, items are miniapp versions & icons.

## Links
[MINI-235](https://jira.rakuten-it.com/jira/browse/MINI-235)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/AppInfoApi.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/AppInfoApi.kt
@@ -8,16 +8,17 @@ import retrofit2.http.Query
 
 internal interface AppInfoApi {
 
-    @GET("host/{hostappId}/miniapps")
+    @GET("host/{hostappId}/miniapps/{testPath}")
     fun list(
         @Path("hostappId") hostAppId: String,
+        @Path("testPath") testPath: String = "",
         @Query("hostVersion") hostAppVersionId: String
     ): Call<List<MiniAppInfo>>
 
-    @GET("host/{hostappId}/miniapps")
+    @GET("host/{hostappId}/miniapps/{testPath}")
     fun fetchInfo(
-        // for readability, maintainability & avoiding `null` default init for optional query param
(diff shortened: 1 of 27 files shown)


minh-rakuten:
Update using the same base url and adding Path on url:
MiniApp api: `.../test`
Manifest api: `.../test/manifest`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/AppInfoApi.kt
```diff
@@ -8,16 +8,17 @@ import retrofit2.http.Query
 
 internal interface AppInfoApi {
 
-    @GET("host/{hostappId}/miniapps")
+    @GET("host/{hostappId}/miniapps/{testPath}")
```
corycaywood, line 11: So this would add an extra `/` on the end of the path for the normal published endpoints? Maybe you could encapsulate the `/` in the value passed to `testPath`? i.e. passing `/test`.

minh-rakuten: I tested with `/test`, because it's going be retrofit api path, it will be encoded as `%2Ftest`. I'm not sure if it's okay for an api call so I will keep it like this for safe 😄  (an extra `/` in the end of the normal published endpoint).

corycaywood: Ah, okay that makes sense.

PR #102, opened 2020-06-29
Feature/geolocation

minh-rakuten:
# Description
Permission request & response between miniapp & hostapp.
The location request from JS SDK Sample App and our client app is `geolocation`.
This PR enables the miniapp using location from mobile via `geolocation` request.
- Handling geolocation prompt on `WebChromeClient`.
- Permission request & response by `MiniAppMessageBridge`.
  Single permission request is declared in this PR.
## Links
[MINI-914](https://jira.rakuten-it.com/jira/browse/MINI-914)
# Checklist

### DIFF
--- a/miniapp/src/main/AndroidManifest.xml
+++ b/miniapp/src/main/AndroidManifest.xml
@@ -4,6 +4,8 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
 
     <application>
         <provider
--- a/miniapp/src/main/AndroidManifest.xml
+++ b/miniapp/src/main/AndroidManifest.xml
@@ -4,8 +4,6 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
 
     <application>
         <provider
--- a/miniapp/src/main/assets/bridge.js
+++ b/miniapp/src/main/assets/bridge.js
@@ -94,6 +94,7 @@ MiniAppBridge.getUniqueId = function() {
   return new Promise(function(resolve, reject) {
     return MiniAppBridge.exec(
       "getUniqueId",
+      null,
       function(id) { return resolve(id) },
       function (error) { return reject(error) }
     );
--- a/miniapp/src/main/assets/bridge.js
+++ b/miniapp/src/main/assets/bridge.js
@@ -1,3 +1,4 @@
+
 var MiniAppBridge = {};
 MiniAppBridge.messageQueue = [];
 var uniqueId = Math.random();
@@ -15,10 +16,12 @@ var isPlatform = {
  * Method to call the native interface methods for respective platforms
  * such as iOS & Android
  * @param  {[String]} action Action command/interface name that native side need to execute
+ * @param  {Object} param Object that contains request parameter values like permissions.
+ * For eg., {permission: 'location'}
  * @param  {[Function]} onSuccess Success callback function
  * @param  {[Function]} onError Error callback function
  */
-MiniAppBridge.exec = function(action, onSuccess, onError) {
+MiniAppBridge.exec = function(action, param, onSuccess, onError) {
   var callback = {};
   callback.onSuccess = onSuccess;
   callback.onError = onError;
@@ -26,11 +29,11 @@ MiniAppBridge.exec = function(action, onSuccess, onError) {
   MiniAppBridge.messageQueue.unshift(callback);
   if (isPlatform.iOS()) {
     webkit.messageHandlers.MiniAppiOS.postMessage(
-      JSON.stringify({ action: action, id: callback.id })
+      JSON.stringify({ action: action, param: param, id: callback.id })
     );
   } else {
     window.MiniAppAndroid.postMessage(
-      JSON.stringify({ action: action, id: callback.id })
+      JSON.stringify({ action: action, param: param, id: callback.id })
     );
   }
 };
@@ -92,6 +95,21 @@ MiniAppBridge.getUniqueId = function() {
     return MiniAppBridge.exec(
       "getUniqueId",
       function(id) { return resolve(id) },
+      function (error) { return reject(error) }
+    );
+  });
+};
+
+/**
+ * Associating requestPermission function to MiniAppBridge object
+ * @param {String} permissionType Type of permission that is requested. For eg., location
+ */
+MiniAppBridge.requestPermission = function(permissionType) {
+  return new Promise(function(resolve, reject) {
+    return MiniAppBridge.exec(
+      "requestPermission",
+      {permission: permissionType},
+      function(success) { return resolve(success) },
       function (error) { return reject(error) }
     );
   });
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
@@ -28,21 +28,6 @@ abstract class MiniApp internal constructor() {
 
     /**
      * Creates a mini app.
-     * @param info metadata of a mini app.
-     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
-     * @param miniAppMessageBridge the interface for communicating between host app & mini app
-     * @throws MiniAppSdkException when there is some issue during fetching,
-     * downloading or creating the view.
-     */
-    @Throws(MiniAppSdkException::class)
-    @Deprecated(message = "Please replace with create(MiniAppId, MiniAppMessageBridge)")
-    abstract suspend fun create(
-        info: MiniAppInfo,
-        miniAppMessageBridge: MiniAppMessageBridge
-    ): MiniAppDisplay
-
-    /**
-     * Creates a mini app.
      * @param appId mini app id.
      * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
      * @param miniAppMessageBridge the interface for communicating between host app & mini app
@@ -54,18 +39,6 @@ abstract class MiniApp internal constructor() {
         appId: String,
         miniAppMessageBridge: MiniAppMessageBridge
     ): MiniAppDisplay
-
-    /**
-     * @deprecated use {@link #create(MiniAppInfo, MiniAppMessageBridge)} instead.
-     * Creates a mini app.
-     * @param info metadata of a mini app.
-     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
-     * @throws MiniAppSdkException when there is some issue during fetching,
-     * downloading or creating the view.
-     */
-    @Throws(MiniAppSdkException::class)
-    @Deprecated(message = "Please replace with create(MiniAppId, MiniAppMessageBridge)")
-    abstract suspend fun create(info: MiniAppInfo): MiniAppDisplay
 
     /**
      * Fetches meta data information of a mini app.
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
@@ -96,10 +96,11 @@ abstract class MiniApp internal constructor() {
         fun instance(settings: MiniAppSdkConfig = defaultConfig): MiniApp =
             instance.apply { updateConfiguration(settings) }
 
-        @Suppress("LongMethod")
+        @Suppress("LongMethod", "LongParameterList")
         internal fun init(
             context: Context,
             baseUrl: String,
+            isTestMode: Boolean,
             rasAppId: String,
             subscriptionKey: String,
             hostAppVersionId: String
@@ -108,13 +109,15 @@ abstract class MiniApp internal constructor() {
                 baseUrl = baseUrl,
                 rasAppId = rasAppId,
                 subscriptionKey = subscriptionKey,
-                hostAppVersionId = hostAppVersionId
+                hostAppVersionId = hostAppVersionId,
+                isTestMode = isTestMode
             )
             val apiClient = ApiClient(
                 baseUrl = baseUrl,
                 rasAppId = rasAppId,
                 subscriptionKey = subscriptionKey,
-                hostAppVersionId = hostAppVersionId
+                hostAppVersionId = hostAppVersionId,
+                isTestMode = isTestMode
             )
             val apiClientRepository = ApiClientRepository().apply {
                 registerApiClient(defaultConfig.key, apiClient)
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
@@ -28,6 +28,21 @@ abstract class MiniApp internal constructor() {
 
     /**
      * Creates a mini app.
+     * @param info metadata of a mini app.
+     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
+     * @param miniAppMessageBridge the interface for communicating between host app & mini app
+     * @throws MiniAppSdkException when there is some issue during fetching,
+     * downloading or creating the view.
+     */
+    @Throws(MiniAppSdkException::class)
+    @Deprecated(message = "Please replace with create(MiniAppId, MiniAppMessageBridge)")
+    abstract suspend fun create(
+        info: MiniAppInfo,
+        miniAppMessageBridge: MiniAppMessageBridge
+    ): MiniAppDisplay
+
+    /**
+     * Creates a mini app.
      * @param appId mini app id.
      * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
      * @param miniAppMessageBridge the interface for communicating between host app & mini app
@@ -39,6 +54,18 @@ abstract class MiniApp internal constructor() {
         appId: String,
         miniAppMessageBridge: MiniAppMessageBridge
     ): MiniAppDisplay
+
+    /**
+     * @deprecated use {@link #create(MiniAppInfo, MiniAppMessageBridge)} instead.
+     * Creates a mini app.
+     * @param info metadata of a mini app.
+     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
+     * @throws MiniAppSdkException when there is some issue during fetching,
+     * downloading or creating the view.
+     */
+    @Throws(MiniAppSdkException::class)
+    @Deprecated(message = "Please replace with create(MiniAppId, MiniAppMessageBridge)")
+    abstract suspend fun create(info: MiniAppInfo): MiniAppDisplay
 
     /**
      * Fetches meta data information of a mini app.
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
@@ -20,14 +20,36 @@ internal class RealMiniApp(
         else -> miniAppInfoFetcher.getInfo(appId)
     }
 
+    @Suppress("TooGenericExceptionThrown")
+    override suspend fun create(info: MiniAppInfo): MiniAppDisplay =
+        executingCreate(info.id, object : MiniAppMessageBridge() {
+            override fun getUniqueId(): String =
+                throw Exception("MiniAppMessageBridge has not been implemented")
+            override fun requestPermission(
+                callbackId: String,
+                miniAppPermissionType: String,
+                permissions: Array<String>
+            ) = throw Exception("MiniAppMessageBridge has not been implemented")
+        })
+
+    override suspend fun create(
+        info: MiniAppInfo,
+        miniAppMessageBridge: MiniAppMessageBridge
+    ): MiniAppDisplay = executingCreate(info.id, miniAppMessageBridge)
+
     override suspend fun create(
         appId: String,
         miniAppMessageBridge: MiniAppMessageBridge
+    ): MiniAppDisplay = executingCreate(appId, miniAppMessageBridge)
+
+    private suspend fun executingCreate(
+        miniAppId: String,
+        miniAppMessageBridge: MiniAppMessageBridge
     ): MiniAppDisplay = when {
-        appId.isBlank() -> throw sdkExceptionForInvalidArguments()
+        miniAppId.isBlank() -> throw sdkExceptionForInvalidArguments()
         else -> {
-            val basePath = miniAppDownloader.getMiniApp(appId)
-            displayer.createMiniAppDisplay(basePath, appId, miniAppMessageBridge)
+            val basePath = miniAppDownloader.getMiniApp(miniAppId)
+            displayer.createMiniAppDisplay(basePath, miniAppId, miniAppMessageBridge)
         }
     }
 
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
@@ -61,11 +61,11 @@ internal class RealMiniApp(
     }
 
     @VisibleForTesting
-    internal fun createApiClient(newConfig: MiniAppSdkConfig) =
-        ApiClient(
-            baseUrl = newConfig.baseUrl,
-            rasAppId = newConfig.rasAppId,
-            subscriptionKey = newConfig.subscriptionKey,
-            hostAppVersionId = newConfig.hostAppVersionId
-        )
+    internal fun createApiClient(newConfig: MiniAppSdkConfig) = ApiClient(
+        baseUrl = newConfig.baseUrl,
+        rasAppId = newConfig.rasAppId,
+        subscriptionKey = newConfig.subscriptionKey,
+        hostAppVersionId = newConfig.hostAppVersionId,
+        isTestMode = newConfig.isTestMode
+    )
 }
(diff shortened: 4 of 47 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -1,30 +1,60 @@
...
+    fun onRequestPermissionsResult(requestCode: Int, permissions: Array<String>, grantResults: IntArray) {
+        if (permissions.size == grantResults.size) {
+            for (i in permissions.indices)
+                webViewListener.onRequestPermissionsResult(requestCode, permissions[i], grantResults[i])
+        }
+    }
+
+    /** Post permission request from external. **/
+    internal fun requestPermission(permissions: Array<String>, requestCode: Int) =
+        ActivityCompat.requestPermissions(context as Activity, permissions, requestCode)
```
corycaywood, line 56: Sorry, I think the ticket for this was created before we had talked about how we want to handle permissions.

I think that we should let the Host App handle the request for device permissions. So we can add a method to `MiniAppMessageBridge` for something like `onRequestPermission` and then just pass the permission to the Host App as an enum or something. Does that sound okay @minh-rakuten @ts-nitish-a-jain?

Also, we need to be able to grant the permission to each Mini App independently of each other and cache which Mini Apps have been granted each permission. So the Host App would potentially show two permissions dialogs: the native permissions dialog and then also a custom dialog asking to grant permission to that particular Mini App. However, I think this task was only for getting the location permission working, so we can wait to implement the caching and independent permissions at a later time.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -21,8 +31,42 @@ internal class MiniAppWebChromeClient(context: Context) : WebChromeClient() {
         super.onReceivedTitle(webView, title)
     }
 
+    // region geolocation
+    private var geoLocationRequestOrigin: String? = null
+    @VisibleForTesting
+    internal var geoLocationCallback: GeolocationPermissions.Callback? = null
+
+    @Suppress("FunctionMaxLength")
+    override fun onGeolocationPermissionsShowPrompt(
```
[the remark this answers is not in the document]
minh-rakuten: There is use case that miniapps do not call `requestPermission`. They still can request a location access by `navigator.geolocation.getCurrentPosition` -> this will automatically trigger `onGeolocationPermissionsShowPrompt`.

The location access without using the js sdk to `requestPermission` is possible. Our js sample app & client app are working like that (using geolocation).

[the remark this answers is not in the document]
corycaywood: That's true, but I mean the same thing is possible on an Android App: an App could try to use the location without requesting the permission first, and then in that case the request for the location would throw an error. 

When a Mini App requests the location, they pass in an error listener `navigator.geolocation.getCurrentPosition(pos => console.log(pos), e => console.error(e))`. So in the case that they didn't called `requestPermission` first, then the browser will call the error listener.

Also, I think it's better if we can leave all of the device permission requesting out of the SDK and leave it up to the Host App to handle.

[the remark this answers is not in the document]
minh-rakuten: Okay, I see your point now.
I'll remove the permission handle in `onGeolocationPermissionsShowPrompt` so the MiniApps can only request a permission via js sdk.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -26,21 +26,6 @@ abstract class MiniApp internal constructor() {
...
-    /**
-     * Creates a mini app.
-     * @param info metadata of a mini app.
-     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful
-     * @param miniAppMessageBridge the interface for communicating between host app & mini app
-     * @throws MiniAppSdkException when there is some issue during fetching,
-     * downloading or creating the view.
-     */
-    @Throws(MiniAppSdkException::class)
-    @Deprecated(message = "Please replace with create(MiniAppId, MiniAppMessageBridge)")
```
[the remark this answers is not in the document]
ts-nitish-a-jain: +1

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -21,8 +31,41 @@ internal class MiniAppWebChromeClient(context: Context) : WebChromeClient() {
         super.onReceivedTitle(webView, title)
     }
 
+    // region geolocation
```
ts-nitish-a-jain, line 34: What is the intent/purpose of `//region geolocation`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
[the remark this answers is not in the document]
ts-nitish-a-jain: +1 -> It's better if we can leave all of the device permission requesting out of the SDK and leave it up to the Host App to handle.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -21,8 +31,41 @@ internal class MiniAppWebChromeClient(context: Context) : WebChromeClient() {
...
+    }
+    // end region
+
     @VisibleForTesting
     internal fun doInjection(webView: WebView) {
         webView.evaluateJavascript(bridgeJs) {}
     }
 }
+
+internal interface WebChromeListener {
```
ts-nitish-a-jain, line 69: Would it sound reasonable to update `WebChromeListener` -> `WebChromePermissionListener` ?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -37,8 +41,9 @@ internal class MiniAppWebView(
         settings.allowUniversalAccessFromFileURLs = true
         settings.domStorageEnabled = true
         settings.databaseEnabled = true
+        settings.setGeolocationEnabled(true)
```
ts-nitish-a-jain, line 44: If we want to mandate the permission access via the bridge, would it make sense to set the geolocation flag to `false` as it defaults to `true`, what do you think @corycaywood @minh-rakuten?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -1,30 +1,60 @@
...
+import android.app.Activity
+import android.content.Context
 import android.webkit.JavascriptInterface
+import androidx.annotation.VisibleForTesting
+import androidx.core.app.ActivityCompat
 import com.google.gson.Gson
 import com.rakuten.tech.mobile.miniapp.display.WebViewListener
 
+@VisibleForTesting
+internal const val GET_UNIQUE_ID = "getUniqueId"
```
ts-nitish-a-jain, line 12: These two can converge in a structure e.g. `ActionType` What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -1,30 +1,60 @@
...
-abstract class MiniAppMessageBridge {
+abstract class MiniAppMessageBridge(val context: Context) {
     private lateinit var webViewListener: WebViewListener
 
     /** Get provided id of mini app for any purpose. **/
     abstract fun getUniqueId(): String
 
     /** Handle the message from external. **/
-    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    @Suppress("TooGenericExceptionCaught", "SwallowedException", "LongMethod")
```
ts-nitish-a-jain, line 24: I think it shall be broken down. What do you think @minh-rakuten ?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
ts-nitish-a-jain: Device Permission access per mini-app is interesting. This would need confirmation on the product side if they wish to show the next popup or grant implicitly.
From a secure zone pov, it makes sense to have two dialogs. 
At the same time, caching can result in a bloated structure for maintaining the intelligence of permission access per mini-app.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
```diff
@@ -0,0 +1,24 @@
+package com.rakuten.tech.mobile.miniapp.js
+
+import android.Manifest
+
+internal object MiniAppPermission {
+
+    internal object Code {
```
ts-nitish-a-jain, line 7: `Code` -> `ReqCode` @minh-rakuten what do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
```diff
@@ -0,0 +1,24 @@
...
+        const val GEOLOCATION = 1001
+    }
+
+    internal object PermissionType {
+        const val GEOLOCATION = "miniapp.permission.geolocation"
+    }
+
+    internal fun getPermissionRequest(permissionType: String) = when (permissionType) {
+        PermissionType.GEOLOCATION -> arrayOf(Manifest.permission.ACCESS_FINE_LOCATION)
```
ts-nitish-a-jain, line 16: Should this have both `ACCESS_FINE_LOCATION` & `ACCESS_COARSE_LOCATION`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
minh-rakuten: it's just a coding style specifies the code block for this feature purpose. By specifying the code block, we can collapse the code visibility in AS.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
minh-rakuten: Ah, do you mean to set it to `false`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
minh-rakuten: When `ACCESS_FINE_LOCATION` is approved, the `ACCESS_COARSE_LOCATION` is implicity approved too.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
corycaywood: But if the host app is controlling the device permissions, we can't know when to change this value from `false` to `true`. Geolocation can't be accessed if this is set to `false` even if the device permission is granted, right?

minh-rakuten: This value must be `true` to send the position data back to miniapp in case that miniapp uses geolocation request.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -2,29 +2,62 @@ package com.rakuten.tech.mobile.miniapp.js
...
 abstract class MiniAppMessageBridge {
     private lateinit var webViewListener: WebViewListener
 
     /** Get provided id of mini app for any purpose. **/
     abstract fun getUniqueId(): String
 
+    /** Post permission request from external. **/
+    abstract fun requestPermission(
+        callbackId: String,
+        miniAppPermissionType: String,
```
corycaywood, line 19: I think we should pass a custom type rather than a string. Like maybe a `MiniAppPermissionType` enum?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -2,29 +2,62 @@ package com.rakuten.tech.mobile.miniapp.js
...
+        permissions: Array<String>
```
corycaywood, line 20: I don't think we need to pass an Array of permissions. It should be enough to pass a custom type such as `MiniAppPermissionType.GeoLocation` and then the Host App is responsible for interpreting which device permissions they need to request. That will let the Host App have more flexibility over the permissions and simplify things in the SDK.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -2,29 +2,62 @@ package com.rakuten.tech.mobile.miniapp.js
...
 /** Bridge interface for communicating with mini app. **/
 abstract class MiniAppMessageBridge {
     private lateinit var webViewListener: WebViewListener
 
     /** Get provided id of mini app for any purpose. **/
     abstract fun getUniqueId(): String
 
+    /** Post permission request from external. **/
+    abstract fun requestPermission(
+        callbackId: String,
```
corycaywood, line 18: It seems like passing the `callbackId` makes things complicated for the host app. Could we instead pass a function which the Host App needs to invoke? Something like `callback: (isGranted: Boolean) -> Unit`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
corycaywood: As I mentioned above, I think this should be moved to the Host App side so they have that control (`ACCESS_FINE_LOCATION` vs `ACCESS_COARSE_LOCATION`)

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
corycaywood: @ts-nitish-a-jain Device permission per Mini App was already discussed by the product side - I think this is what's expected. We had discussed using similar mechanisms as the custom permissions for caching which permissions have been granted.

In the end, it's really up to the Host App though. If they don't care about granting it on a per Mini App basis, then they don't have to show a Mini App specific permissions dialog and can just implicitly grant the permission to all Mini Apps.

miniapp/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,8 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
```
minh-rakuten, line 8: @corycaywood @ts-nitish-a-jain if we let the hostapp in full control of permission, should we define those permissions on hostapp `AndroidManifest`?

ts-nitish-a-jain: For sure.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -1,30 +1,67 @@
...
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
 /** Bridge interface for communicating with mini app. **/
 abstract class MiniAppMessageBridge {
     private lateinit var webViewListener: WebViewListener
 
     /** Get provided id of mini app for any purpose. **/
     abstract fun getUniqueId(): String
 
+    /** Post permission request from external. **/
+    abstract fun requestPermission(
```
corycaywood, line 18: Oh, sorry, I just noticed that the iOS side is using `requestPermission(permissionType: MiniAppPermissionType, completionHandler: @escaping (Result<String, Error>) -> Void)`. 

So maybe it makes more sense to move the `miniAppPermissionType` as the first parameter and `callback` as the second. What do you think?

miniapp/src/main/assets/bridge.js
```diff
@@ -15,22 +16,24 @@ var isPlatform = {
  * Method to call the native interface methods for respective platforms
  * such as iOS & Android
  * @param  {[String]} action Action command/interface name that native side need to execute
+ * @param  {Object} param Object that contains request parameter values like permissions.
+ * For eg., {permission: 'location'}
  * @param  {[Function]} onSuccess Success callback function
  * @param  {[Function]} onError Error callback function
  */
-MiniAppBridge.exec = function(action, onSuccess, onError) {
+MiniAppBridge.exec = function(action, param, onSuccess, onError) {
```
ts-nitish-a-jain, line 24: Can we follow the convention lines of JSON RPC instead (`method`, `params`)?
@ts-leojoseph-r @corycaywood @minh-rakuten

miniapp/src/main/assets/bridge.js
```diff
@@ -91,14 +94,30 @@ MiniAppBridge.getUniqueId = function() {
   return new Promise(function(resolve, reject) {
     return MiniAppBridge.exec(
       "getUniqueId",
+      null,
```
ts-nitish-a-jain, line 97: What do you think of passing an empty payload instead of null?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -23,7 +24,13 @@ internal class RealMiniApp(
     @Suppress("TooGenericExceptionThrown")
     override suspend fun create(info: MiniAppInfo): MiniAppDisplay =
         executingCreate(info.id, object : MiniAppMessageBridge() {
-            override fun getUniqueId(): String = throw Exception("MiniAppMessageBridge has not been implemented")
+            override fun getUniqueId(): String =
+                throw Exception("MiniAppMessageBridge has not been implemented")
```
ts-nitish-a-jain, line 28: Can the exception object be extracted into a single place?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
ts-nitish-a-jain: Just that all of the SDK has `suspend` and this would be the first place to have a callback.

miniapp/src/main/assets/bridge.js
minh-rakuten: `null` represents the non-usage so I think null is better.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
```diff
@@ -0,0 +1,22 @@
...
+
+    internal companion object {
+
+        internal fun getValue(type: String) = values().find { it.type == type } ?: UNKNOWN
+    }
+}
+
+internal enum class MiniAppPermissionResult(val type: String) {
+    ALLOWED("Allowed"),
+    DENIED("Denied");
```
ts-nitish-a-jain, line 16: This should have a third type as well, depicting the state of `DO_NOT_ASK_AGAIN`. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
ts-nitish-a-jain: Okay. 
I could also suggest using the logical purpose-based grouping as a sequence for methods in the class as recommended as well in the kotlin docs.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
minh-rakuten: There is a state of `DO_NOT_ASK_AGAIN` and from android 10 there is also a new state. However, the result from `PackageManager` only has two values: PERMISSION_GRANTED & PERMISSION_DENIED.

The result from both android & ios should only return two basic results.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
minh-rakuten: Oh, can you refer me to that kotlin doc?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
corycaywood: @ts-nitish-a-jain But the host app has to rely on `onRequestPermissionsResult` in the Activity, so it seems like using a `suspend` function would make things more complicated for them. I can't really think of a good way to make it work with `suspend`. How about you @minh-rakuten ?

miniapp/src/main/assets/bridge.js
corycaywood: That's basically what we're doing, so it's really just a matter of naming, right?

ts-leojoseph-r: Yeah, that's right. We send the request like a JSON RPC already

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
ts-nitish-a-jain: Thanks for sharing @minh-rakuten. I will check through the updates. 
It should have three results, accordingly, the Mini App could propose going to the Settings screen if required.

Platform differences are acceptable.
@ts-leojoseph-r @Climbatize is it handled the same way in iOS?

miniapp/src/main/assets/bridge.js
ts-nitish-a-jain: Yes, primarily the naming in this turn. Structure uniformity we could probably target in the next ones. What do you guys think?

Thanks @ts-leojoseph-r for attempting based on the historical discussion.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
ts-nitish-a-jain: https://developer.android.com/kotlin/style-guide#top-level_declarations

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
ts-leojoseph-r: @ts-nitish-a-jain  In iOS we are returning **Allowed**, **Denied**, **Not Determined** (When user taps on Not now)

ts-nitish-a-jain: @minh-rakuten @ts-leojoseph-r @corycaywood 
I think the Result should model the superset of all the result cases. What do you guys think?
Android 11 has one-time permission and so does iOS.

minh-rakuten: What android & ios are doing is:
- post `success` callback if permission is granted.
- post `error` callback if permission is denied. (whatever the reason is).

Android & iOS have different text results for different cases. I don't think the miniapp wants to handle all the result text cases. Miniapp just cares about whether the permission is granted or not.

ts-nitish-a-jain: But mini-app would need to know if they need to ask for permission again or not for subsequent use.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
ts-nitish-a-jain: Okay. Makes sense. Thanks. I don't have any solution either for this at this moment.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppPermission.kt
minh-rakuten: I believe miniapp should always ask for permission because user anytime can disable / enable it on phone Setting.

corycaywood: Yeah, I think @minh-rakuten is right. We can require that the Mini App always request the permission before use of features. We discussed this during the architecture discussions and product side seemed to agree.

ts-nitish-a-jain: @minh-rakuten you are right about that. 
From the pov of the code, it should speak the same language as it will be recommended in the docs; to be indicative that they must ask for permission before using the feature. Need to update the script code too for that.

This tells me that we should also expose a hasPermissionAccess method?

minh-rakuten: @ts-nitish-a-jain Yeah, we need to update the document to mention that miniapp should always request permissions before getting data.

What does `hasPermissionAccess` do on mobile sdk side?

corycaywood: I think that `hasPermissionAccess` will add a lot of complexity. We won't know what device permissions the Host App is using and the user could revoke the permission at any time, so we wouldn't be able to know anything about this on the SDK side. It would have to be another method that the Host App needs to integrate. Maybe we can leave it for a possible future enhancement if we can find a good way of doing it?

ts-nitish-a-jain: Ok. Future is fine too.

PR #110, opened 2020-07-14
feature: support js module type

minh-rakuten:
# Description

- Rakuten Casa does not work on android due to module MIME type. i.e `<script src="main-es2015.js" type="module">`.
This PR defines the MIME type `application/javascript` for js resource.

- Hide the visibility of Testbox toggle because the api is not ready. Temporarily move this toggle to the bottom so it does not affect the layout structure.

## Links
[MINI-1071](https://jira.rakuten-it.com/jira/browse/MINI-1071)

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -14,10 +14,24 @@ internal class MiniAppWebViewClient(
...
+        return interceptedWebRequest
+    }
+
+    @VisibleForTesting
+    internal fun interceptMimeType(
+        interceptedWebRequest: WebResourceResponse?,
         request: WebResourceRequest
-    ): WebResourceResponse? = loader.shouldInterceptRequest(request.url)
+    ) {
+        interceptedWebRequest?.let {
```
ts-nitish-a-jain, line 30: I'm not sure if it is a good idea as it may slow down. Isn't it solvable in the mini-app build setting itself?
@minh-rakuten @corycaywood @Rajesh-Sundaram

minh-rakuten: @Rajesh-Sundaram on Android 9, RakutenCasa has this error
`Failed to load module script: The server responded with a non-JavaScript MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.`

Is there anything miniapp can solve by itself?

corycaywood: I don't think there's any way for the mini app to solve this other than not using the `module` type. So they would have to always bundle their JS into a single file instead.

Mini Apps usually shouldn't have too many files, so would this have that much affect on the loading time?

Also, on iOS side we added the MIME types for everything supported by the backend: [see here](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/Display/MimeTypes.swift). Would it make sense to do the same for Android?

minh-rakuten: @corycaywood For now, Android only has a problem with `module` type so I think only MIME type for javascript is enough.
OS 10 and later, android already solves this issue.

testapp/src/main/res/layout/settings_menu_activity.xml
```diff
@@ -72,6 +65,14 @@
         android:inputType="text" />
     </com.google.android.material.textfield.TextInputLayout>
 
+    <com.google.android.material.switchmaterial.SwitchMaterial
+        android:id="@+id/switchTestMode"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="@dimen/small_12"
+        android:text="@string/lb_test_mode"
+        android:visibility="invisible"/>
```
khairul-alam-licon, line 74: Does it need to be invisible? I don't know the spec though.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -14,10 +14,24 @@ internal class MiniAppWebViewClient(
     private val customScheme: String
 ) : WebViewClient() {
 
-    override fun shouldInterceptRequest(
-        view: WebView,
+    override fun shouldInterceptRequest(view: WebView, request: WebResourceRequest): WebResourceResponse? {
+        val interceptedWebRequest = loader.shouldInterceptRequest(request.url)
+
+        interceptMimeType(interceptedWebRequest, request)
+
```
khairul-alam-licon, line 21: (optional) maybe we can remove the empty line L19, L21. But, it's upto you.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -14,10 +14,24 @@ internal class MiniAppWebViewClient(
...
+        return interceptedWebRequest
+    }
+
+    @VisibleForTesting
+    internal fun interceptMimeType(
+        interceptedWebRequest: WebResourceResponse?,
```
khairul-alam-licon, line 27: writing `response: WebResourceResponse?` would be nicer.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -14,10 +14,24 @@ internal class MiniAppWebViewClient(
...
         request: WebResourceRequest
-    ): WebResourceResponse? = loader.shouldInterceptRequest(request.url)
+    ) {
+        interceptedWebRequest?.let {
+            if (request.url.toString().endsWith("js", true))
```
khairul-alam-licon, line 31: Do we need to check the `null` as well? e.g. `if (request.url != null && request.url.toString()..)`

khairul-alam-licon: I would put a dot `(.)` before the `"js"` for safe call.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -14,10 +14,24 @@ internal class MiniAppWebViewClient(
...
+                it.mimeType = "application/javascript"
```
khairul-alam-licon, line 32: Do we need to use any `const val` to put this `"application/javascript"`?

e.g.

```
private companion object {
    const val MIME_TYPE_JS = "application/javascript"
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -14,10 +14,24 @@ internal class MiniAppWebViewClient(
     private val customScheme: String
 ) : WebViewClient() {
 
-    override fun shouldInterceptRequest(
-        view: WebView,
+    override fun shouldInterceptRequest(view: WebView, request: WebResourceRequest): WebResourceResponse? {
+        val interceptedWebRequest = loader.shouldInterceptRequest(request.url)
+
+        interceptMimeType(interceptedWebRequest, request)
```
ts-nitish-a-jain, line 20: From SRP point of view, it makes sense to separate the functionality into a new method. 

Though, because I think this callback is quite related to the rendering cycle of the mini-app in WebView, I am not sure if applying code within the same method has faster response in millis. @minh-rakuten do you think we can check which is faster without a jump to new method in call stack or without jumping?

Also, because, this is invoked for all the resources, the `if` would always be executing. I was looking at issue [issue](https://bugs.chromium.org/p/chromium/issues/detail?id=824651). Should we really support it or provide documentation on an alternate solution? What do you think @corycaywood @minh-rakuten?

testapp/src/main/res/layout/settings_menu_activity.xml
ts-nitish-a-jain: I think this is temporary because Testing API endpoints are not ready yet. We can probably put a TODO here to change it back to visible once they are available.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
ts-nitish-a-jain: @corycaywood Would bundling be more natural for a prod-type of mini-app, especially when a javascript is written in a `type=module` format by a developer?
@minh-rakuten this issue happens only for Android 9?

minh-rakuten: This issue happens on Android 9 and below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
minh-rakuten: For now, I keep it as in function scope because there is only javascript support. If we have more than one, I will implement it similar to [ios](https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/Display/MimeTypes.swift).

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
corycaywood: I think we need to support `type=module`: it's a new feature from ES6 so it provides some benefits from bundling the JS in a single file. One important benefit is that they have deferred loading by default, which means they don't block rendering of the page while the browsing loading/parsing the JS.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
minh-rakuten: when the method `shouldInterceptRequest` was triggered automatically by webview, `request.url` is never null, that's what I know 😅 but not so sure.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
minh-rakuten: I will put it in the same block scope then.

I don't have an idea of a decision. All I can tell is This is a bug of android webview itself and had been fixed since Android 10.

corycaywood: As I mentioned above, I think we do need to support this because the `module` type provides some benefits to mini app developers. It doesn't seem like this should have a significant performance impact as long as the Mini App has a reasonable number of files.

By the way @ts-leojoseph-r @Climbatize Are there any issues with using `type=module` on iOS side?

minh-rakuten: iOS had the problem with Rakuten Casa `type=module` too.
But iOS had fixed it in this [PR](https://github.com/rakutentech/ios-miniapp/pull/96).

Solution for both android & ios is defining MIME type.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
khairul-alam-licon: Yeah, I was concerned if any client send a non-null file e.g. "somethingjs", which is an invalid JS file. But, "something.js" is a valid JS file.

khairul-alam-licon: Yeah, I was concerned if any client send a file e.g. "somethingjs", which is an invalid JS file. But, "something.js" is a valid JS file.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
Climbatize: Yeah we had to artificially provide mime types in order to make Case work on iOS

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
minh-rakuten: absolutely true.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
ts-nitish-a-jain: From the look of it, I understand that we might have to support it as a mimetype even though it seems to be a limitation on the web tech of not being able to provide the correct mimetype when this type - module is selected.

At the same time, number of files inside a mini-app would be out of control area.

@minh-rakuten, could you share the time diff from the extra hop? 
UT can take a back seat on this.

minh-rakuten: @ts-nitish-a-jain 
Loading Rakuten Casa

Separate method: 43 ms
Same scope: 35 ms

corycaywood: Couldn't you make it an `inline` function?

minh-rakuten: I doubt the efficient of using `inline` here 😂.

ts-nitish-a-jain: To clear the doubt, once you profile that section and scale it up with an assumption of 1000/10000 files, you will probably find it out.

minh-rakuten:
> LGTM 👍
> But needs change for inline keyword.

I put `inline` in experiment and there is no change improvement. Looks like `inline` is only efficient for higher-order function.

ts-nitish-a-jain:
> > LGTM 👍
> > But needs change for inline keyword.
> 
> I put `inline` in experiment and there is no change improvement. Looks like `inline` is only efficient for higher-order function.

@minh-rakuten I think what you mentioned is correct about inline.
The aim is to prevent a new call stack from this override. So could you please account for those changes in another PR. The reason we can further analyze only by comparing the values when there is no jump and when there is a jump to a new method.

minh-rakuten:
> > > LGTM 👍
> > > But needs change for inline keyword.
> > 
> > 
> > I put `inline` in experiment and there is no change improvement. Looks like `inline` is only efficient for higher-order function.
> 
> @minh-rakuten I think what you mentioned is correct about inline.
> The aim is to prevent a new call stack from this override. So could you please account for those changes in another PR. The reason we can further analyze only by comparing the values when there is no jump and when there is a jump to a new method.

@ts-nitish-a-jain 
I got this warning from android for our implementation case here:
`Expected performance impact of inlining 'function definition' is insignificant. Inlining works best for functions with parameters of functional types`
Using inline would create a larger byte-code file, it's not the win-win.

ts-nitish-a-jain:
> > > > LGTM 👍
> > > > But needs change for inline keyword.
> > > 
> > > 
> > > I put `inline` in experiment and there is no change improvement. Looks like `inline` is only efficient for higher-order function.
> > 
> > 
> > @minh-rakuten I think what you mentioned is correct about inline.
> > The aim is to prevent a new call stack from this override. So could you please account for those changes in another PR. The reason we can further analyze only by comparing the values when there is no jump and when there is a jump to a new method.
> 
> @ts-nitish-a-jain
> I got this warning from android for our implementation case here:
> `Expected performance impact of inlining 'function definition' is insignificant. Inlining works best for functions with parameters of functional types`
> Using inline would create a larger byte-code file, it's not the win-win.

@minh-rakuten Yes, the same thing, I tried to acknowledge in my last comment. I think what you mentioned is correct about `inline`. It looks to be working for higher-order function.

So that's why: The aim is to prevent a new call stack from this override. 
`func1() {
  callToFunc2()
}
vs
func1() {
   codeOfFunc2
}`

We can further analyze only by comparing the values when there is no jump and when there is a jump to a new method. Most likely, the second one will be faster as the call stack operation might get skipped. Because this scenario is related to rendering, it is important to profile the changes. This is also referred to as Shikumika in Rakuten I think. 😄

minh-rakuten:
> We can further analyze only by comparing the values when there is no jump and when there is a jump to a new method. Most likely, the second one will be faster as the call stack operation might get skipped. Because this scenario is related to rendering, it is important to profile the changes. This is also referred to as Shikumika in Rakuten I think. 😄
@ts-nitish-a-jain 
How much faster for the second one to be faster? Tell me why you skip the warning from the android kotlin?
Your suggestion is not persuasive for me here but you can try to convince other team members and have a PR for it.

ts-nitish-a-jain:
> > We can further analyze only by comparing the values when there is no jump and when there is a jump to a new method. Most likely, the second one will be faster as the call stack operation might get skipped. Because this scenario is related to rendering, it is important to profile the changes. This is also referred to as Shikumika in Rakuten I think. 😄
> 
> @ts-nitish-a-jain
> How much faster for the second one to be faster? Tell me why you skip the warning from the android Kotlin?
> Your suggestion is not persuasive for me here but you can try to convince other team members and have a PR for it.

If there is a difference wouldn't that be technically, mathematically faster (along with a focus on immediate rendering to effect), would that make sense?
In the PR I couldn't see the warning 🤣 Sorry about that.

minh-rakuten:
@ts-nitish-a-jain
I put a warning message in previous comment already :)
The mathematically faster here is the thing that you need to prove, it's not about what you & me think.

PR #115, opened 2020-07-20
Update docs for 1.2.0 release

ts-nitish-a-jain:
# Description
Update docs for 1.2.0 release

## Links
[MINI-1252](https://jira.rakuten-it.com/jira/browse/MINI-1252?src=confmacro)

# Checklist
- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md).
- [ ] I removed all sensitive data before every commit, including API endpoints and keys
- [ ] I ran `./gradlew check` without errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 ### 1.2.0 (2020-07-21)
 **SDK**
 - Upgraded build setup and some internal dependencies
-- Added support for customization of user agent information
+- Added support for customization of user agent information. See [this](https://github.com/rakutentech/android-miniapp/blob/master/miniapp/USERGUIDE.md#2-configure-sdk-settings-in-androidmanifestxml)
 - Added feature to support backward & forward navigation support in a mini app
-- Added support for javascript of type module
+- Added support for javascript of type module e.g. `<script src="some_module.js" type="module"></script>` would render just fine now even though the mimetype is missed out by the web technologies for this scenario.
 - Added feature to obtain geolocation data in a mini app
 - Hotfix for redirection over custom scheme and http
 
@@ -29,7 +29,7 @@
 - Added JavaScript bridge for passing data between Mini App and Host App. Your App now must implement `MiniAppMessageBridge` and provide the implementation when calling `MiniApp#create`.
 - Deprecated `MiniApp#create(info: MiniAppInfo)`. Your App should instead use `MiniApp#create(info: MiniAppInfo, miniAppMessageBridge: MiniAppMessageBridge)`.
 - Added `getUniqueId` function to `MiniAppMessageBridge`. This function should provide a unique identifier (unique to the user and device) to Mini Apps.
-- Added support for custom scheme URL redirect. The URL `mscheme.MINI_APP_ID://miniapp/index.html` can be used from within the Mini App view to redirect to the Mini App. This matches the URL used in the iOS Mini App SDK.
+- Added support for custom scheme URL redirect. The URL `mscheme.MINI_APP_ID://miniapp/index.html` can be used by a Mini App for redirection. This matches the URL used in the iOS Mini App SDK.
 
 ### 1.0.0 (2020-04-21)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,34 @@
+### 1.2.0 (2020-07-21)
+**SDK**
+- Upgraded build setup and some internal dependencies
+- Added support for customization of user agent information
+- Added feature to support backward & forward navigation support in a mini app
+- Added support for javascript of type module
+- Added feature to obtain geolocation data in a mini app
+- Hotfix for redirection over custom scheme and http
+
+**Sample App**
+- Updated styling of the mini app list
+- Updated App's setting screen with build information
+- Demo usage of customized user agent information
+- Add navigation into mini app
+
+### 1.1.1 (2020-06-11)
+
+**SDK**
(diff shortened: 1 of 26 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
+## CHANGELOG
+
+### 1.2.0 (2020-07-21)
+**SDK**
+- Upgraded build setup and some internal dependencies
+- Added support for customization of user agent information
+- Added feature to support backward & forward navigation support in a mini app
```
khairul-alam-licon, line 7: I think, we could write: `Added backward & forward navigation support in a mini app`

CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
...
+
+**Sample App**
+- Display first time setup instructions on first launch of App.
+
+### 1.1.0 (2020-06-02)
+
+- Added JavaScript bridge for passing data between Mini App and Host App. Your App now must implement `MiniAppMessageBridge` and provide the implementation when calling `MiniApp#create`.
+- Deprecated `MiniApp#create(info: MiniAppInfo)`. Your App should instead use `MiniApp#create(info: MiniAppInfo, miniAppMessageBridge: MiniAppMessageBridge)`.
+- Added `getUniqueId` function to `MiniAppMessageBridge`. This function should provide a unique identifier (unique to the user and device) to Mini Apps.
+- Added support for custom scheme URL redirect. The URL `mscheme.MINI_APP_ID://miniapp/index.html` can be used from within the Mini App view to redirect to the Mini App. This matches the URL used in the iOS Mini App SDK.
```
khairul-alam-licon, line 32: Does it mean: `can be used to redirect from Mini App view to the Mini App` ?

CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
+## CHANGELOG
+
+### 1.2.0 (2020-07-21)
+**SDK**
+- Upgraded build setup and some internal dependencies
+- Added support for customization of user agent information
```
corycaywood, line 6: This should offer more explanation on how they can do this. i.e. `Added support for customization of user agent information. This can be done by setting a value for `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` in your manifest.` Or maybe just add a link to `Configure SDK settings in AndroidManifest.xml` section?

CHANGELOG.md
corycaywood, line 7: This also needs more explanation. Maybe just a link to the correct section in the user guide? Actually, I don't see this in the user guide, so maybe that needs to be added?

CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
...
+- Added feature to support backward & forward navigation support in a mini app
+- Added support for javascript of type module
```
corycaywood, line 8: This isn't very clear what was fixed. Maybe add an example: `<script src="some_module.js" type="module"></script>`.

CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
...
+- Added feature to obtain geolocation data in a mini app
```
corycaywood, line 9: Need to also mention that `requestPermission` was added and should be implemented by host app.

CHANGELOG.md
ts-nitish-a-jain: I think that is out of scope of the `changelog`.
`how-to` should be a part of the UserGuide instead. What do you think?

CHANGELOG.md
ts-nitish-a-jain: I think it is fine either way.

ts-nitish-a-jain: No, there is a difference in both statements.

CHANGELOG.md
ts-nitish-a-jain: Yes, it seems that this can be updated slightly. What it is trying to convey is that a custom scheme of the type declared can be used for redirection by the mini-app.

CHANGELOG.md
ts-nitish-a-jain: Okay, we can give an example too.

CHANGELOG.md
ts-nitish-a-jain: The user guide is updated now.

CHANGELOG.md
ts-nitish-a-jain: This has a mention in the user guide at this moment. I think this will get better with the linking afterwards.

CHANGELOG.md
ts-nitish-a-jain: This also got addressed in the userguide. This will get better with the linking afterwards.

miniapp/USERGUIDE.md
```diff
@@ -153,6 +161,8 @@ class MiniAppActivity : Activity(), CoroutineScope {
 }
 ```
 
+`MiniAppDisplay.navigateBackward` and `MiniAppDisplay.navigateForward` facilitates the navigation inside a mini app if the history stack is available in it. A common usage pattern could be to link it up to the Android Back Key navigation.
```
corycaywood, line 164: I think we should have a code example of this. Maybe this can be moved to the `Advanced` section and then add a code sample similar to what's in the Sample App:

```
override fun onBackPressed() {
    if(!miniAppDisplay.navigateBackward()) {
        super.onBackPressed()
    }
}
```

ts-nitish-a-jain: Okay. This could be added too.

PR #117, opened 2020-07-21
feature: webview custom dialog

minh-rakuten:
# Description
- Override js dialog event to not show URL.
- The action texts of the native dialog are shown as localized strings.
- Show miniapp display name in dialog title. This made the changes in source code:
Pass `MiniAppInfo` to webview. Remove `miniapp id`.
Save `MiniAppInfo` as json string in local.  Remove `version id` in storage.

## Links
[MINI-1228](https://jira.rakuten-it.com/jira/browse/MINI-1228)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -26,6 +26,7 @@ internal class MiniAppDownloader(
             var versionPath = storage.getMiniAppVersionPath(miniAppInfo.id, miniAppInfo.version.versionId)
             if (!miniAppStatus.isVersionDownloaded(miniAppInfo.id, miniAppInfo.version.versionId, versionPath))
                 versionPath = startDownload(miniAppInfo)
+
             storeDownloadedMiniApp(miniAppInfo)
 
             return Pair(versionPath, miniAppInfo)
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -23,14 +23,8 @@ internal class MiniAppDownloader(
     suspend fun getMiniApp(appId: String): Pair<String, MiniAppInfo> {
         try {
             val miniAppInfo = apiClient.fetchInfo(appId)
-            var versionPath =
-                storage.getMiniAppVersionPath(miniAppInfo.id, miniAppInfo.version.versionId)
-            if (!miniAppStatus.isVersionDownloaded(
-                    miniAppInfo.id,
-                    miniAppInfo.version.versionId,
-                    versionPath
-                )
-            )
+            var versionPath = storage.getMiniAppVersionPath(miniAppInfo.id, miniAppInfo.version.versionId)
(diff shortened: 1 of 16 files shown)


corycaywood:
It's looking good, but I think we should be able to inject the Mini App title into the webview.

I'm thinking that instead of fetching the latest `MiniAppInfo` from within `MiniAppDownloader`, you could instead move this to `RealMiniApp#executingCreate` (using `miniAppInfoFetcher.getInfo(miniAppId)`). And then we can inject the `MiniAppInfo` into the the call to `Displayer#createMiniAppDisplay`. What do you think?

minh-rakuten:
> It's looking good, but I think we should be able to inject the Mini App title into the webview.
> 
> I'm thinking that instead of fetching the latest `MiniAppInfo` from within `MiniAppDownloader`, you could instead move this to `RealMiniApp#executingCreate` (using `miniAppInfoFetcher.getInfo(miniAppId)`). And then we can inject the `MiniAppInfo` into the the call to `Displayer#createMiniAppDisplay`. What do you think?

@corycaywood We can let the `MiniAppDownloader#getMiniApp` return pair of values `Pair<MiniAppPath: String, miniappInfo: MiniAppInfo>` and we can pass that `MiniAppInfo` to `Displayer#createMiniAppDisplay`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/WebViewDialog.kt
```diff
@@ -0,0 +1,93 @@
...
+internal fun onShowDialog(
+    context: Context,
+    message: String?,
+    defaultValue: String? = "",
+    result: JsResult?,
+    dialogType: DialogType,
+    miniAppInfo: MiniAppInfo
+): Boolean {
+    val dialogBuilder = AlertDialog.Builder(context)
+    dialogBuilder.setTitle(miniAppInfo.displayName).setMessage(message)
```
corycaywood, line 23:
```suggestion
    val dialogBuilder = AlertDialog.Builder(context)
        .setTitle(miniAppInfo.displayName)
        .setMessage(message)
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -55,20 +60,16 @@ internal class MiniAppDownloader(
...
+        val versionId = miniAppInfo.version.versionId
         val baseSavePath = storage.getMiniAppVersionPath(appId, versionId)
         when {
             isManifestValid(manifest) -> {
                 for (file in manifest.files) {
                     val response = apiClient.downloadFile(file)
                     storage.saveFile(file, baseSavePath, response.byteStream())
                 }
-                miniAppStatus.setVersionDownloaded(appId, versionId, true)
-                miniAppStatus.saveDownloadedVersion(appId, versionId)
```
khairul-alam-licon, line 71: Just curious to know, could we call `storeDownloadedMiniApp(miniAppInfo)` here?

minh-rakuten: There is the reason why I changed the call of this method in `getMiniApp` function.

We cached only `versionId` before and now replace it with `miniAppInfo`. Our SDK checks that there is cache but no `miniAppInfo` saved so we need to update the status of downloaded miniapp even it's already downloaded.

PR #119, opened 2020-07-22
fix: no invalid field warning in settings screen

khairul-alam-licon:
# Description
 A warning should appear while there are invalid inputs like the following cases:
- App host id is not UUID
- empty / blank characters 

For the UI, this PR aims to use `setError(CharSequence error)` from [TextView](https://developer.android.com/reference/android/widget/TextView).

## Links
- [MINI-1271](https://jira.rakuten-it.com/jira/browse/MINI-1271)

## Screenshots
Screen Shot 2020-07-27 at 12 36 45

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
@@ -9,6 +9,3 @@ fun String.isInvalidUuid(): Boolean = try {
 } catch (e: IllegalArgumentException) {
     true
 }
-
-fun String.includeHyphen(): String =
-    replace("(.{8})(.{4})(.{4})(.{4})(.+)".toRegex(), "$1-$2-$3-$4-$5")
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
@@ -1,10 +1,14 @@
 package com.rakuten.tech.mobile.testapp.helper
 
+import java.lang.IllegalArgumentException
 import java.util.UUID
 
 fun String.isInvalidUuid(): Boolean = try {
     UUID.fromString(this)
     false
-} catch (e: Exception) {
+} catch (e: IllegalArgumentException) {
     true
 }
+
+fun String.includeHyphen(): String =
+    replace("(.{8})(.{4})(.{4})(.{4})(.+)".toRegex(), "$1-$2-$3-$4-$5")
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
@@ -1,7 +1,6 @@
 package com.rakuten.tech.mobile.testapp.helper
 
-import java.lang.Exception
-import java.util.*
+import java.util.UUID
 
 fun String.isInvalidUuid(): Boolean = try {
     UUID.fromString(this)
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
@@ -40,7 +40,7 @@ class SettingsMenuActivity : BaseActivity() {
         override fun beforeTextChanged(s: CharSequence?, start: Int, count: Int, after: Int) {}
 
         override fun onTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {
-            enableSaveView()
+            validateInputView()
         }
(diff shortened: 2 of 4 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
```diff
@@ -99,17 +101,34 @@ class SettingsMenuActivity : BaseActivity() {
...
         return "Build " + getString(R.string.miniapp_sdk_version) + " - " +
                 getString(R.string.build_version)
     }
 
-    private fun enableSaveView() {
-        saveViewEnabled =
-            !(editAppId.text.toString().isInvalidUuid() || editSubscriptionKey.text.isNullOrEmpty())
+    internal fun validateInputIDs() {
+        val isAppIdInvalid = editAppId.text.toString().isInvalidUuid()
+        val isSubsKeyInvalid = editSubscriptionKey.text.toString().includeHyphen().isInvalidUuid()
```
corycaywood, line 114: I'm not sure we need to validate the subscription key as a UUID. The spec doesn't mention anything about it being a UUID, so I don't think we can rely on it always being that format.

The iOS SDK just validates that it's not empty ([see here](https://github.com/rakutentech/ios-miniapp/blob/master/Example/SettingsTableViewController.swift#L192)), so maybe that's enough?

khairul-alam-licon:
Thanks for the reviews. I am checking with QA related to the invalid inputs spec of subscription key.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
ts-leojoseph-r: @corycaywood is right. The subscription key doesn't need to a UUID. It could be any string.

khairul-alam-licon:
> Delete the screenshots from PR. We shouldn't be exposing Subscription and Host app ID

I have updated the screenshot. Thank you for pointing it.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
```diff
@@ -1,11 +1,11 @@
...
-import java.lang.Exception
-import java.util.*
+import java.lang.IllegalArgumentException
+import java.util.UUID
 
 fun String.isInvalidUuid(): Boolean = try {
     UUID.fromString(this)
     false
-} catch (e: Exception) {
+} catch (e: IllegalArgumentException) {
```
ts-nitish-a-jain, line 9: What will be the other case when this fails?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
```diff
@@ -99,17 +100,32 @@ class SettingsMenuActivity : BaseActivity() {
...
     private fun createSettingsInfo(): String {
         return "Build " + getString(R.string.miniapp_sdk_version) + " - " +
                 getString(R.string.build_version)
     }
 
-    private fun enableSaveView() {
-        saveViewEnabled =
-            !(editAppId.text.toString().isInvalidUuid() || editSubscriptionKey.text.isNullOrEmpty())
+    internal fun validateInputIDs() {
+        val isAppIdInvalid = editAppId.text.toString().isInvalidUuid()
```
ts-nitish-a-jain, line 112: The validation logic can be separated from fetching the values and setting error values.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
khairul-alam-licon: Currently, `UUID.fromString(this)` is designed to throw only `IllegalArgumentException`. So, it's enough to use this specific exception. But, we should add the related exception in future when the implementation will be changed inside `String.isInvalidUuid()` function.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
khairul-alam-licon: Yes, I agree to have SRP principle per operation e.g. fetching/setting values. In this PR, I focused to write the validation operation in a single function. We can separate the operations precisely while adding more spec in future.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
khairul-alam-licon: Currently, `UUID.fromString(this)` is designed to throw `IllegalArgumentException`. So, it's enough to use this specific exception now. But, we should add the related exception in future when the implementation will be changed inside `String.isInvalidUuid()` function.

khairul-alam-licon:
Thank you for your reviews.

PR #122, opened 2020-07-29
Add format for changelogs

khairul-alam-licon:
# Description
Changelogs could be formatted by a convention. This PR aims to contribute a format.

# Checklist
- [X] I have read the [contributing guidelines](../CONTRIBUTING.md).
- [X] I removed all sensitive data before every commit, including API endpoints and keys
- [X] I ran `./gradlew check` without errors

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -82,7 +82,7 @@ Sample App
 ```
 
 ### Description
-A sentence which will describe the type of change with an information.
+A sentence describes the type of change with an information.
 
 ```
 <bold><change type></bold>: <subject>
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -68,3 +68,24 @@ The footer should contain any information about **Breaking Changes** and is also
 reference GitHub issues that this commit **Closes**.
 
 **Breaking Changes** should start with the word `BREAKING CHANGE:` with a space or two newlines. The rest of the commit message is then used for this.
+
+## Changelog Guidelines
+All changes will be documented [in this file](https://github.com/rakutentech/android-miniapp/blob/master/CHANGELOG.md) by ordering chronologically for each version with the following format.
+
+```
+Release Version (YYYY-MM-DD)
+SDK
+- <description>
+<BLANK LINE>
+Sample App
+- <description>
+```
+
+### Description
+A sentence which will describe the type of change with some information.
+
+```
+<change type:><SPACE><information>
+```
+
+#### Change type
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -89,9 +89,9 @@ A sentence describes the type of change with an information.
 ```
 
 #### Change type
-* **Feature**: Added a new feature
-* **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too
-* **Deprecated**: Still available something for using but will no longer be developed or supported. This is an indication for not using it
-* **Obsoleted**: No longer works as expected since a better alternative is available and will maybe discontinued from the next releases
-* **Removed**: When a feature has been expired by the sunsetting deadline
-* Write "no changes" when there is no difference with previous release version
+* **Feature**: Added a new feature.
+* **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too.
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -67,4 +67,31 @@ The body should include the motivation for the change and contrast this with pre
...
+A sentence describes the type of change with an information.
+
+```
+<bold><change type></bold>: <subject>
+```
+
+#### Change type
+* **Feature**: Added a new feature
+* **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too
+* **Deprecated**: Still available something for using but will no longer be developed or supported. This is an indication for not using it
```
corycaywood, line 94:
```suggestion
* **Deprecated**: Still available for use but will no longer be developed or supported. This is an indication it should not be used and will be removed in the next major version release.
```

khairul-alam-licon:
@ts-nitish-a-jain @ts-leojoseph-r @minh-rakuten 
What do you think guys? 🤔

Climbatize:
> @ts-nitish-a-jain @ts-leojoseph-r @minh-rakuten
> What do you think guys? 🤔

right now I'm hungry

.github/CONTRIBUTING.md
```diff
@@ -67,4 +67,31 @@ The body should include the motivation for the change and contrast this with pre
...
+Release Version (YYYY-MM-DD)
+SDK
+- <description>
+<BLANK LINE>
+Sample App
+- <description>
+```
+
+### Description
+A sentence describes the type of change with an information.
```
ts-nitish-a-jain, line 85:
```suggestion
A sentence describing the type of change with necessary and relevant information.
```

.github/CONTRIBUTING.md
```diff
@@ -67,4 +67,31 @@ The body should include the motivation for the change and contrast this with pre
...
+
+```
+<bold><change type></bold>: <subject>
+```
+
+#### Change type
+* **Feature**: Added a new feature.
+* **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too.
+* **Deprecated**: Still available for use but will no longer be developed or supported. This is an indication it should not be used and will be removed in the next major version release.
+* **Obsoleted**: No longer works as expected since a better alternative is available and will maybe discontinued from the next releases.
```
ts-nitish-a-jain, line 95: What is the use case of `Obsoleted`? I think maybe this can be removed.

.github/CONTRIBUTING.md
```diff
@@ -67,4 +67,31 @@ The body should include the motivation for the change and contrast this with pre
...
+* **Removed**: When a feature has been expired by the sunsetting deadline.
+* Write "no changes" when there is no difference with previous release version.
```
ts-nitish-a-jain, line 97:
```suggestion
* Write "No Changes" when there is no difference with the previous release version.
```

.github/CONTRIBUTING.md
```diff
@@ -67,4 +67,31 @@ The body should include the motivation for the change and contrast this with pre
...
+### Description
+A sentence describes the type of change with an information.
+
+```
+<bold><change type></bold>: <subject>
+```
+
+#### Change type
+* **Feature**: Added a new feature.
+* **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too.
```
ts-nitish-a-jain, line 93: The improvements can be contained either in Feature or in Improvements IMO, where it is more logical, considering their enhancing nature.
Whereas a fix is more of the type of making the existing thing work.

ts-nitish-a-jain:
Nice attempt..! I have recorded some comments.

.github/CONTRIBUTING.md
khairul-alam-licon: `Obsoleted` here is a previous step of `Removed` phase which means kinda out-of-use. I don't find any "Obsoleted" annotation in Android by default yet, rather Android uses [Deprecated](https://developer.android.com/reference/java/lang/Deprecated) to define the situation. 

in iOS, I am not sure if we need to use `Obsoleted`. 
cc: @Climbatize @ts-leojoseph-r

.github/CONTRIBUTING.md
khairul-alam-licon: Yeah, I am okay with use of "Improvement" keyword instead of "Fixed".

khairul-alam-licon:
Thanks for your opinion guys!

PR #123, opened 2020-07-29
Refactor to a template for build info string in settings

khairul-alam-licon:
# Description
To refactor the build info string by template in sample app settings screen.

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
@@ -29,10 +29,6 @@ class SettingsMenuActivity : BaseActivity() {
     private lateinit var settings: AppSettings
     private lateinit var settingsProgressDialog: SettingsProgressDialog
 
-    companion object {
-        private const val buildInfoTemplate = "Build %s - %s"
-    }
-
     private var saveViewEnabled by Delegates.observable(true) { _, old, new ->
         if (new != old) {
             invalidateOptionsMenu()
@@ -108,11 +104,9 @@ class SettingsMenuActivity : BaseActivity() {
     }
 
     private fun createBuildInfo(): String {
-        return String.format(
-            buildInfoTemplate,
-            getString(R.string.miniapp_sdk_version),
-            getString(R.string.build_version)
-        )
+        val sdkVersion = getString(R.string.miniapp_sdk_version)
+        val buildVersion = getString(R.string.build_version)
+        return "Build $sdkVersion - $buildVersion"
     }
 
     internal fun validateInputIDs() {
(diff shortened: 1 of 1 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
```diff
@@ -29,6 +29,10 @@ class SettingsMenuActivity : BaseActivity() {
     private lateinit var settings: AppSettings
     private lateinit var settingsProgressDialog: SettingsProgressDialog
 
+    companion object {
```
minh-rakuten, line 32: We don't need a static define here, do we?

khairul-alam-licon: `companion object` has a support to keep static fields inside. In this PR, I have used `const` as an alternative of `@JvmField`.  I could use `@JvmStatic`, but currently my intension is using it only inside this class as private.

minh-rakuten: why do we need `companion object` here? I suppose the the normal private val declaration is enough 🤔

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
```diff
@@ -103,9 +107,12 @@ class SettingsMenuActivity : BaseActivity() {
         validateInputIDs()
     }
 
-    private fun createSettingsInfo(): String {
-        return "Build " + getString(R.string.miniapp_sdk_version) + " - " +
-                getString(R.string.build_version)
+    private fun createBuildInfo(): String {
+        return String.format(
```
corycaywood, line 111: I guess it's personal preference, but I prefer this style of String templating because it's easier to understand what the final string will look like: 

```
val sdkVersion = getString(R.string.miniapp_sdk_version)
val buildVersion = getString(R.string.build_version)
return "Build $sdkVersion - $buildVersion"
```

But I don't have a strong opinion, so either way is okay with me.

khairul-alam-licon: Yeah, this template is robust while reading. I have updated it. Thank you.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
khairul-alam-licon: Yeah, we can avoid `companion object` also. I have updated the PR.

PR #124, opened 2020-07-30
Fix to load miniapps after save settings

khairul-alam-licon:
# Description
Miniapps are not loading after saving the settings. Sample app has to relaunch to see the miniapps. This PR will fix the issue. 

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/miniapplist/MiniAppListActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/miniapplist/MiniAppListActivity.kt
@@ -1,5 +1,6 @@
 package com.rakuten.tech.mobile.testapp.ui.miniapplist
 
+import android.app.Activity
 import android.app.SearchManager
 import android.content.Intent
 import android.os.Build
@@ -20,6 +21,10 @@ class MiniAppListActivity : MenuBaseActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.mini_app_list_activity)
+        launchMiniApps()
+    }
+
+    private fun launchMiniApps() {
         if (AppSettings.instance.isSettingSaved) {
             layoutTut.visibility = View.GONE
             supportFragmentManager.beginTransaction()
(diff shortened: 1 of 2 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/miniapplist/MiniAppListActivity.kt
```diff
@@ -39,6 +39,7 @@ class MiniAppListActivity : MenuBaseActivity() {
         val intent = Intent(this, SettingsMenuActivity::class.java)
         intent.putExtra(MENU_SCREEN_NAME, MINI_APP_LIST_ACTIVITY)
         startActivity(intent)
+        finish()
```
Climbatize, line 42: Shouldn't you call a startActivityForResult and check if the setting screen has not been canceled ?

khairul-alam-licon: That's another solution. `startActivityForResult` can be used when we will have a spec to carry data.
Since now, it's a matter of reloading the entire screen after saved settings, we can go with current implementation.

corycaywood: But it seems using `finish` will cause another issue: tapping the back button from the settings screen will close the App because the previous screen has been finished. I'm not really sure what the best solution is. You could use `startActivityForResult` or find some other fix.

khairul-alam-licon: Yeah, I got it. I have updated this using `startActivityForResult`, for now I am using `RESULT_CANCELED` instead of `RESULT_OK` as I don't need to carry any data while returning back from settings.

khairul-alam-licon:
> Can you specify the commit that caused this bug? It's better to know the code change in previous commit.
Yeah, previously when there was settings dialog, user could see the MiniApp list after click on "SAVE" button after fresh launch. But, the feature was broken after refactoring settings. This PR will fix the issue.

khairul-alam-licon:
> Can you specify the commit that caused this bug? It's better to know the code change in previous commit.

Yeah, previously when there was settings dialog, user could see the MiniApp list after click on "SAVE" button after fresh launch. But, the feature was broken after refactoring settings. This PR will fix the issue.

PR #127, opened 2020-08-05
Update min SDK to 23

corycaywood:
# Description
Also I realized that the contribution guide for the changelog was missing a type for changes to already existing features, so I've added the `Changed` type.

## Links
Add links to github/jira issues, design documents and other relevant resources (e.g. previous discussions, platform/tool documentation etc.)

# Checklist

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -89,8 +89,9 @@ A sentence describing the type of change with necessary and relevant information
 ```
 
 #### Change type
-* **Feature**: Added a new feature.
-* **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too.
+* **Feature**: Added a new feature or changed an existing feature.
+* **Fixed**: When there is a hotfix or bugfix.
+* **Upgraded:** Upgraded a dependency or other version.
 * **Deprecated**: Still available for use but will no longer be developed or supported. This is an indication it should not be used and will be removed in the next major version release.
 * **Obsoleted**: No longer works as expected since a better alternative is available and will maybe discontinued from the next releases.
 * **Removed**: When a feature has been expired by the sunsetting deadline.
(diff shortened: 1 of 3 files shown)


.github/CONTRIBUTING.md
```diff
@@ -90,8 +90,9 @@ A sentence describing the type of change with necessary and relevant information
 
 #### Change type
 * **Feature**: Added a new feature.
+* **Changed:**: Changes to already existing features.
 * **Fixed**: When there is a hotfix or bugfix. Something has been upgraded could be mentioned too.
```
khairul-alam-licon, line 94: Can we append `Something has been upgraded could be mentioned too.` text with "Changed:" type mentioned above?  I feel that we also could avoid using `**Changed:**:` keyword. I am still finding a good keyword.

corycaywood: I don't think that `Fixed` is the right usage for this. `Fixed` means that there was a bug or some behavior that was incorrectly. There are some cases where the behavior was correct as originally defined, but something was changed due to product or technical requirements. So I would prefer to also remove the ` Something has been upgraded could be mentioned too.` line from `Fixed:`.

I don't think `Chore` and `Refactor` should be added because these types of tasks are only relevant to the SDK developers, not to the users of the SDK. If we did the refactoring correctly, then the SDK user shouldn't notice that there's been any refactoring.

Maybe we can add an `Upgrade:` for things like dependency and version upgrades? And then change Feature to this: `**Feature:** Added a new feature or changed an existing feature`.

What do you think?

khairul-alam-licon: > So I would prefer to also remove the "Something has been upgraded could be mentioned too." line from Fixed:

I agree. 

> I don't think Chore and Refactor should be added because these types of tasks are only relevant to the SDK developers, not to the users of the SDK

Makes sense!

> Maybe we can add an Upgrade: for things like dependency and version upgrades? And then change Feature to this: **Feature:** Added a new feature or changed an existing feature

Yep (+1).

PR #130, opened 2020-08-05
Use shared Orbs in CircleCI config 

corycaywood:
# Description
Updated the CircleCI config to use our shared [CircleCI Orbs](https://github.com/rakutentech/android-buildconfig/blob/master/circleci/README.md).

Also updated the `config` submodule to the latest version and applied the config for publishing to github pages.

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,19 @@
 version: 2.1
 
 orbs:
-  android-sdk: rakutentech/android-sdk@0.1.0
+  android-sdk: rakutentech/android-sdk@0.2.0
   app-center: rakutentech/app-center@0.1.0
 
 commands:
-  updateSubmodule: set -e
-    git submodule update --init
-    cp miniapp/src/main/assets/js-miniapp/js-miniapp-bridge/export/android/bridge.js miniapp/src/main/assets
-    rm -r miniapp/src/main/assets/js-miniapp
-    (cd miniapp/src/main/assets && mkdir -p js-miniapp/js-miniapp-bridge/export/android)
-    mv miniapp/src/main/assets/bridge.js miniapp/src/main/assets/js-miniapp/js-miniapp-bridge/export/android/
+  update-submodules:
+    steps:
+      - run: |
+          set -e
+          git submodule update --init
+          cp miniapp/src/main/assets/js-miniapp/js-miniapp-bridge/export/android/bridge.js miniapp/src/main/assets
+          rm -r miniapp/src/main/assets/js-miniapp
+          (cd miniapp/src/main/assets && mkdir -p js-miniapp/js-miniapp-bridge/export/android)
+          mv miniapp/src/main/assets/bridge.js miniapp/src/main/assets/js-miniapp/js-miniapp-bridge/export/android/
 
 workflows:
   version: 2.1
@@ -32,7 +35,7 @@ workflows:
                   base64 -d \<<< "$RELEASE_KEYSTORE_BASE64" > ./testapp/release-keystore.jks
                 fi
           after-prepare-steps:
-            - updateSubmodule
+            - update-submodules
           filters:
             tags:
               only: /^v.*/
@@ -63,7 +66,7 @@ workflows:
           requires:
             - release-verification
           after-prepare-steps:
-            - updateSubmodule
+            - update-submodules
           filters:
             tags:
               only: /^v.*/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,198 +1,83 @@
-version: 2
+version: 2.1
 
-gradle_cache_key: &gradle_cache_key gradle-{{ checksum "build.gradle" }}-{{ checksum "miniapp/build.gradle" }}
-maven_cache_key: &maven_cache_key maven-{{ checksum "miniapp/src/test/AndroidManifest.xml" }}
+orbs:
+  android-sdk: rakutentech/android-sdk@0.1.0
+  app-center: rakutentech/app-center@0.1.0
 
-jobs:
-  build:
-    docker:
-      - image: circleci/android@sha256:45eb47efaf498cdbf01b16e6a294ef20a4979f6acd6aaf75913707190806bb4c
-    working_directory: ~/code
-    environment:
-      JVM_OPTS: "-Xmx3200m"
-    steps:
-      - checkout
-      # ToDo Find a solution to upgrade git version of Debian to use sparse-checkout.
-      - run:
-          name: Update submodule
-          command: |
-            set -e
-            git submodule update --init
-            cp miniapp/src/main/assets/js-miniapp/js-miniapp-bridge/export/android/bridge.js miniapp/src/main/assets
-            rm -r miniapp/src/main/assets/js-miniapp
-            (cd miniapp/src/main/assets && mkdir -p js-miniapp/js-miniapp-bridge/export/android)
-            mv miniapp/src/main/assets/bridge.js miniapp/src/main/assets/js-miniapp/js-miniapp-bridge/export/android/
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,26 +1,97 @@
...
 
-## How to test the Sample app
+Next, you must define a few settings used by the Sample App. These can be defined either as environment variables or Gradle properties (in your `~/.gradle/gradle.properties` file).
 
-We are still working on this, please watch this space for future updates.
+```
+MINIAPP_SERVER_BASE_URL=https://www.example.com/
+HOST_APP_ID=test-host-app-id
+HOST_APP_SUBSCRIPTION_KEY=test-subs-key
+HOST_APP_UA_INFO=MiniAppDemoApp/1.0.0
```
minh-rakuten, line 25: Because we add `HOST_APP_UA_INFO` as an example here so we should include it in the example of production config too. I also noticed that we use the same `HOST_APP_UA_INFO` in build.gradle for `stg` and `prod`. We can update it to `HOST_APP_PROD_UA_INFO` in `prod`.

miniapp/build.gradle
```diff
@@ -95,15 +95,21 @@ dokka {
 }
 
 apply from: '../config/publish/android.gradle'
-publishing {
-    publications {
-        MiniApp(MavenPublication, androidArtifact())
+afterEvaluate {
```
minh-rakuten, line 98: Why do we need `afterEvaluate` block here 🤔 ?

.circleci/config.yml
```diff
@@ -1,181 +1,71 @@
...
+            {{ checksum "gradle/wrapper/gradle-wrapper.properties" }}
+          maven-cache-key: maven-{{ checksum "miniapp/src/test/AndroidManifest.xml" }}
+          sdk-path: miniapp
+          sample-app-path: testapp
+          pre-steps:
+            # Retrieve Base64 Keystore from Env variable and save to testapp project
+            - run: |
+                if [[ "$RELEASE_KEYSTORE_BASE64" != "" ]]; then
+                  base64 -d \<<< "$RELEASE_KEYSTORE_BASE64" > ./testapp/release-keystore.jks
+                fi
```
NemoOudeis, line 25: out of curiosity: this keystore is only for assembling a release variant of the test app, right? 
If so - why do we treat it as a secret? (my assumption is that injecting the base64 encoded keystore is to keep it secret from the source code)

README.md
```diff
@@ -1,26 +1,97 @@
...
+The following describes the steps performed by CI when releasing a new version of the SDK.
+
+1. A Git tag is pushed to repo which is in the format `vX.X.X` and triggers the CI to start.
+2. Build SDK and Sample App, run tests, linting, etc.
+3. Pause CI workflow for verification from user.
+4. After approval, publish the SDK to [Bintray](https://bintray.com/ssed-oss-jcenter/ssed-mobile-libs/android-miniapp).
+5. Publish generated documentation to [Github Pages site](https://rakutentech.github.io/android-miniapp/).
+6. Publish production build of Sample App to App Center's "Production" group.
+
+*Note:* It's also possible to publish a snapshot version of the SDK by using a `-` in the version name, such as `v1.0.0-alpha` (and the generated version name will be `1.0.0-SNAPSHOT`). In this case, the snapshot will be published to the [JFrog OSS Snapshot repository](https://oss.jfrog.org/) and the documentation publishing will be skipped.
```
NemoOudeis, line 89: A general note on writing: try to avoid passive voice. (see also [google's tech writing training](https://developers.google.com/tech-writing/one/active-voice))

For example in the release section is a mix of active and passive.

```diff
- The following describes the steps performed by CI when releasing a new version of the SDK.
+ The following describes the steps CircleCI performs to release a new version of the SDK.

- 1. A Git tag is pushed to repo which is in the format `vX.X.X` and triggers the CI to start.
+ 1. We trigger CircleCI by pushing a git tag to this repo which is in the format `vX.X.X`.
2. Build SDK and Sample App, run tests, linting, etc.
3. Pause CI workflow for verification from user.
4. After approval, publish the SDK to [Bintray](https://bintray.com/ssed-oss-jcenter/ssed-mobile-libs/android-miniapp).
5. Publish generated documentation to [Github Pages site](https://rakutentech.github.io/android-miniapp/).
6. Publish production build of Sample App to App Center's "Production" group.
```

I've gotten used to using "we", but it could also be "you" or "one" (<-- I kinda don't like that tho).

miniapp/build.gradle
NemoOudeis: yeah that looks surprising - if we didn't need it before, why do we need it now?

README.md
corycaywood: Yeah, that's a good point. I have added `HOST_APP_PROD_UA_INFO` as a config option.

README.md
corycaywood: That's a good idea. I sometimes have a habit of switching between active and passive when writing.

miniapp/build.gradle
corycaywood: AGP added built-in support for maven publishing of android artifacts, so we've switched to using that for the android-buildconfig repo. 

Previously, we were using [this plugin](https://github.com/wupdigital/android-maven-publish) which allows you to set the android components during Gradle's evaluation phase. However, with AGP's publishing plugin, the android components can't be accessed until `afterEvaluate` ([see here](https://developer.android.com/studio/build/maven-publish-plugin)).

.circleci/config.yml
corycaywood: Yeah, it is just for the test app so it's not really critical that it's kept secret. But it seems better to be on the safe side and keep it secret if possible. We distribute the release variant to external customers to use for testing Mini Apps.

nitish-a-jain (commented):
It is mostly looking good..! 👍

README.md
```diff
@@ -1,26 +1,98 @@
...
+
+The following describes the steps CircleCI performs when a branch is merged to master.
+
+1. We trigger CircleCI by merging a branch to master.
+2. CI builds the SDK and Sample App, run tests, linting, etc.
+3. CI publishes staging build of Sample App to App Center "Testers" group.
+
+### Release
+
+The following describes the steps CircleCI performs to release a new version of the SDK.
```
nitish-a-jain, line 81: I think we can just mention something like:
The releases are managed by `rakutentech` aligned with the product roadmap, with SDK binary published to bintray and generated docs to GH Pages site.
What do you think?

minh-rakuten (approved):

README.md
```diff
@@ -1,28 +1,101 @@
...
+
+1. We trigger CircleCI by pushing a git tag to this repo which is in the format `vX.X.X`.
+2. CI builds SDK and Sample App, run tests, linting, etc.
+3. CI pauses the workflow for verification from user.
+4. After approval, CI publishes the SDK to [Bintray](https://bintray.com/ssed-oss-jcenter/ssed-mobile-libs/android-miniapp).
+5. CI publishes generated documentation to [Github Pages site](https://rakutentech.github.io/android-miniapp/).
+6. CI publishes production build of Sample App to App Center's "Production" group.
+
+*Note:* It's also possible to publish a snapshot version of the SDK by using a `-` in the version name, such as `v1.0.0-alpha` (and the generated version name will be `1.0.0-SNAPSHOT`). In this case, the snapshot will be published to the [JFrog OSS Snapshot repository](https://oss.jfrog.org/) and the documentation publishing will be skipped.
```
khairul-alam-licon, line 93: Do we need to mention jfrog url as https://oss.jfrog.org/webapp/#/artifacts/browse/tree/General/oss-snapshot-local/com/rakuten/tech/mobile/miniapp/miniapp instead of https://oss.jfrog.org/ 🤔

khairul-alam-licon (approved):
LGTM 👍

PR #133, opened 2020-08-10
fix: create intermediate directory when unzip

minh-rakuten:
# Description
- When doing decompression, it's mandatory to create the intermediate directory container of the zip entry.
`JS-SDK-Sample` zip and other zips which have the same issue will load successfully from this fix. 
- This fix will also allow android sdk to extract the hidden folder.

# Checklist


minh-rakuten:
> 
> 
> Looks good 👍
> 
> What is the specific situation that this fixes? Is it when a Mini App has sub-folders inside the zip file?

@corycaywood There is a case that the zip entry (file to be extracted) goes first before the folder contains that zip entry.

For example, file to be extracted is `js/main.js` but the `js` folder has not been created yet. The zip entry of `js` goes after `js/main.js` in decompression.

PR #140, opened 2020-08-28
docs: Fix some issues in the user guide

corycaywood:
# Description
The MiniApp#create function was still using the old method (which has now been removed). Also removed the note about AppCompat because this was relevant to Android 22 and below only (we now support 23+).

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -36,10 +36,10 @@ The SDK is configured via manifest meta-data, the configurable values are:
 | Field                        | Datatype| Manifest Key                                           | Optional   | Default  |
 |------------------------------|---------|--------------------------------------------------------|----------- |--------- |
 | Base URL                     | String  | `com.rakuten.tech.mobile.miniapp.BaseUrl`              | ❌         | 🚫        |
-| Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
-| Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
 | App ID                       | String  | `com.rakuten.tech.mobile.ras.AppId`                    | ❌         | 🚫        |
 | RAS Project Subscription Key | String  | `com.rakuten.tech.mobile.ras.ProjectSubscriptionKey`   | ❌         | 🚫        |
+| Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
+| Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
 
 **Note:**  
 * We don't currently host a public API, so you will need to provide your own Base URL for API requests.
@@ -88,7 +88,7 @@ Information about Mini Apps can be fetched in two different ways: by using `Mini
 Use `MiniApp.listMiniApp` if you want a list of all Mini Apps:
 
 ```kotlin
-CoroutineScope(Dispatchers.Default).launch {
+CoroutineScope(Dispatchers.IO).launch {
     try {
         val miniAppList = MiniApp.instance().listMiniApp()
     } catch(e: MiniAppSdkException) {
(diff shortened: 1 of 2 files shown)


minh-rakuten (commented):
Thank you for updating Userguide.

`Note: This SDK uses suspend functions, so you should use Kotlin Coroutines when calling the functions. These examples use Dispatchers.Default, but you can use whichever CoroutineContext and CouroutineScope that is appropriate for your App as long as it does not run in the main thread.`
-> I'm sorry that I miss this in document before. We should also specify that the Miniapp api call should not be executed in main thread (UI).

corycaywood:
> Thank you for updating Userguide.
> 
> `Note: This SDK uses suspend functions, so you should use Kotlin Coroutines when calling the functions. These examples use Dispatchers.Default, but you can use whichever CoroutineContext and CouroutineScope that is appropriate for your App as long as it does not run in the main thread.`
> -> I'm sorry that I miss this in document before. We should also specify that the Miniapp api call should not be executed in main thread (UI).

Yeah, good catch. I've updated this and also added this to the `TroubleShooting` section.

Also, I updated the `MiniAppSdkException` so that it will set a custom message when `NetworkOnMainThreadException` is thrown because that exception has a `null` message.

miniapp/USERGUIDE.md
```diff
@@ -37,9 +37,9 @@ The SDK is configured via manifest meta-data, the configurable values are:
 |------------------------------|---------|--------------------------------------------------------|----------- |--------- |
 | Base URL                     | String  | `com.rakuten.tech.mobile.miniapp.BaseUrl`              | ❌         | 🚫        |
 | Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
-| Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
 | App ID                       | String  | `com.rakuten.tech.mobile.ras.AppId`                    | ❌         | 🚫        |
 | RAS Project Subscription Key | String  | `com.rakuten.tech.mobile.ras.ProjectSubscriptionKey`   | ❌         | 🚫        |
+| Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
```
nitish-a-jain, line 42: I think Host App Version can also be grouped downwards in the list.
```suggestion
| Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
| Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
```

miniapp/USERGUIDE.md
```diff
@@ -37,9 +37,9 @@ The SDK is configured via manifest meta-data, the configurable values are:
 |------------------------------|---------|--------------------------------------------------------|----------- |--------- |
 | Base URL                     | String  | `com.rakuten.tech.mobile.miniapp.BaseUrl`              | ❌         | 🚫        |
 | Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
```
nitish-a-jain, line 39:
```suggestion
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkException.kt
```diff
@@ -29,3 +35,9 @@ internal fun sdkExceptionForInvalidArguments(message: String = "") =
 @Suppress("FunctionMaxLength")
 internal fun sdkExceptionForNoActivityContext() =
     MiniAppSdkException("Only accept context of type Activity or ActivityCompat")
+
+private fun exceptionMessage(exception: Exception) = when (exception) {
+    is NetworkOnMainThreadException -> "Network requests must not be performed on the main thread. " +
```
nitish-a-jain, line 40: How about template string or `.plus`?

nitish-a-jain (approved):
👍 
Just minor comments for consideration.

minh-rakuten (approved):
Looks good.

PR #143, opened 2020-09-07
feature: share content

minh-rakuten:
# Description
SDK:
Interface for receiving shared content and the controller for hostapp. The object from miniapp is expected to be
`{content: "this is the shared text"}`

Sample App:
Handle the shared content with `ACTION_SEND`.

## Links
MINI-1594

# Checklist


corycaywood (commented):
Looks good 👍 Could you also add this to the user guide?

Also, maybe we should add this to js-miniapp-bridge first so we can update the submodule here before merging this?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -121,19 +120,23 @@ class MiniAppDisplayActivity : BaseActivity() {
...
                     )
                 }
+
+                override fun share(content: String) {
+                    val sendIntent: Intent = Intent().apply {
+                        action = Intent.ACTION_SEND
+                        putExtra(Intent.EXTRA_TEXT, content)
+                        type = "text/plain"
+                    }
+                    startActivity(sendIntent)
```
khairul-alam-licon, line 130: if you use `startActivity(Intent.createChooser(sendIntent, null))` then, you can adopt some extra features provided by `Intent`.

khairul-alam-licon (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -111,6 +115,17 @@ abstract class MiniAppMessageBridge {
         }
     }
 
+    private fun onShareContent(callbackId: String, jsonStr: String) {
+        try {
+            val callbackObj = Gson().fromJson(jsonStr, ShareInfoCallbackObj::class.java)
+
+            shareContent(callbackObj.param.shareInfo.content)
+            postValue(callbackId, "Share content successfully")
```
khairul-alam-licon, line 123: "Content has been shared successfully." would be good? 🤔

khairul-alam-licon: It's better to send this message after confirming the successfully execution from the HostApp. In that case, we need to rewrite the `abstract`  function as following:

```
abstract fun shareContent(
        content: String,
        callback: (String) -> Unit 
)
```

Therefore, we can invoke `callback` for passing a message of share result from HostApp.

corycaywood: How about passing back some some constant value like `SUCCESS`?

Also, I agree with @khairul-alam-licon that we should have some way for the Host App to asynchronously send the callback success.

minh-rakuten: Updated with status callback.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -40,6 +40,16 @@ abstract class MiniAppMessageBridge {
         callback: (List<Pair<MiniAppCustomPermissionType, MiniAppCustomPermissionResult>>) -> Unit
     )
 
+    /**
+     * Share content info [ShareInfo]. This info is provided by mini app.
+     * @param content The content property of [ShareInfo] object.
+     * @param callback The executed action status should be notified back to mini app.
+     **/
+    abstract fun shareContent(
```
corycaywood, line 48: I was thinking: maybe we should add the new methods in `MiniAppMessageBridge` as normal functions which can be overridden instead of as `abstract` functions? And then if the Host App fails to override the method, then the SDK will return an error message to the Mini App when they try to call it.

That is what the iOS SDK is doing (though I think this is due to a technical limitation of Obj-c/Swift): https://github.com/rakutentech/ios-miniapp/blob/2e46a3d57ab3f717717c771a211565d9aaaf7678/MiniApp/Classes/RealMiniApp.swift#L149

The reason I was thinking this is:
1. It's technically a breaking change when we add new methods which must be implemented - the Host App will fail to compile if they update to the latest version of the SDK but don't implement the new method.
2. It lowers the effort required for a minimal integration of the SDK.

@minh-rakuten @khairul-alam-licon What do you think? You don't have to do that in this PR, I just wanted to discuss it and maybe we can do this in another PR.

corycaywood (approved):
Looks good 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: I think, it's a good idea as users don't need to implement every functions/feature from SDK. We can address this improvement in a separate PR.

khairul-alam-licon (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: We can go that way and have a ticket for another PR 👍.

Our current implementation has an advantage that let developers aware of what they are required to implement and those features cannot be missed. I believe some developers don't even read document and we can avoid some false bug reports. For example, there is a miniapp does not work as expected, people mind will think it is sdk fault even though the fact is the problem on host app side. Of course, it's the issue from host app so either way is fine for me.


@khairul-alam-licon Actually, all features from `MiniAppMessageBridge` are required from miniapps. The miniapp itself surely wants their features working on hostapp so users must implement it.

khairul-alam-licon: Well, we can identify all the required and non-required features firstly, maybe the miniapp teams can help by sharing their needs. If all the features are required, then we don't have to change anything from current implementation. Otherwise, we can make non-required features optional to be implemented.

corycaywood: But I'm more concerned about introducing breaking changes. When we add a new `abstract` function, it's a breaking change because the Host App's build will fail if they update the SDK and don't update their code. 

I don't know if my suggestion is the best solution, so feel free to think about any other solutions. Basically this is the problem:
- How do we introduce new features to the JavaScript bridge without making breaking changes every time?

minh-rakuten: I think your suggestion is good. Alam has a good point on determine which one is required / non-required.

Do you think that I should convert all abstract functions to normal one? or only `shareContent` feature?

corycaywood: Probably you can convert just the functions that are new since the last release: `shareContent` and `requestCustomPermissions`.

Actually, I think it will make sense to give these functions (and most of the bridge functionality) some default functionality. So they will still work using the default functionality even if the Host App doesn't implement them. We already have a ticket to do that for `requestCustomPermissions`, but probably we can do it sometime later for `shareContent`.

PR #146, opened 2020-09-09
api: accept content encoding

minh-rakuten:
# Description
Bypass the problem of returning incorrect `Content-Length` from server.

We can test this by changing api endpoint which was posted on chat thread.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
@@ -5,6 +5,7 @@ import com.google.gson.GsonBuilder
 import com.rakuten.tech.mobile.miniapp.BuildConfig
 import com.rakuten.tech.mobile.sdkutils.RasSdkHeaders
 import com.rakuten.tech.mobile.sdkutils.okhttp.addHeaderInterceptor
+import okhttp3.Interceptor
 import okhttp3.OkHttpClient
 import retrofit2.Retrofit
 import retrofit2.converter.gson.GsonConverterFactory
@@ -31,6 +32,7 @@ internal fun createRetrofitClient(
     @Suppress("SpreadOperator")
     val httpClient = OkHttpClient.Builder()
         .addHeaderInterceptor(*headers.asArray())
+        .addInterceptor(provideHeaderInterceptor())
         .build()
     return Retrofit.Builder()
         .addConverterFactory(
@@ -41,4 +43,16 @@ internal fun createRetrofitClient(
         .baseUrl(baseUrl)
         .client(httpClient)
         .build()
+}
+
+private fun provideHeaderInterceptor(): Interceptor {
+    return Interceptor { chain ->
+        val original = chain.request()
+
+        val request = original.newBuilder()
+            .header("Accept-Encoding", "identity")
+            .build()
+
+        return@Interceptor chain.proceed(request)
+    }
 }
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
@@ -35,24 +35,17 @@ internal fun createRetrofitClient(
         .addInterceptor(provideHeaderInterceptor())
         .build()
     return Retrofit.Builder()
-        .addConverterFactory(
-            GsonConverterFactory.create(
-                GsonBuilder().setLenient().create()
-            )
-        )
+        .addConverterFactory(GsonConverterFactory.create(GsonBuilder().setLenient().create()))
         .baseUrl(baseUrl)
         .client(httpClient)
         .build()
 }
 
-private fun provideHeaderInterceptor(): Interceptor {
-    return Interceptor { chain ->
-        val original = chain.request()
+private fun provideHeaderInterceptor(): Interceptor = Interceptor { chain ->
+    val original = chain.request()
+    val request = original.newBuilder()
+        .header("Accept-Encoding", "identity")
+        .build()
 
-        val request = original.newBuilder()
-            .header("Accept-Encoding", "identity")
-            .build()
-
-        return@Interceptor chain.proceed(request)
-    }
+    chain.proceed(request)
 }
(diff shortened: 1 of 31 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
```diff
@@ -31,14 +32,20 @@ internal fun createRetrofitClient(
...
+        .addConverterFactory(GsonConverterFactory.create(GsonBuilder().setLenient().create()))
         .baseUrl(baseUrl)
         .client(httpClient)
         .build()
 }
+
+private fun provideHeaderInterceptor(): Interceptor = Interceptor { chain ->
+    val original = chain.request()
+    val request = original.newBuilder()
+        .header("Accept-Encoding", "identity")
```
corycaywood, line 47: Was this header being automatically set to something else by OkHttp? Or it just wasn't being set at all?

Also, from what I see [here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding), `identity` means that the server will never send a compressed response. The Mini App is downloaded as a compressed ZIP file anyways, so I don't think that's a big deal. 

But it's possible that the `listing` endpoint can return a very long list, so the request will take longer when uncompressed. I guess the listing endpoint doesn't need to be called very often, so probably it's not a big deal. What do you guys think?

minh-rakuten: This header will be included on every api call request by OkHttp.

Yeah, there is not much impact on performance. For now, there are a lot of miniapps in the list but I don't see the difference. What do you think @khairul-alam-licon ?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
```diff
@@ -31,14 +32,20 @@ internal fun createRetrofitClient(
     @Suppress("SpreadOperator")
     val httpClient = OkHttpClient.Builder()
         .addHeaderInterceptor(*headers.asArray())
+        .addInterceptor(provideHeaderInterceptor())
```
corycaywood, line 35: I think you should be able to use something like `.addHeaderInterceptor("Accept-Encoding" to "identity")` instead of manually defining your own interceptor. `.addHeaderInterceptor` is an extension function from the SDK Utils library (https://github.com/rakutentech/android-sdkutils#okhttp-header-interceptor).

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
Climbatize: in fact this could lead to an unexpected high traffic when MiniApp will be popular. But this leaves us time to think about it later when the problem will be raised, if it is ever raised. This fix can be considered until then, with a comment maybe.

Climbatize (approved):
I think this fix is good as long as backend team is aware that we are forced to do this, so they know that Android mini apps may hypothetically impact data traffic on servers.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
minh-rakuten: `.addHeaderInterceptor("Accept-Encoding" to "identity")` has no affect 🤔 . There is different that our sdk utils lib uses `addNetworkInterceptor`, not `addInterceptor`.

corycaywood: Ah, okay yeah you're right. A network interceptor is different than a normal interceptor.

corycaywood (dismissed):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
```diff
@@ -31,14 +32,20 @@ internal fun createRetrofitClient(
...
-                GsonBuilder().setLenient().create()
-            )
-        )
+        .addConverterFactory(GsonConverterFactory.create(GsonBuilder().setLenient().create()))
         .baseUrl(baseUrl)
         .client(httpClient)
         .build()
 }
+
+private fun provideHeaderInterceptor(): Interceptor = Interceptor { chain ->
```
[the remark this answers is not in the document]
khairul-alam-licon: Or, can we move this function to a different place e.g.

```
class MiniAppInterceptor : Interceptor {

    override fun intercept(chain: Interceptor.Chain): Response {
        val original = chain.request()
        val request = original.newBuilder()
            .header("Accept-Encoding", "identity")
            .build()
        return chain.proceed(request)
    }
}
```

corycaywood (commented):
Let's hold off on merging this for now.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
khairul-alam-licon: Looks like "Accept-Encoding", "identity" here is overriding the `Accept-Encoding:gzip` provided from RAS, so we are losing the `gzip` compression feature. I could see a few miniapps in STG to trace the performance issue. I am not sure if we could use a testing api which provides enough miniapps to test the traffic. But, of course, we can address the performance with long list in future, in that case, adding a TODO comment would be better.

khairul-alam-licon (commented):
Thank you for your attempt!

corycaywood (approved):
Seems this will be okay from the backend side, so this is okay to merge 👍 

@minh-rakuten Should we also add a unit test for this?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
minh-rakuten: We don't need the long list of miniapp to see the performance issue on sdk call side. Because if there is a long list of miniapps, the backend should do paging on that list.

minh-rakuten:
Add unit test.

PR #149, opened 2020-09-15
Fix: HostApp is crashing with NoSuchFieldException in staging 

khairul-alam-licon:
# Description
HostApp is crashing in `staging` with following exception when requesting custom permission e.g. `USER_NAME`

```
java.lang.AssertionError: java.lang.NoSuchFieldException: USER_NAME
        at com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter.<init>(:792)
        at com.google.gson.internal.bind.TypeAdapters$30.a(:818)
        at com.google.gson.Gson.k(:458)
       ......
Caused by: java.lang.NoSuchFieldException: USER_NAME
        at java.lang.Class.getField(Class.java:1601)
        at com.google.gson.internal.bind.TypeAdapters$EnumTypeAdapter.<init
```

This is not happening in `debug`. After reading few issues e.g. https://github.com/google/gson/issues/1776 and https://github.com/google/gson/issues/924, I have come out with two solutions:

- Solution-1: Set `minifyEnabled false` in [here](https://github.com/rakutentech/android-miniapp/blob/e5e15f3fb876b9af7064be2a379cfeb0b28b62c3/testapp/build.gradle#L72) as the [staging buildTypes](https://github.com/rakutentech/android-miniapp/blob/e5e15f3fb876b9af7064be2a379cfeb0b28b62c3/testapp/build.gradle#L77) takes some favors from `release`.
- Solution-2: Add `-keepclassmembers enum * { *; }` in `miniapp/proguard-rules.pro `
- Solution-3: Use `@Keep` annotation as @minh-rakuten suggested.

## Links
[MINI-1690](https://jira.rakuten-it.com/jira/browse/MINI-1690)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
@@ -27,7 +27,6 @@ enum class MiniAppCustomPermissionType(val type: String) {
     }
 }
 
-@Keep
 internal enum class MiniAppPermissionResult(val type: String) {
     ALLOWED("Allowed"),
     DENIED("Denied");
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
@@ -1,5 +1,7 @@
 package com.rakuten.tech.mobile.miniapp.permission
 
+import androidx.annotation.Keep
+
 /** Type of miniapp permission. **/
 enum class MiniAppPermissionType(val type: String) {
     UNKNOWN("unknown"),
@@ -12,6 +14,7 @@ enum class MiniAppPermissionType(val type: String) {
 }
 
 /** Type of miniapp custom permission. **/
+@Keep
 enum class MiniAppCustomPermissionType(val type: String) {
     USER_NAME("rakuten.miniapp.user.USER_NAME"),
     PROFILE_PHOTO("rakuten.miniapp.user.PROFILE_PHOTO"),
@@ -24,6 +27,7 @@ enum class MiniAppCustomPermissionType(val type: String) {
     }
 }
 
+@Keep
 internal enum class MiniAppPermissionResult(val type: String) {
     ALLOWED("Allowed"),
     DENIED("Denied");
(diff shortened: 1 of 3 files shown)


minh-rakuten (commented):
You can fix it by putting `@Keep` in enum class. Those enum classes are in `MiniAppPermission.kt`. You should check the data class of `MiniAppCustomPermission` too.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
```diff
@@ -1,5 +1,7 @@
 package com.rakuten.tech.mobile.miniapp.permission
 
+import androidx.annotation.Keep
+
 /** Type of miniapp permission. **/
 enum class MiniAppPermissionType(val type: String) {
```
corycaywood, line 6: Does this one also need `@Keep`?

corycaywood (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
khairul-alam-licon: Currently, there is no affect has been traced for this class without `@Keep`. But for safety we can address it here, what do you think @minh-rakuten?

minh-rakuten: @khairul-alam-licon I guess just putting `@Keep` on data class `MiniAppCustomPermission` will help you fix the problem, no need to update `@Keep` for enum in `MiniAppPermission.kt`. I'm not sure so let try it :)

minh-rakuten (approved):
👍 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
khairul-alam-licon: @minh-rakuten I understand your concern, as `MiniAppCustomPermission` has a `pairValues`. Here, I re-checked that, adding `@Keep` annotation only in `MiniAppCustomPermissionType` and `MiniAppCustomPermissionResult` helps actually. So, I updated this PR. For now, let's keep it to understand clearly where we require this annotation.

khairul-alam-licon:
Thanks guys for the reviews 👍

PR #150, opened 2020-09-15
JS Bridge updates for Custom Permission support

corycaywood:
# Description
- build: Update JS bridge submodule
  - This fixes an issue where the response was being returned to the user as as string rather than an object
- refactor: (Custom permissions) Rename 'isGranted' param to 'status'
  -  This param doesn't represent a boolean - it represents a status of allowed, denied, or not available so it should be named accordingly.
- fix: JS objects weren't being sent correctly over JS bridge
  - Objects were being sent with normal double quotes which causes problems because the keys/values in the object are surrounded by double quotes. So I have changed this to use backticks and also added a fix so that backticks will be escaped if included in the response value.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
@@ -78,7 +78,7 @@ internal class MiniAppWebView(
     override fun runSuccessCallback(callbackId: String, value: String) {
         post {
             evaluateJavascript(
-                "MiniAppBridge.execSuccessCallback(\"$callbackId\", \"$value\")"
+                "MiniAppBridge.execSuccessCallback(`$callbackId`, `${value.replace("`", "\\`")}`)"
             ) {}
         }
     }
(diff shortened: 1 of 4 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -78,7 +78,7 @@ internal class MiniAppWebView(
     override fun runSuccessCallback(callbackId: String, value: String) {
         post {
             evaluateJavascript(
-                "MiniAppBridge.execSuccessCallback(\"$callbackId\", \"$value\")"
+                "MiniAppBridge.execSuccessCallback(`$callbackId`, `${value.replace("`", "\\`")}`)"
```
khairul-alam-licon, line 81: Thanks 🙏

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionResponse.kt
```diff
@@ -11,6 +11,6 @@ data class MiniAppCustomPermissionResponse(
      */
     data class CustomPermissionResponseObj(
         val name: String,
-        val isGranted: String
+        val status: String
```
khairul-alam-licon, line 14: 👍

khairul-alam-licon (approved):
Thank you for the fix!

minh-rakuten (approved):
👍

PR #151, opened 2020-09-16
feature: AdMob setting

minh-rakuten:
# Description
- androidx_appcompat version stable 1.2.0.
- Check admob app id provided from hostapp.
When the hostapp implements AdMob library, Google requires app to config admob app id. That's how we can detect if hostapp wants to run ads or not.
- AdMob setting spec:

```
MUST allow the Host App to set the Ad Mob Application ID.
SHOULD be optional for the Host App to use Ads
Could add Ad Mob as a `compileOnly` dependency and require the host app to add the dependency themselves.
```

The `compileOnly` for AdMob prevents us to hit the test coverage on code regarding ads execution.

## Links
Reference from internal document Google Ad Mob in Mini Apps.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMob.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMob.kt
@@ -1,0 +1,14 @@
+package com.rakuten.tech.mobile.miniapp.ads
+
+/**
+ * Check whether hostapp provides AdMob dependency.
+ * If the dependency is provided then it means the AdMob app id has also been set from hostapp.
+ */
+inline fun <T> whenAdMobProvided(callback: () -> T) {
+    try {
+        Class.forName("com.google.android.gms.ads.MobileAds")
+        callback.invoke()
+    } catch (e: ClassNotFoundException) {
+        e.printStackTrace()
+    }
+}
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMob.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMob.kt
@@ -4,11 +4,10 @@ package com.rakuten.tech.mobile.miniapp.ads
  * Check whether hostapp provides AdMob dependency.
  * If the dependency is provided then it means the AdMob app id has also been set from hostapp.
  */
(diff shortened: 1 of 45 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -45,6 +47,10 @@ internal class MiniAppWebView(
             ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT
         )
 
+        whenAdMobProvided {
+            MobileAds.initialize(context)
```
corycaywood, line 51: Maybe we should require that the Host App call `MobileAds.initialize` in their own App initialization code? I was thinking that makes more sense because they also have to add the ad-mob dependency and the Ad Mob Application ID. It also gives them more control over when it's initialized. What do you think?

Are there any problems if `MobileAds.initialize` is called multiple times?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMob.kt
```diff
@@ -0,0 +1,13 @@
+package com.rakuten.tech.mobile.miniapp.ads
+
+/**
+ * Check whether hostapp provides AdMob dependency.
+ * If the dependency is provided then it means the AdMob app id has also been set from hostapp.
+ */
+@Suppress("EmptyCatchBlock", "SwallowedException")
+inline fun <T> whenAdMobProvided(callback: () -> T) {
```
corycaywood, line 8: Nice 👍

corycaywood (approved):
Looks good 👍  Just a question about initialization.

khairul-alam-licon (approved):
👍 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
minh-rakuten: There is no problem on calling `MobileAds.initialize` multiple times but let the hostapp control is good for reducing workload on MiniAppWebView -> then we need to handle the exception of AdMob if it has not been initialized.

PR #152, opened 2020-09-16
refactor: default share implementation

minh-rakuten:
# Description
Default implementation of Share feature in `MiniAppMessageBridge`.
Passing `Activity` to `MiniAppMessageBridge` as optional parameter. `Activity` is mandatory to execute some methods.

## Links
MINI-1695

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -125,7 +125,7 @@ There are some methods have default implementation but host app can override the
 | shareContent                 | ✅       |
 
 ```kotlin
-val miniAppMessageBridge = object: MiniAppMessageBridge(activity) {
+val miniAppMessageBridge = object: MiniAppMessageBridge() {
     override fun getUniqueId() {
         val id: String = ""
         // Implementation details to generate a Unique ID
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -115,9 +115,14 @@ CoroutineScope(Dispatchers.IO).launch {
 ### #4 Implement the MiniAppMessageBridge
 
 The `MiniAppMessageBridge` is used for passing messages between the Mini App (JavaScript) and the Host App (your native Android App) and vice versa. Your App must provide the implementation for these functions and pass this implementation to the `MiniApp#create` function.
-
 There are some methods have default implementation but host app can override them to fully control.
-The parameter `activity` is not required for `MiniAppMessageBridge` but it is essential to ensure all default methods work as expected.
+
+| Method                       | Default  |
+|------------------------------|----------|
+| getUniqueId                  | 🚫       |
+| requestPermission            | 🚫       |
+| requestCustomPermissions     | 🚫       |
+| shareContent                 | ✅       |
 
 ```kotlin
 val miniAppMessageBridge = object: MiniAppMessageBridge(activity) {
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -116,8 +116,11 @@ CoroutineScope(Dispatchers.IO).launch {
 
 The `MiniAppMessageBridge` is used for passing messages between the Mini App (JavaScript) and the Host App (your native Android App) and vice versa. Your App must provide the implementation for these functions and pass this implementation to the `MiniApp#create` function.
 
+There are some methods have default implementation but host app can override them to fully control.
+The parameter `activity` is not required for `MiniAppMessageBridge` but it is essential to ensure all default methods work as expected.
+
 ```kotlin
-val miniAppMessageBridge = object: MiniAppMessageBridge() {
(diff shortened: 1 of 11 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -16,7 +18,7 @@ import com.rakuten.tech.mobile.miniapp.permission.MiniAppPermissionResult
 
 @Suppress("TooGenericExceptionCaught", "SwallowedException", "TooManyFunctions")
 /** Bridge interface for communicating with mini app. **/
-abstract class MiniAppMessageBridge {
+abstract class MiniAppMessageBridge(val activity: Activity? = null) {
```
corycaywood, line 21: They already need to provide the activity context when calling `MiniApp.create`, right? So could you reuse that here? 

@khairul-alam-licon added a function for `MiniAppMessageBridge.init`, so maybe you could provide the context through that?

miniapp/USERGUIDE.md
```diff
@@ -116,8 +116,11 @@ CoroutineScope(Dispatchers.IO).launch {
 
 The `MiniAppMessageBridge` is used for passing messages between the Mini App (JavaScript) and the Host App (your native Android App) and vice versa. Your App must provide the implementation for these functions and pass this implementation to the `MiniApp#create` function.
 
+There are some methods have default implementation but host app can override them to fully control.
```
corycaywood, line 119: I think this is okay for now, but in the future we will need to reformat this section about `MiniAppMessageBridge`. I think it will get kind of long and confusing as we add more and more features 😅

corycaywood (commented):
Looking good, just one suggestion about the `Activity`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -45,10 +47,25 @@ abstract class MiniAppMessageBridge {
      * @param content The content property of [ShareInfo] object.
      * @param callback The executed action status should be notified back to mini app.
      **/
-    abstract fun shareContent(
+    open fun shareContent(
```
khairul-alam-licon, line 50: Just a suggestion: how about moving this function a bit later probably before/after permission functions to maintain a consistent sorting of features in this file?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -121,20 +121,6 @@ class MiniAppDisplayActivity : BaseActivity() {
                         callback
                     )
                 }
-
-                override fun shareContent(
```
khairul-alam-licon, line 125: How about keeping this function here?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Can we use this function as the [USERGUIDE](https://github.com/minh-rakuten/android-miniapp/blob/f9737d9a41b7b52864760325d6f34a1db703ef75/miniapp/USERGUIDE.md) mentioned?

minh-rakuten: Custom permission goes after permission function.
Yes, the user still can override it. MINI-1695 goal is to have default implementation but not prevent the override.

miniapp/USERGUIDE.md
minh-rakuten: Updated.

miniapp/USERGUIDE.md
```diff
@@ -115,9 +115,17 @@ CoroutineScope(Dispatchers.IO).launch {
...
+| Method                       | Default  |
+|------------------------------|----------|
+| getUniqueId                  | 🚫       |
+| requestPermission            | 🚫       |
+| requestCustomPermissions     | 🚫       |
+| shareContent                 | ✅       |
 
 ```kotlin
-val miniAppMessageBridge = object: MiniAppMessageBridge() {
+val miniAppMessageBridge = object: MiniAppMessageBridge(activity) {
```
khairul-alam-licon, line 128: I think, we can remove `activity` as parameter from now.

khairul-alam-licon (approved):
Nice 👍

khairul-alam-licon (approved):
👍

corycaywood (approved):
Looks good 👍 👍

PR #153, opened 2020-09-18
feat: Verify cached mini app files using SHA-256 hash

corycaywood:
# Description
This is to prevent a cache poisining vulnerability. This works by saving an encrypted hash of the Mini App fils. If someone modifies the cached files for the Mini App (i.e. on a rooted device), then the hash verification will fail and the Mini App will be re-downloaded. It's also not possible for someone to modify the hash since it's encrypted - if they attempt to modify it then the decryption of the hash will fail.

I also needed to delete some unit tests related to SDK initialization - this is because the CachedMiniAppVerifier needs to use the AndroidKeyStore which isn't mocked by the test runner. However I think it's okay to not test the initialization code because part of it's purpose is to be the place where we can initialize things which aren't mockable.

## Links
MINI-1405

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -1,9 +1,11 @@
 package com.rakuten.tech.mobile.miniapp
 
+import android.util.Log
 import androidx.annotation.VisibleForTesting
 import com.rakuten.tech.mobile.miniapp.api.ApiClient
 import com.rakuten.tech.mobile.miniapp.api.ManifestEntity
 import com.rakuten.tech.mobile.miniapp.api.UpdatableApiClient
+import com.rakuten.tech.mobile.miniapp.storage.CachedMiniAppVerifier
 import com.rakuten.tech.mobile.miniapp.storage.MiniAppStatus
 import com.rakuten.tech.mobile.miniapp.storage.MiniAppStorage
 import kotlinx.coroutines.CoroutineDispatcher
@@ -11,11 +13,13 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.Job
+import java.io.File
 
 internal class MiniAppDownloader(
     private val storage: MiniAppStorage,
     private var apiClient: ApiClient,
     private val miniAppStatus: MiniAppStatus,
+    private val verifier: CachedMiniAppVerifier,
     private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
 ) : UpdatableApiClient {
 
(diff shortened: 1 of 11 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/CachedMiniAppVerifier.kt
```diff
@@ -0,0 +1,88 @@
...
+                prefs.edit().putString(appId, hash).apply()
+            } catch (e: IOException) {
+                Log.e(TAG, "Failed to store hash for the Mini App.", e)
+            }
+        }
+    }
+
+    private fun calculateHash(
+        directory: File
+    ): String {
```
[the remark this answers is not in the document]
khairul-alam-licon: Also, we can catch the `IOException` in here instead of `verify(...)` and `storeHash(...)` function. My total idea would be -

```
private fun calculateHash(
    directory: File
): String = runBlocking {
        try {
             // calculating hash
        } catch (e: IOException) {
            throw MiniAppSdkException(e)
        }
 }
```

khairul-alam-licon (commented):
Thank you for adding this security layer, it looks great 👍
I have few comments to discuss few minor things.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -48,6 +57,24 @@ internal class MiniAppDownloader(
...
+    private suspend fun retrieveDownloadedVersionPath(miniAppInfo: MiniAppInfo): String? {
+        val versionPath = storage.getMiniAppVersionPath(miniAppInfo.id, miniAppInfo.version.versionId)
+
+        if (miniAppStatus.isVersionDownloaded(miniAppInfo.id, miniAppInfo.version.versionId, versionPath)) {
+            return if (verifier.verify(miniAppInfo.id, File(versionPath))) {
+                versionPath
+            } else {
+                Log.e(TAG, "Failed to verify the hash of the cached files. " +
+                        "The files will be deleted and the Mini App re-downloaded.")
+                storage.removeOutdatedVersionApp(miniAppInfo.id, NON_EXISTENT_VERSION_ID)
```
minh-rakuten, line 69: Why do we use `NON_EXISTENT_VERSION_ID`? Should it be `miniAppInfo.version.versionId`?

minh-rakuten (commented):
It's okay to remove the initialization code test but we need to exclude them in test coverage as well.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
minh-rakuten: I understand it now. I forget my own implementation on `removeOutdatedVersionApp` 😆. Please ignore my previous comment.

I think it's better to remove the whole package (miniapp id directory) of this miniapp.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/CachedMiniAppVerifier.kt
```diff
@@ -0,0 +1,88 @@
...
+                        digest.update(buffer, 0, count)
+                    }
+                }
+            }
+
+        return Base64.encodeToString(digest.digest(), Base64.NO_WRAP)
+    }
+
+    companion object {
+        private const val TAG = "CacheVerifier"
```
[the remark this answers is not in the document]
corycaywood: Oops, forgot to change this

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
corycaywood: You're right, it's kind of confusing that I'm using `removeOutdatedVersionApp` to remove the whole app... I've added a new function to remove the entire app folder: `removeApp`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/CachedMiniAppVerifier.kt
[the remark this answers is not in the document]
corycaywood: I've moved the `Exception` into here instead. But I don't think using `runBlocking` is the right solution here. We want the `verify` function to run synchronously but the `storeHash` function should run asynchronously. 

However, I realized that we don't need to create a new CoroutineScope here - instead I've changed `storeHash` to a `suspend` function and run the coroutine using `async`. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -48,6 +57,24 @@ internal class MiniAppDownloader(
         miniAppStatus.saveDownloadedMiniApp(miniAppInfo)
     }
 
+    private suspend fun retrieveDownloadedVersionPath(miniAppInfo: MiniAppInfo): String? {
```
khairul-alam-licon, line 60: Do we need to specify `suspend` here?

khairul-alam-licon (approved):
Looks good 👍 
Just a qus..

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/CachedMiniAppVerifier.kt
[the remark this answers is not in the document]
khairul-alam-licon: > We want the verify function to run synchronously but the storeHash function should run asynchronously.

I think, it sounds good 👍

minh-rakuten (approved):
MiniApp SDK so far so cool 💪

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
corycaywood: Nope, I added this on accident 😣

khairul-alam-licon (approved):
Thank you 👍

PR #157, opened 2020-09-24
feature: revoke custom permissions

khairul-alam-licon:
# Description
* Android SDK: Add `listDownloadedWithCustomPermissions` public interface 
* HostApp: Display Permissions button in Settings screen
* HostApp: Display List of Downloaded apps with custom permission results 
* HostApp: Allow Users to Revoke/Change the permissions in Permissions Settings screen

# Screenshots
<table><tr><td>
    
</td>
<td>
</td>
<td>
     
</td>
</tr>
</table>
<table><tr><td>
    
</td>
<td>
</td>
<td>
     
</td>
</tr>
</table>

## Links
[MINI-960](https://jira.rakuten-it.com/jira/browse/MINI-960)

# Checklist


khairul-alam-licon:
Seems need to increase the coverage.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -80,6 +81,14 @@ abstract class MiniApp internal constructor() {
         miniAppCustomPermission: MiniAppCustomPermission
     )
 
+    /**
+     * lists out the mini applications available with custom permissions in cache.
+     * @return [List<MiniAppInfo>] list of MiniApp what is downloaded and containing
+     * custom permissions data.
+     */
+    @Suppress("FunctionMaxLength")
+    abstract fun listDownloadedWithCustomPermissions(): List<MiniAppInfo>
```
corycaywood, line 90: See the return type for this on iOS: https://github.com/rakutentech/ios-miniapp/blob/master/MiniApp/Classes/MiniApp.swift#L76

This should return both the `MiniAppInfo` and the `MiniAppCustomPermission` for each Mini App. So the return type could be like this: `List<Pair<MiniAppInfo, MiniAppCustomPermission>>`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -30,6 +30,22 @@ internal class RealMiniApp(
...
         miniAppCustomPermissionCache.storePermissions(miniAppCustomPermission)
 
+    @Suppress("FunctionMaxLength")
+    override fun listDownloadedWithCustomPermissions(): List<MiniAppInfo> {
+        val result: ArrayList<MiniAppInfo> = arrayListOf()
+        val unfilteredMiniApps = miniAppDownloader.getDownloadedMiniAppList()
+        val filteredMiniApps =
+            miniAppCustomPermissionCache.readAllStoredPermissions(unfilteredMiniApps)
+
+        filteredMiniApps.forEach { (miniAppId) ->
```
corycaywood, line 40: This should return all Mini Apps which have been downloaded, whether they have any permissions granted or not. So no need to filter here. 

Mini Apps which are downloaded but don't have any permissions granted should just have all permissions set to DENIED.

corycaywood (commented):
It's looking good 👍  I think there was just some confusion on the meaning of `listDownloadedWithCustomPermissions`: the `withCustomPermisssions` part means that it returns the Mini App AND the Custom Permissions. Not that it returns only the Mini Apps which have custom permissions granted.

So maybe the method name should be `listDownloadedAndCustomPermissions` instead? I'm not sure...

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: Thanks, I'll update this part. Also, I'll keep the same name as `listDownloadedWithCustomPermissions` for the public interface.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
khairul-alam-licon: Thanks, I'll update this part. Also, I'll keep the same name as listDownloadedWithCustomPermissions for the public interface.

minh-rakuten (commented):
Nice 👍 
Still need to increase test coverage a little bit 😭.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStatus.kt
```diff
@@ -25,6 +25,18 @@ internal class MiniAppStatus(context: Context) {
             null
         }
 
+    @Suppress("EmptyCatchBlock", "SwallowedException")
+    fun getDownloadedMiniAppList(): List<MiniAppInfo> {
+        val list = arrayListOf<MiniAppInfo>()
+        prefs.all.map { entry ->
+            try {
+                list.add(gson.fromJson(entry.value.toString(), MiniAppInfo::class.java))
```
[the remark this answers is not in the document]
khairul-alam-licon: Yeah, I was thinking the same thing that we need to handle the caching for multi-versions per MiniApp. Now, a MiniApp ID is considering as a `key` while caching. But, we need to rethink the `key` with including version info as well.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -30,6 +30,18 @@ internal class RealMiniApp(
...
+    @Suppress("FunctionMaxLength")
+    override fun listDownloadedWithCustomPermissions(): List<Pair<MiniAppInfo, MiniAppCustomPermission>> {
+        val result: ArrayList<Pair<MiniAppInfo, MiniAppCustomPermission>> = arrayListOf()
+        val downloadedMiniApps = miniAppDownloader.getDownloadedMiniAppList()
+
+        downloadedMiniApps.forEach {
+            result.add(Pair(it, miniAppCustomPermissionCache.readPermissions(it.id)))
+        }
+
+        return result
```
corycaywood, line 42: You could use `map` and avoid needing to create a temporary list:

```suggestion
        return miniAppDownloader.getDownloadedMiniAppList().map {
            Pair(it, miniAppCustomPermissionCache.readPermissions(it.id))
        }
```

corycaywood (approved):
Looks good 👍 👍  Just a couple of suggestions.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: Thanks, I have updated it.

minh-rakuten (approved):
Tested and it's working perfectly.
Still need to do something with unit test.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppStatus.kt
```diff
@@ -25,6 +25,18 @@ internal class MiniAppStatus(context: Context) {
...
+            } catch (error: JsonSyntaxException) {}
```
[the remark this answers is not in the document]
khairul-alam-licon: I have added an empty list instead as it can be easier to be in unit test coverage.

khairul-alam-licon:
Thanks, I have increased few coverage.

khairul-alam-licon:
Thanks for the review.

PR #160, opened 2020-09-28
refactor: android 23 compatibility

minh-rakuten:
# Description
SDK:
Resolve phone scheme detection issue on Android 6 by replacing the current`shouldOverrideUrlLoading`. Have tested on other OS and found no problem as well.

Sample app:
Add UUID length check.

## Links
MINI-1614
MINI-1620

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
@@ -28,6 +28,7 @@ internal class MiniAppWebViewClient(
         return response
     }
 
+    // ToDo When the minimum support is upgraded to android 24, replace this function.
     override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
         var shouldCancelLoading = super.shouldOverrideUrlLoading(view, url)
         if (url != null) {
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
@@ -28,17 +28,16 @@ internal class MiniAppWebViewClient(
         return response
     }
 
-    override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
-        var shouldCancelLoading = super.shouldOverrideUrlLoading(view, request)
-        if (request.url != null) {
-            val requestUrl = request.url.toString()
-            if (requestUrl.startsWith("tel:")) {
-                openPhoneDialer(requestUrl)
+    override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
+        var shouldCancelLoading = super.shouldOverrideUrlLoading(view, url)
+        if (url != null) {
+            if (url.startsWith("tel:")) {
+                openPhoneDialer(url)
                 shouldCancelLoading = true
-            } else if (!miniAppScheme.isMiniAppUrl(requestUrl)) {
+            } else if (!miniAppScheme.isMiniAppUrl(url)) {
                 // check if there is navigator implementation on miniapp.
                 if (miniAppNavigator != null) {
-                    miniAppNavigator.openExternalUrl(requestUrl, externalResultHandler)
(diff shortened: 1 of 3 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
```diff
@@ -28,17 +28,16 @@ internal class MiniAppWebViewClient(
         return response
     }
 
-    override fun shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest): Boolean {
-        var shouldCancelLoading = super.shouldOverrideUrlLoading(view, request)
-        if (request.url != null) {
-            val requestUrl = request.url.toString()
-            if (requestUrl.startsWith("tel:")) {
-                openPhoneDialer(requestUrl)
+    override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
```
khairul-alam-licon, line 31: According to the [official doc of WebViewClient](https://developer.android.com/reference/android/webkit/WebViewClient), this method has been deprecated in API level `24`.  But, this is the working function below of `24`, I think, it's a bug itself inside Android SDK.. 🤔 

My idea is using latest `shouldOverrideUrlLoading(view: WebView, request: WebResourceRequest)` for API level `24+`, and using `shouldOverrideUrlLoading(view: WebView?, url: String?)` for below than `24`. 

So, we can refactor this part as following:

```
    @SuppressWarnings("deprecation")
    override fun shouldOverrideUrlLoading(view: WebView?, url: String?): Boolean {
        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N)
            return overrideUrlLoading(url.toString(), super.shouldOverrideUrlLoading(view, url))

        return false
    }

    @RequiresApi(Build.VERSION_CODES.N)
    override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest): Boolean {
        return overrideUrlLoading(request.url.toString(), super.shouldOverrideUrlLoading(view, request))
    }

    private fun overrideUrlLoading(requestUrl: String, shouldOverride: Boolean): Boolean {
        var shouldCancelLoading = shouldOverride
        if (requestUrl.startsWith("tel:")) {
            openPhoneDialer(requestUrl)
            shouldCancelLoading = true
        } else if (....) {
           ...
        }
        return shouldCancelLoading
    }
```

testapp/src/main/java/com/rakuten/tech/mobile/testapp/helper/UUIDHelper.kt
```diff
@@ -3,9 +3,10 @@ package com.rakuten.tech.mobile.testapp.helper
 import java.lang.IllegalArgumentException
 import java.util.UUID
 
+private const val UUID_LENGTH = 36
```
khairul-alam-licon, line 6: 👍

khairul-alam-licon (approved):
Thanks for adding the fix, I left one idea.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
minh-rakuten: It's a good idea but I think we can omit this version checking by using only one `shouldOverrideUrlLoading`. I tested and it's working fine on other android versions.
We may update our sdk to 24 as minimum later so let's keep this fix simple.

khairul-alam-licon: Yes, we need to use `override fun shouldOverrideUrlLoading(view: WebView?, request: WebResourceRequest)` when we can update the sdk to 24 as minimum level. For now, the current change is also accepted, so it's approved from my side 👍

corycaywood: @minh-rakuten Can you add a `TODO` here so we remember to update this in the future?

corycaywood (approved):
Looks okay 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebViewClient.kt
minh-rakuten: ToDo has been added.

PR #161, opened 2020-09-28
Fix/permissions public api

corycaywood:
# Description
This is to avoid making a breaking change to the SDK - if we add a new 'abstract' method, then the Host App's builds will fail unless they implement the new method. Instead, we can make the new method be 'open' and throw an error if a Mini App tries to use it when it's not implemented. In the future, we can add a default implementation of 'requestCustomPermissions' instead of throwing an error

# Checklist


khairul-alam-licon (approved):
👍 👍

minh-rakuten (approved):
👍 👍

PR #163, opened 2020-09-30
feat: get user name and profile photo

khairul-alam-licon:
# Description
- Add interface to the Javascript bridge for receiving User Name and Profile Photo where the SDK must first check if the required permissions have been granted. If the permissions have not been granted, then should return error response
- Update the Demo App to implement the new interface. Should return the User Name and Profile Photo value which is set on the settings screen

## Links
- MINI-1683
- MINI-1685

# Checklist


miniapp/USERGUIDE.md
```diff
@@ -163,6 +164,12 @@ val miniAppMessageBridge = object: MiniAppMessageBridge() {
         
         callback.invoke(true, null) // or callback.invoke(false, "error message")
     }
+
+    override fun requestUserName(callback: (isSuccess: Boolean, data: String?) -> Unit) {
+        // requesting user name implementation.
+        // .. .. ..
+        callback.invoke(true, null) // or callback.invoke(false, "error message")
```
corycaywood, line 171: Looks like the actual implementation of `requestUserName` has `data` as non-nullable. So I think this should be updated to use some example string like `"user_name"`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -62,8 +62,21 @@ abstract class MiniAppMessageBridge {
...
+            "The `MiniAppMessageBridge.requestCustomPermissions`" +
+                    " method has not been implemented by the Host App."
+        )
+    }
+
+    /**
+     * Post user name request from external.
+     * @param callback to return user name.
+     */
+    open fun requestUserName(callback: (isSuccess: Boolean, data: String) -> Unit) {
```
corycaywood, line 75: I think the function name should be `getUserName` - this matches the naming convention of `getUniqueId`.

Also, I'm not sure if we need to use a `callback` here. I was thinking this function would be the same as `getUniqueId` and just directly return a `String`. What do you think @khairul-alam-licon @minh-rakuten ?

corycaywood (commented):
Looking good 👍  Just left a few comments.

miniapp/USERGUIDE.md
corycaywood: It looks like iOS is taking the same approach of having a callback: https://github.com/rakutentech/ios-miniapp/pull/149/files#diff-eeb7f9f814d262c7e3820df82237603bR6

But on iOS, they have a `Result` with `success` and `failure` callbacks. Looks like Kotlin also has a `Result` class now, so maybe we should try that? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/#result

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
corycaywood: It looks like iOS is taking the same approach of having a callback: https://github.com/rakutentech/ios-miniapp/pull/149/files#diff-eeb7f9f814d262c7e3820df82237603bR6

But on iOS, they have a `Result` with `success` and `failure` callbacks. Looks like Kotlin also has a `Result` class now, so maybe we should try that? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/#result

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - **Feature:** Added public interface to read and store the custom permissions inside Android SDK. 
 - **Feature:** Share content from mini app. 
 - **Feature:** Added `listDownloadedWithCustomPermissions()` public interface that enables the host app to retrieve the list of downloaded mini-apps and their respective custom permissions.
+- **Feature:** Added `requestUserName()` interface in JavaScript bridge for receiving user name if the permission has been granted.
```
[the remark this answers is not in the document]
khairul-alam-licon: Yes, I'll adjust the related changes.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: Directly return a string is fine to me. We should simplify the usage for the developers in host app.

minh-rakuten, line 75: I believe we should break down the `MiniAppMessageBridge` here.

Can we delegate `requestUserName` & `requestUserPhoto` into another component class such as `UserInfoBridge`? The `UserInfoBridge` belongs to `MiniAppMessageBridge`.
I will divide other existed functions in the future too because `MiniAppMessageBridge` is too big now. 

@khairul-alam-licon @corycaywood Please feel free to suggest other design patterns to divide `MiniAppMessageBridge` smaller.

minh-rakuten (commented):
I think you can include the task for request user photo in this PR too. The same for iOS PR https://github.com/rakutentech/ios-miniapp/pull/149

We should have `requestUserName` and `requestUserPhoto` being under the same wrapper class such as `UserInfoBridge`, not in `MiniAppMessageBridge` directly. Please see my suggestion below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
corycaywood: I was thinking about the same thing, but I haven't thought of a good way of doing this. The class is also getting quite long for the Host App to implement since they need to override so many functions.

Maybe we could add a function for `MiniAppMessageBridge.setUserInfoProvider`? (similar to `MiniAppMessageBridge.setAdMobDisplayer`)

khairul-alam-licon: Thanks for the suggestion. I liked the concept of using [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/#result). 
Firstly, I wanted to do `fun getUserName(): Result<String>` but `Result` can't be used as a return type unfortunately. 

So, I went with `fun requestUserName(callback: (result: Result<String>) -> Unit)`. But, at the end I think, we don't need to handle those `success` or `failure` in HostApp, rather we can handle everything inside SDK, let users to use this feature easily.

So, I am preferring finally `abstract fun getUserName(): String`. I think, @minh-rakuten already has a `+1` on it.
@corycaywood What do you think?

khairul-alam-licon: Well I think, the purpose of defining `Bridge` postfix means a component to communicate with external MiniApp customers. We may have `UserInfoBridge`, but the name may confuses the users with multiple bridges and the convention will be difficult to manage with lots of features. 
Or, we could write `MiniAppMessageBridgeDelegate` which will be extended by `MiniAppMessageBridge` but it will solve the "longer component" issue temporarily. 

In the other side, logically `UserInfoProvider` is only for providing the data, not for other operations. We can have this class optionally, but it will not solve the actual problem. 

In my opinion, we can have something like-

```
internal class UserInfoHandler(private val bridge: MiniAppMessageBridge) {
  
   fun onRequestUserName(callbackObj: CallbackObj) {

         // get data from UserInfoProvider 
         // implement the usecases of postValue / postError

   }
}
```
And, we can do the same for all the other features to hold the single responsibility. And, in that's way, we can have a single messaging `Bridge` for the users.

Let me know, what is in your mind.

khairul-alam-licon: Well I think, the purpose of defining `Bridge` postfix means a component to communicate with external MiniApp customers. We may have `UserInfoBridge`, but the name may confuses the users with multiple bridges and the convention will be difficult to manage with lots of features. 
Or, we could write `MiniAppMessageBridgeDelegate` which will be extended by `MiniAppMessageBridge` but it will solve the "longer component" issue temporarily. 

In the other side, logically `UserInfoProvider` is only for providing the data, not for other operations. In my opinion, we can have something like-

```
internal class UserInfoHandler(private val bridge: MiniAppMessageBridge) {
  
   fun onRequestUserName(callbackObj: CallbackObj) {

         // get data from UserInfoProvider 
         // implement the usecases of postValue / postError

   }
}
```

`UserInfoHandler` is an example, we can have alternative postfix like `UserInfoHelper`, `UserInfoRequester` etc.

And, we can do the same for all the other features to hold the single responsibility. 
Let me know, what is in your mind.

minh-rakuten: @khairul-alam-licon I think it's good to go.

khairul-alam-licon: I have updated the PR with `UserInfoHandler` idea. if it's sound good, I'll forward to get profile-photo. In this PR, it should be fine.

corycaywood: @khairul-alam-licon Yes, I think `fun getUserName(): String` is okay - it's best to be consistent among the functions that have similar functionality (`getUniqueId`).

@ts-leojoseph-r I think you implemented a callback for this on iOS. Can it be updated to directly return a string instead?

ts-leojoseph-r: Sure @corycaywood  👍 . Initial plan I had was to return some more custom errors from the Host app, I guess just returning the string would suffice now for both Username and Profile Photo.

minh-rakuten (dismissed):
Nice 👍 
We could include the request photo otherwise another PR is also fine.

khairul-alam-licon:
I have added profile photo part also.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/ProfileSettingsActivity.kt
```diff
@@ -75,9 +83,20 @@ class ProfileSettingsActivity : BaseActivity() {
...
+            profileUrlBase64 = encodeImageForMiniApp(profileUrl)
         }
     }
 
+    private fun encodeImageForMiniApp(profileUrl: String): String {
+        val uri = Uri.parse(profileUrl)
+        val imageStream = contentResolver.openInputStream(uri)
+        val bitmap = BitmapFactory.decodeStream(imageStream)
+        val byteArrayOutputStream = ByteArrayOutputStream()
+        bitmap.compress(Bitmap.CompressFormat.PNG, 100, byteArrayOutputStream)
```
corycaywood, line 95: Should we use `JPEG` here instead? Using PNG will give us a much larger file size. I'm worried this will affect the performance when Apps are retrieving the image, so it will give mini app developers an inaccurate view of this feature (most host apps will be using an HTTP URL here instead of base64 data).

Also, maybe we can set the quality to a lower value? Probably even `50` is okay. What do you think?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/ProfileSettingsActivity.kt
```diff
@@ -75,9 +83,20 @@ class ProfileSettingsActivity : BaseActivity() {
             val imageUri = data?.data
             setProfileImage(imageUri)
             profileUrl = imageUri.toString()
+            profileUrlBase64 = encodeImageForMiniApp(profileUrl)
         }
     }
 
+    private fun encodeImageForMiniApp(profileUrl: String): String {
```
minh-rakuten, line 90: This function should run on worker thread. For example, you can use `withContext(Dispatchers.IO){}`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -42,6 +46,7 @@ abstract class MiniAppMessageBridge {
         this.webViewListener = webViewListener
         this.customPermissionCache = customPermissionCache
         this.miniAppInfo = miniAppInfo
+        this.userInfoHandler = UserInfoHandler(this)
```
corycaywood, line 49: This is nice! 

I'd like to see a way of separating this for the Host App developer too so they don't have to define so many methods in `MiniAppMessageBridge`. That's especially important for the user data feature because we could end up having many methods related to this. But I haven't thought of a good way of doing this 🤔

I was thinking we could make `UserInfoHandler` be public, and then add a function for `MiniAppMessageBridge.setUserInfoHandler`. Something like this:
```
class UserInfoHandler {
  open getUserName(): String {
    throw MiniAppSdkException()
  }

  internal getUserName() {
    val userName = getUserName()
    // ....
  }
}

class MiniAppMessageBridge {
  fun setUserInfoHandler(handler: UserInfoHandler) { 
    this.userInfoHandler = handler
    this.userInfoHandler.bridge = this
  } 
}
```

What do you think? Do you have any other ideas how we can separate this implementation for the Host App?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/userinfo/UserInfoHandler.kt
```diff
@@ -0,0 +1,76 @@
+package com.rakuten.tech.mobile.miniapp.userinfo
+
+import com.rakuten.tech.mobile.miniapp.js.CallbackObj
+import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageBridge
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionResult
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
+
+@Suppress("TooGenericExceptionCaught", "LongMethod")
+internal class UserInfoHandler(private val bridge: MiniAppMessageBridge) {
```
minh-rakuten, line 9: There is one thing I missed. We should pass only the listener of `MiniAppMessageBridge` to execute `postValue`, `postError` and `MiniAppCustomPermissionCache`.

```suggestion
internal class UserInfoHandler(
private val bridgeListener: MAMessageBridgeListener,
private val customPermissionCache: MiniAppCustomPermissionCache) {
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/userinfo/UserInfoHandler.kt
```diff
@@ -0,0 +1,76 @@
+package com.rakuten.tech.mobile.miniapp.userinfo
```
corycaywood, line 1: Shouldn't this class be under `js` instead since it's closely coupled to `MiniAppMessageBridge`? Maybe it can go under `js/userinfo` if you think there will be more classes related to user in the future.

corycaywood (commented):
It's looking good. I've left a few more comments.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: I think `UserInfoHandler` cannot go to be public because it depends on internal component `MiniAppCustomPermissionCache`. However, the current implementation `UserInfoHandler` should not hold `MiniAppMessageBridge` because it caused cross reference https://github.com/rakutentech/android-miniapp/pull/163#discussion_r500675471

corycaywood: Yeah, we had better not have a cyclical dependency.

But couldn't we still do something like this?

```
class MiniAppMessageBridge {
  fun setUserInfoHandler(handler: UserInfoHandler) { 
    this.userInfoHandler = handler
    this.userInfoHandler.init(bridgeListener, permissionCache)
  } 
}
```

minh-rakuten: yeah it's nice.

khairul-alam-licon: I agree with the suggestions as `MiniAppMessageBridge` will act just like a bridge, and all the feature will be in their independent pattern. I'll update according to the ideas.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/ProfileSettingsActivity.kt
khairul-alam-licon: I think it's reasonable, I have updated with `JPEG` with `50` quality.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/userinfo/UserInfoHandler.kt
khairul-alam-licon: I have updated this part. Now, we need to make sure that miniapp has been created before initializing `UserInfoHandler` as `MiniAppCustomPermissionCache` is an internal class.

corycaywood (approved):
Looks good! 👍 👍

miniapp/USERGUIDE.md
```diff
@@ -163,7 +163,26 @@ val miniAppMessageBridge = object: MiniAppMessageBridge() {
         
         callback.invoke(true, null) // or callback.invoke(false, "error message")
     }
+
+// initialize UserInfoHandler after ensuring that MiniApp has been created.
```
corycaywood, line 167: What happens if they don't create the Mini App first? It seems inconvenient because they need to create `MiniAppMessageBridge` before calling `MiniApp.create`, but then they're supposed to set the `UserInfoHandler` after calling `MiniApp.create`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -102,6 +105,8 @@ abstract class MiniAppMessageBridge {
             ActionType.SHARE_INFO.action -> onShareContent(callbackObj.id, jsonStr)
             ActionType.LOAD_AD.action -> onLoadAd(callbackObj.id, jsonStr)
             ActionType.SHOW_AD.action -> onShowAd(callbackObj.id, jsonStr)
+            ActionType.GET_USER_NAME.action -> userInfoHandler.onGetUserName(callbackObj.id)
```
corycaywood, line 108: Just one thing here and below: it's possible the Host App didn't call `setUserInfoHandler`, so we need to handle that case. Maybe either a default implementation of `UserInfoHandler` that just returns an error or just directly return an error here?

corycaywood: Ah, I see you're already throwing errors from the default methods, so I guess you could just remove `lateinit` from `userInfoHandler` variable and directly instantiate the default `UserInfoHandler()`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -111,6 +116,12 @@ abstract class MiniAppMessageBridge {
         this.isAdMobEnabled = isAdMobProvided()
     }
 
+    /** Set implemented userInfoHandler. Can use the default provided class from sdk [UserInfoHandler]. **/
+    fun setUserInfoHandler(handler: UserInfoHandler) {
+        this.userInfoHandler = handler
+        this.userInfoHandler.init(this, customPermissionCache, miniAppInfo.id)
```
minh-rakuten, line 122: Unfortunately, this does not work as we expected.

The `customPermissionCache` and `miniAppInfo` were initialized only when webview is created. There is a case that hostapp developer can set `userInfoHandler` before webview is created, `MiniAppMessageBridge` object is separated from miniapp webview.

@corycaywood I think we should go back with the solution init `userInfoHandler` in `MiniAppMessageBridge` constructor.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoHandler.kt
```diff
@@ -0,0 +1,104 @@
...
+import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageBridgeListener
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionResult
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
+
+/**
+ * A class to provide the interfaces for getting user info e.g. user-name, profile-photo etc.
+ */
+@Suppress("TooGenericExceptionCaught", "LongMethod")
+open class UserInfoHandler {
```
corycaywood, line 13: Should we prevent the Host App from directly instantiating this class? We could either make it `abstract class` or add an `internal constructor`?

miniapp/USERGUIDE.md
khairul-alam-licon: It will show error as `MiniAppCustomPermissionCache` hasn't been initialized internally. I had initialize `UserInfoHandler` inside `MiniAppDisplayViewModel.obtainMiniAppDisplay(~)` which is not user friendly.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: It's okay to pass the only the `MiniAppMessageBridge` instead of passing those parameters separately. That's how, we don't need to worry about when miniapp will be created. 

```
fun setUserInfoHandler(handler: UserInfoHandler) {
        this.userInfoHandler = handler
        this.userInfoHandler.init(miniAppBridge)
    }
```

minh-rakuten: but it will cause cyclical dependency because `MiniAppMessageBridge` contains `UserInfoHandler`.

Maybe there is a solution to keep using `setUserInfoHandler` and still allow developers to set it before webview is created.

minh-rakuten:
For me there are only two problems left:
`setUserInfoHandler` when miniapp is not created.
`postValue` and `postError` are public -> it should be keep as internal.

I have created [the PR](https://github.com/khairul-alam-licon/android-miniapp/pull/4) into this. @khairul-alam-licon @corycaywood Please review it.

corycaywood:
> For me there are only two problems left:
> `setUserInfoHandler` when miniapp is not created.
> `postValue` and `postError` are public -> it should be keep as internal.
> 
> I have created [the PR](https://github.com/khairul-alam-licon/android-miniapp/pull/4) into this. @khairul-alam-licon @corycaywood Please review it.

@minh-rakuten Nice, I think this is a good solution 👍

khairul-alam-licon:
Thank you @minh-rakuten for [the PR](https://github.com/khairul-alam-licon/android-miniapp/pull/4) as a solution of your mentioned problems. After reading your PR, I realized that we need to call `UserInfoHandler.init()` inside `MiniAppMessageBridge.init()` to handle initialization of those internal parameters.  Besides, we can declare `MiniAppMessageBridgeListener` as internal. 

I have updated the PR after taking those ideas, please review.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -111,6 +119,11 @@ abstract class MiniAppMessageBridge {
         this.isAdMobEnabled = isAdMobProvided()
     }
 
+    /** Set implemented userInfoHandler. Can use the default provided class from sdk [UserInfoHandler]. **/
+    fun setUserInfoHandler(handler: UserInfoHandler) {
+        this.userInfoHandler = handler
+    }
```
minh-rakuten, line 125: There is a lack of initialization for `setUserInfoHandler` here. The developer may call `setUserInfoHandler` after miniapp webview is created.
Here is [suggestion](https://github.com/khairul-alam-licon/android-miniapp/blob/bb1083fc3263c6878592d64adb69f27b2a140665/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt#L129) from my PR.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -264,13 +278,13 @@ abstract class MiniAppMessageBridge {
             postError(callbackId, "${ErrorBridgeMessage.ERR_SHOW_AD} ${ErrorBridgeMessage.ERR_NO_SUPPORT_HOSTAPP}")
     }
 
-    /** Return a value to mini app. **/
-    internal fun postValue(callbackId: String, value: String) {
+    /** Emit a value to mini app. **/
+    override fun postValue(callbackId: String, value: String) {
```
minh-rakuten, line 282: `postValue` and `postError` are still public. The `MiniAppMessageBridge` is a public class that's why it's still exposed even we declare the listener interface as internal. 
Here is [suggestion](https://github.com/khairul-alam-licon/android-miniapp/blob/bb1083fc3263c6878592d64adb69f27b2a140665/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/BridgeExecutor.kt) from my PR.

minh-rakuten (commented):
The issues are still there. It can be resolved by my PR suggestion but now there is conflicted files.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Ah, I didn't noticed it. Updated according to your suggestion.

minh-rakuten (approved):
👍 👍

corycaywood (approved):
Looks good 👍

PR #174, opened 2020-10-19
fix: provide keyboard visibility function to MiniApp

khairul-alam-licon:
# Description
I was doing few investigations in Android to find a way to solve MINI-1706.  The issue is: when there is any bottom button inside JS MiniApp which is hidden because of keyboard is shown. 

By this PR, I came up with a solution is that, we can provide a functionality from MiniApp SDK called `getKeyboardVisibility()` which will return `"visible" / "invisible"`. Although, it's difficult to find when Android soft keyboard is shown or hidden since there is no way to detect if keyboard is opened or not, please have a look at the existing solution by this PR.

TODO: Unit test coverage once the solution looks good.

## Links
- MINI-1706

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/KeyboardBridgeDispatcher.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/KeyboardBridgeDispatcher.kt
@@ -2,6 +2,9 @@ package com.rakuten.tech.mobile.miniapp.js
 
 import android.app.Activity
 import android.content.Context
+import android.os.Bundle
+import android.os.ResultReceiver
+import android.view.ViewGroup
 import android.view.inputmethod.InputMethodManager
 
 internal class KeyboardBridgeDispatcher(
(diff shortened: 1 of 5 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/KeyboardBridgeDispatcher.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val bridgeExecutor: MiniAppBridgeExecutor,
+    private val activity: Activity
+) {
+
+    fun onGetKeyboardVisibility(callbackId: String) {
+        try {
+            val inputMethodManager: InputMethodManager =
+                activity.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager
+
+            if (inputMethodManager.isActive && inputMethodManager.isAcceptingText)
```
khairul-alam-licon, line 17: There is a case, when an input field inside JS MiniApp has a focus but keyboard is hidden, then `inputMethodManager.isAcceptingText` will return `true` means Keyboard is shown. So, this solution doesn't solve this use-case.

khairul-alam-licon: Solved by https://github.com/rakutentech/android-miniapp/pull/174/commits/15df9f7f046621ecc91ccb40f8a2c772aa702946

khairul-alam-licon:
Closing this PR because currently there is no reliable way found yet.

PR #177, opened 2020-10-21
fix: clean fullscreen view

minh-rakuten:
# Description
Remove fullscreen view when the hostapp has to clear miniapp view manually.
Update document about clearing miniapp.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,9 @@
 ## CHANGELOG
 
-### x.x.x
+### 2.4.0
 **SDK**
 - **Feature:** Handle the screen orientation change request from miniapp.
+- **Fix:** Clean up fullscreen view when exit miniapp. `MiniAppDisplay.destroyView` is mandatory for single activity architect.
 
 ### 2.3.0 (2020-10-15)
 **SDK**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### 2.4.0
 **SDK**
 - **Feature:** Handle the screen orientation change request from miniapp.
-- **Fix:** Clean up fullscreen view when exit miniapp. `MiniAppDisplay.destroyView` is mandatory for single activity architect.
+- **Fix:** Clean up fullscreen view when exit miniapp. `MiniAppDisplay.destroyView` is mandatory for single activity architecture.
 
 ### 2.3.0 (2020-10-15)
 **SDK**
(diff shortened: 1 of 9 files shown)


miniapp/USERGUIDE.md
```diff
@@ -216,11 +216,14 @@ class MiniAppActivity : Activity(), CoroutineScope {
 
 `MiniAppDisplay.navigateBackward` and `MiniAppDisplay.navigateForward` facilitates the navigation inside a mini app if the history stack is available in it. A common usage pattern could be to link it up to the Android Back Key navigation.
 
+**Note:** Clearing up the mini app display is essential. `MiniAppDisplay.destroyView` is required to be called when exit miniapp.
```
corycaywood, line 219: Maybe we should also update the code sample above to include this? For example:
```
override fun onDestroy() {
    super.onDestroy()
    miniAppDisplay.onDestroy()
}
```

corycaywood (approved):
Looks good 👍  See my suggestion below. Also, could you add this to the changelog?

miniapp/USERGUIDE.md
minh-rakuten: I don't know which example is good here. The `onDestroy()` example above is correct for separated miniapp activity case but for the fragment, the example could be something like this 🤔

```
// hostapp example
fun onMiniappRemove() {
   miniAppDisplay.destroyView() // miniapp sdk call
   closeScreen()
}
```

CHANGELOG.md
```diff
@@ -1,8 +1,9 @@
 ## CHANGELOG
 
-### x.x.x
+### 2.4.0
 **SDK**
 - **Feature:** Handle the screen orientation change request from miniapp.
+- **Fix:** Clean up fullscreen view when exit miniapp. `MiniAppDisplay.destroyView` is mandatory for single activity architect.
```
khairul-alam-licon, line 6: - Can we say `MiniAppDisplay.destroyView()`?
- architect -> architecture?

miniapp/USERGUIDE.md
khairul-alam-licon: I think, the existing line is fine becuase including code example may arise confusion in other clients who have different architecture. Though, we could write `miniAppDisplay.destroyView()` instead of `miniAppDisplay.destroyView`.

khairul-alam-licon (approved):
👍

CHANGELOG.md
minh-rakuten: My bad, architecture is correct.

miniapp/USERGUIDE.md
minh-rakuten: Yeah I agree on just only description line here.

PR #178, opened 2020-10-21
fix: use binding as a reference for the views to import

khairul-alam-licon:
# Description
Recently, we have encountered one issue in sample app while importing the views which throws sometimes:
```
java.lang.IllegalStateException: swipeRefreshLayout must not be null
        at com.rakuten.tech.mobile.testapp.ui.miniapplist.MiniAppListFragment$onStart$1.run(MiniAppListFragment.kt:71)
        .....
```

This is happened because synthetic import of `swipeRefreshLayout` sometimes executes too early which can produce `NPE`.
We can solve this issue either using `swipeRefreshLayout?.` but it'll not execute the action we want when it's `null` at `onStart()`.

Another way is using a global reference i.e. `binding` for getting `swipeRefreshLayout`. I added the same solution in other views as well.

# Checklist


minh-rakuten (approved):
Looks good. Thanks for refactoring 👍

corycaywood (approved):
Looks good 👍

khairul-alam-licon:
Thanks for the review.

PR #180, opened 2020-10-24
fix: remove orientation request when fullscreen

minh-rakuten:
# Description
- Fix rotation in fullscreen
Our sdk already has the forced screen orientation feature. That feature applies to every case including fullscreen so I have removed redundant codes here. This solves the problem of fullscreen miniapp in hostapp with no orientation change handling.

- Fix HTML video fullscreen
I notice that when video goes fullscreen, our toolbar is visible (but cannot touch) so the flag `View.SYSTEM_UI_FLAG_LAYOUT_STABLE` is required. Then there is white space in position of status bar and navigation so I have to set the background of fullscreen view in black.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
@@ -73,16 +73,19 @@ internal class MiniAppWebView(
         loadUrl(getLoadUrl())
     }
 
-    @Suppress("TooGenericExceptionCaught", "SwallowedException")
-    fun destroyView() = try {
+    override fun onDetachedFromWindow() {
+        super.onDetachedFromWindow()
+
+        if (!(context as Activity).isDestroyed) {
+            miniAppWebChromeClient.onWebViewDetach()
+            miniAppMessageBridge.onWebViewDetach()
+        }
+    }
+
+    fun destroyView() {
         stopLoading()
         webViewClient = null
-        miniAppWebChromeClient.destroy()
-        webChromeClient = null
-        miniAppMessageBridge.onWebViewDestroy()
         destroy()
-    } catch (e: Exception) {
-        // The activity may release before those executions.
     }
 
     override fun runSuccessCallback(callbackId: String, value: String) {
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
@@ -73,9 +73,15 @@ internal class MiniAppWebView(
         loadUrl(getLoadUrl())
     }
 
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+        onResume()
+    }
+
     override fun onDetachedFromWindow() {
         super.onDetachedFromWindow()
 
+        onPause()
         if (!(context as Activity).isDestroyed) {
             miniAppWebChromeClient.onWebViewDetach()
             miniAppMessageBridge.onWebViewDetach()
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,3 +1,10 @@
 ignore:
   - "miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt"
   - "miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt"
+coverage:
+  status:
+    patch: off
+    project:
+      default:
+        target: 80%
+        threshold: 4%
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
@@ -149,7 +149,7 @@ internal class MiniAppWebChromeClient(
     }
     // end region video fullscreen
 
-    fun destroy() {
+    fun onWebViewDetach() {
         if (customView != null)
             onHideCustomView()
     }
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
@@ -2,7 +2,6 @@ package com.rakuten.tech.mobile.miniapp.display
 
 import android.app.Activity
 import android.content.Context
-import android.content.pm.ActivityInfo
 import android.view.View
 import android.view.ViewGroup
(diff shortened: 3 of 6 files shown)


corycaywood (commented):
This is looking better, so I think this is the right approach.

But there is a problem now with full screen video in single screen activity Apps. It seems that the video view automatically tries to set landscape mode, but this causes a reload of the mini app view if the App doesn't handle configuration changes. So maybe we should add something about this in the user guide - maybe under the troubleshooting section?

Also, I was thinking maybe we should expose something like `onShowFullScreen` and `onHideFullScreen` to the Host App so they can override the default functionality. What do you think? We could do this in another PR.

corycaywood:
> This is looking better, so I think this is the right approach.
> 
> But there is a problem now with full screen video in single screen activity Apps. It seems that the video view automatically tries to set landscape mode, but this causes a reload of the mini app view if the App doesn't handle configuration changes. So maybe we should add something about this in the user guide - maybe under the troubleshooting section?
> 
> Also, I was thinking maybe we should expose something like `onShowFullScreen` and `onHideFullScreen` to the Host App so they can override the default functionality. What do you think? We could do this in another PR.

@minh-rakuten Sorry, I was wrong about the full screen video issue. I was testing in a mini app which was using `requestOrientation`, so that's why it was automatically getting set to full screen. I think you've added this information to the user guide in the PR for `requestOrientation`, so this should be okay.

corycaywood (approved):
👍

khairul-alam-licon (approved):
👍

corycaywood:
@minh-rakuten @khairul-alam-licon I was thinking about something else related to this: should we also override `onDetachedFromWindow` in `MiniAppWebView`? This could allow the full-screen view to be cleaned up when the Host App removes the view but doesn't call `MiniAppDisplay.destroyView`. For example they might want to temporarily remove the view and then attach it again. I was thinking something like this:

```
internal class MiniAppWebView: WebView() {
    override fun onDetachedFromWindow() {
        super.onDetachedFromWindow()

        this.onPause() // Just call the default WebView pause - not sure if this is necessary but I think it helps ensure that video and audio gets paused

        miniAppMessageBridge.onWebViewPause() // Does the same thing that `MiniAppMessageBridge.onWebViewDestroy` currently does
        miniAppWebChromeClient.pause() // Does the same thing that `MiniAppWebChromeClient.destroy` currently does
    }

    override fun onAttachedToWindow() {
        super.onAttachedToWindow()
        this.onResume()
    }
}
```

What do you think?

minh-rakuten:
> @minh-rakuten @khairul-alam-licon I was thinking about something else related to this: should we also override `onDetachedFromWindow` in `MiniAppWebView`? This could allow the full-screen view to be cleaned up when the Host App removes the view but doesn't call `MiniAppDisplay.destroyView`. For example they might want to temporarily remove the view and then attach it again. I was thinking something like this:
> 
> ```
> internal class MiniAppWebView: WebView() {
>     override fun onDetachedFromWindow() {
>         super.onDetachedFromWindow()
> 
>         this.onPause() // Just call the default WebView pause - not sure if this is necessary but I think it helps ensure that video and audio gets paused
> 
>         miniAppMessageBridge.onWebViewPause() // Does the same thing that `MiniAppMessageBridge.onWebViewDestroy` currently does
>         miniAppWebChromeClient.pause() // Does the same thing that `MiniAppWebChromeClient.destroy` currently does
>     }
> 
>     override fun onAttachedToWindow() {
>         super.onAttachedToWindow()
>         this.onResume()
>     }
> }
> ```
> 
> What do you think?

@corycaywood I tested when in fullscreen and close miniapp, had this error ` java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.dispatchDetachedFromWindow()' on a null object reference`.

There is another risky thing here is that when `onDestroy` of activity called, `MiniAppWebView.destroyView()` is called before `onDetachedFromWindow()` -> strange to me.

corycaywood:
> 
> @corycaywood I tested in our Demo app, when in fullscreen and close miniapp, had this error ` java.lang.NullPointerException: Attempt to invoke virtual method 'void android.view.View.dispatchDetachedFromWindow()' on a null object reference`.
> 
> There is another risky thing here is that when `onDestroy` of activity called, `MiniAppWebView.destroyView()` is called before `onDetachedFromWindow()` -> strange to me.

Yeah, that's strange. I would think that detach gets called before destroy.

Could you save an `isDestroyed` variable to check before performing the actions in `onDetachedFromWindow`? It sounds like it's specifically the call to `super.onDetachedFromWindow()` which causes problems.

minh-rakuten:
> Yeah, that's strange. I would think that detach gets called before destroy.
> 
> Could you save an `isDestroyed` variable to check before performing the actions in `onDetachedFromWindow`? It sounds like it's specifically the call to `super.onDetachedFromWindow()` which causes problems.

The error happens if we execute hide fullscreen code but when I run into debug, the error is not thrown from hide fullscreen codes.

minh-rakuten:
We are able to solve the error onDetach when the activity is destroyed now. This is the refactor improvement so please do a code review again @corycaywood @khairul-alam-licon.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -73,16 +73,25 @@ internal class MiniAppWebView(
...
+    override fun onAttachedToWindow() {
+        super.onAttachedToWindow()
+        onResume()
+    }
+
+    override fun onDetachedFromWindow() {
+        super.onDetachedFromWindow()
+
+        onPause()
+        if (!(context as Activity).isDestroyed) {
```
corycaywood, line 85: Is there the possibility that this cast could fail? Maybe this should be `if (context is Activity && !(context as Activity).isDestroyed)`?

corycaywood (approved):
Nice, looks good! 👍  I just have one concern below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
khairul-alam-licon: I think, we can skip it as we are already have a check called `isContextValid` when providing this web view from `RealMiniAppDisplay`.

khairul-alam-licon (approved):
Ok to go 👍 Just one qus to ask.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -73,16 +73,25 @@ internal class MiniAppWebView(
...
+            miniAppWebChromeClient.onWebViewDetach()
+            miniAppMessageBridge.onWebViewDetach()
```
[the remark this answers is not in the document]
minh-rakuten: I think it's not necessary because when MiniAppWebView init -> miniAppMessageBridge init -> screenBridgeDispatcher init

PR #181, opened 2020-10-24
refactor: optional screen orientation

minh-rakuten:
# Description
- There are the possibilities that hostapp does not want miniapp to change screen orientation -> there is no orientation handler from hostapp.
To solve the ambiguous awareness of miniapp lock screen, the default setting of locking screen is false. Hostapp will enable the allowance by itself when they have handled the screen orientation change.

- Update the document guideline for hostapp.

- Define codecov rule.

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/ScreenBridgeDispatcher.kt
```diff
@@ -8,37 +8,42 @@ import com.rakuten.tech.mobile.miniapp.js.ErrorBridgeMessage.ERR_SCREEN_ACTION
...
-            bridgeExecutor.postValue(callbackObj.id, SUCCESS)
-        } catch (e: Exception) {
-            bridgeExecutor.postError(callbackObj.id, "$ERR_SCREEN_ACTION ${e.message}")
         }
     }
 
     fun releaseLock() {
-        if (isLocked) {
+        if (allowScreenOrientation && isLocked) {
             activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER
```
corycaywood, line 47: If we use `ActivityInfo.SCREEN_ORIENTATION_USER` here, then doesn't this mean it will override the default for the Host App? I mean if the Host App has set `PORTRAIT`, then this would reset it to `USER` instead. Also, if it's a single activity App, I think this would reset the orientation for the whole App.

Could we instead save the original orientation, and then reset to the original value when `releaseLock()` is called? (Similar to what we were doing in the full-screen view)

corycaywood (commented):
This is looking good 👍  This will prevent unexpected behavior for the Host App.

See my comment below: I'm not sure if my comment is correct or not. We can investigate this.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/ScreenBridgeDispatcher.kt
minh-rakuten: On fullscreen view code before, I assigned the fixed value for original orientation. Basically, we can only get the portrait or landscape state, we cannot get the screen config state.

There is a situation that Hostapp supports both portrait & landscape but when we reset it to portrait (in case the previous is portrait) -> on android it means we force the portrait lock.
We cannot detect to see whether hostapp config only single or both screen orientations.

miniapp/USERGUIDE.md
```diff
@@ -381,6 +381,17 @@ class CustomAdDisplayer: MiniAppAdDisplayer {
...
+### #6 Screen Orientation
+The default setting does not allow miniapp to change hostapp screen orientation.
+Hostapp can allow miniapp to control the screen orientation for better experience by calling 
+```kotlin
+miniAppMessageBridge.allowScreenOrientation(true)
+```
+
+In case miniapp is allowed to control, please ensure that your activity handles screen orientation.
+There are several ways to prevent the view from being reset.
+In our Demo App, we set the config on activity. See [here](https://developer.android.com/guide/topics/resources/runtime-changes#HandlingTheChange).
```
corycaywood, line 393: Can we also give a basic example? For example that they can add `android:configChanges="orientation|screenSize"` to their manifest?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/ScreenBridgeDispatcher.kt
corycaywood: Could you do something like this?

```
internal class ScreenBridgeDispatcher(
    private val activity: Activity
) {
  private val hostAppOrientation = activity.requestedOrientation

  // ...

  fun releaseLock() {
    activity.requestedOrientation = hostAppOrientation
    // ...
  }
}
```

minh-rakuten: ah yeah it's the right approach 🙇

corycaywood (approved):
Looks good! 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -29,9 +29,11 @@ abstract class MiniAppMessageBridge {
     private lateinit var miniAppInfo: MiniAppInfo
     private lateinit var activity: Activity
     private lateinit var userInfoBridgeDispatcher: UserInfoBridgeDispatcher
-    internal lateinit var screenBridgeDispatcher: ScreenBridgeDispatcher
     private val adBridgeDispatcher = AdBridgeDispatcher()
 
+    private lateinit var screenBridgeDispatcher: ScreenBridgeDispatcher
+    private var allowScreenOrientation = false
```
khairul-alam-licon, line 35: I think, if we directly assign `false` inside `ScreenBridgeDispatcher`'s constructor, then we can remove this line:

e.g.

```
internal class ScreenBridgeDispatcher(
    ....
    var allowScreenOrientation: Boolean = false
) 

abstract class MiniAppMessageBridge {
   ...
    private lateinit var screenBridgeDispatcher: ScreenBridgeDispatcher

    internal fun init(
        ...
    ) {
        ...
        this.screenBridgeDispatcher = ScreenBridgeDispatcher(activity, bridgeExecutor)
        ...
   }

  fun allowScreenOrientation(isAllowed: Boolean) {
        if (this::screenBridgeDispatcher.isInitialized)
            screenBridgeDispatcher.allowScreenOrientation = isAllowed
   }
}
```

khairul-alam-licon (approved):
Good to go 👍 
I have left one comment if it's reasonable to address.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: yeah, let me update it.

minh-rakuten: ah I check it again and see that we cannot go that way. The `allowScreenOrientation(isAllowed: Boolean)` will not change the state of `ScreenBridgeDispatcher` after it is initialized.

PR #182, opened 2020-10-26
feat: permission default UI in SDK

khairul-alam-licon:
# Description
- Add default UI inside SDK for managing custom permissions which is looked like iOS. Have "Close" and "Save" button for user-experience
- `MiniAppMessageBridge#requestCustomPermissions` is already optional function, so this PR doesn't need to include this change.
- If the Host App DOES NOT implement `requestCustomPermissions`, then the SDK should automatically use the default, built-in Permissions UI 
- Move few codes about custom permissions from `MiniAppMessageBridge` to `CustomPermissionBridgeDispatcher`
- Fix: Execute `requestCustomPermissions` or show default UI after verifying if there is any denied permissions, otherwise just send the existing permissions json to miniapp.

## Screenshots
Screen Shot 2020-10-29 at 12 40 15

## Links
- MINI-1678

# Checklist


corycaywood (commented):
The new UI is looking really good 👍 

I just noticed something about the public interface which I think might affect what you're working on, so please take a look at my comment below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDisplay.kt
```diff
@@ -43,4 +45,16 @@ interface MiniAppDisplay : LifecycleObserver {
...
      */
     fun navigateForward(): Boolean
+
+    /**
+     * Show default custom permissions UI to manage permission grant results.
+     * @param miniAppId mini app id as the key to retrieve data from cache.
+     * @param permissionsWithDescription a list of Pair includes MiniAppCustomPermissionType and description as String.
+     * @param callback to invoke the permission grant results for sending to the Mini App.
+     */
+    fun showCustomPermissionWindow(
```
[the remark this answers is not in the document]
khairul-alam-licon: Yeah, it makes sense. I am working on not to expose `showCustomPermissionWindow`, so we can use it internally.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/ui/MiniAppCustomPermissionWindow.kt
```diff
@@ -0,0 +1,148 @@
...
+    private lateinit var customPermissionLayout: View
+
+    fun displayPermissions(
+        miniAppId: String,
+        permissionsWithDescription: List<Pair<MiniAppCustomPermissionType, String>>
+    ) {
+        if (miniAppId.isEmpty())
+            return
+
+        val deniedPermissions = getDeniedPermissions(miniAppId, permissionsWithDescription)
```
corycaywood, line 48: I just noticed this implementation doesn't match iOS. We should filter the permissions before passing them to be displayed. So the Host App should only receive a list of denied permissions, not the ones which are already accepted.

If all of the permissions are already accepted, we shouldn't even call the Host App's implementation of `requestCustomPermissions`. In that case, we should just automatically return success to the mini app.

So basically, I think you should move this `getDeniedPermissions` functionality into [MiniAppMessageBridge.onRequestCustomPermissions](https://github.com/rakutentech/android-miniapp/blob/c63ee5683a1300a0fe498a7cbcb381230a4cf03c/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt#L167), and also prevent the Host App's `requestCustomPermissions` function from being called in the case that all of the permissions are already allowed. See the [iOS implementation](https://github.com/rakutentech/ios-miniapp/blob/d0719004c4b4057a5f673d4a94ec2db786224b82/MiniApp/Classes/JavascriptBridge/MiniAppJavaScriptPermissionManager.swift#L19) for details.

khairul-alam-licon: You are right! 
That's something we need to inside SDK. I'll update that issue in this PR.

khairul-alam-licon:
@corycaywood Thank you for your feedbacks. I have updated the PR based on last reviews.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -165,45 +175,31 @@ abstract class MiniAppMessageBridge {
...
-            }
-        } catch (e: Exception) {
-            callbackObj?.id?.let {
-                bridgeExecutor.postError(
-                    it,
-                    "${ErrorBridgeMessage.ERR_REQ_CUSTOM_PERMISSION} ${e.message}"
-                )
+                }
+            } catch (e: Exception) {
+                if (e.message.toString() == NO_IMPLEMENT_CUSTOM_PERMISSION) {
```
corycaywood, line 194: The typical way of doing this is to catch different exception types, so it seems a little strange to check the string message instead. Could you make an internal exception type for `CustomPermissionsNotImplementedException`?

Also, should we handle the case here where the Host App throws an exception in `requestCustomPermissions`?

So basically something this:
```
catch (e: CustomPermissionsNotImplementedException) {
  // Display permissions dialog....
} catch (e: Exception) {
  customPermissionBridgeDispatcher.postCustomPermissionError(e.localizedMessage)
}
```

corycaywood (approved):
Looks good 👍  Just see my suggestion below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: I have updated this part using `CustomPermissionsNotImplementedException`.

minh-rakuten (commented):
Nice.

I think we should remove the implementation in our Demo app (some UI dependencies will be removed too) to make the code in sample app less complicated. What do you think @khairul-alam-licon @corycaywood ?

khairul-alam-licon:
> I think we should remove the implementation in our Demo app (some UI dependencies will be removed too) to make the code in sample app less complicated. 

We need to keep the implementation in our Demo app to verify `requestCustomPermissions` implementation. This PR aims to add default screen when users doesn't implement `requestCustomPermissions` interface.

minh-rakuten (approved):
Need to update the status of [requestCustomPermission](https://github.com/khairul-alam-licon/android-miniapp/blob/master/miniapp/USERGUIDE.md#4-implement-the-miniappmessagebridge) default implementation is true.

corycaywood:
> > I think we should remove the implementation in our Demo app (some UI dependencies will be removed too) to make the code in sample app less complicated.
> 
> We need to keep the implementation in our Demo app to verify `requestCustomPermissions` implementation. This PR aims to add default screen when users doesn't implement `requestCustomPermissions` interface.

I'm not sure I'm understanding why we need to keep the demo app implementation. Do you mean just so we can verify that the Host App is able to implement their own custom permissions dialog?

But if we keep the demo app implementation, then we aren't able to verify the default implementation when testing. I think it makes more sense to remove the demo app implementation so we can simplify the code and be able to test the default. What do you think?

khairul-alam-licon:
Yeah, I have understood the context. I am updating by removing demo app implementation.

corycaywood (approved):
👍

minh-rakuten (approved):

PR #186, opened 2020-11-04
Cleanup miniAppId parameter while passing to userInfoBridgeDispatcher

khairul-alam-licon:
# Description
We can remove `miniAppId` from `userInfoBridgeDispatcher.onGetAccessToken(callbackId: String, miniAppId: String)` because we are already providing it while initializing `userInfoBridgeDispatcher`. 

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
@@ -104,22 +104,27 @@ abstract class MiniAppMessageBridge {
         }
     }
 
-    /** Handle the message from external. **/
+    @SuppressWarnings("UndocumentedPublicFunction", "LongMethod")
     @JavascriptInterface
     fun postMessage(jsonStr: String) {
         val callbackObj = Gson().fromJson(jsonStr, CallbackObj::class.java)
 
-        when (callbackObj.action) {
-            ActionType.GET_UNIQUE_ID.action -> onGetUniqueId(callbackObj)
-            ActionType.REQUEST_PERMISSION.action -> onRequestPermission(callbackObj)
-            ActionType.REQUEST_CUSTOM_PERMISSIONS.action -> onRequestCustomPermissions(jsonStr)
-            ActionType.SHARE_INFO.action -> onShareContent(callbackObj.id, jsonStr)
-            ActionType.LOAD_AD.action -> adBridgeDispatcher.onLoadAd(callbackObj.id, jsonStr)
-            ActionType.SHOW_AD.action -> adBridgeDispatcher.onShowAd(callbackObj.id, jsonStr)
-            ActionType.GET_USER_NAME.action -> userInfoBridgeDispatcher.onGetUserName(callbackObj.id)
-            ActionType.GET_PROFILE_PHOTO.action -> userInfoBridgeDispatcher.onGetProfilePhoto(callbackObj.id)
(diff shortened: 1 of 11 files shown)


corycaywood (approved):
LGTM 👍

minh-rakuten (commented):
Thanks, it's better 👍 

I just figure out there is the crash hole of `userInfoBridgeDispatcher` here.
The hostapp does not set `UserInfoBridgeDispatcher` and miniapp webview set `MiniAppMessageBridge ` -> the call on `postMessage()` is triggered -> `userInfoBridgeDispatcher` has not been set so it will crash there.

I suggest to init the default constructor of `UserInfoBridgeDispatcher` on `else` case of [this](https://github.com/rakutentech/android-miniapp/blob/67fb873719925450409ee6a55f7313d6e232690d/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt#L50).  The default constructor does not need to init the properties because all functions in `UserInfoBridgeDispatcher` already placed in try/catch.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -104,23 +104,27 @@ abstract class MiniAppMessageBridge {
...
+                ActionType.LOAD_AD.action -> adBridgeDispatcher.onLoadAd(callbackObj.id, jsonStr)
+                ActionType.SHOW_AD.action -> adBridgeDispatcher.onShowAd(callbackObj.id, jsonStr)
+                ActionType.GET_USER_NAME.action -> userInfoBridgeDispatcher.onGetUserName(callbackObj.id)
+                ActionType.GET_PROFILE_PHOTO.action -> userInfoBridgeDispatcher.onGetProfilePhoto(callbackObj.id)
+                ActionType.GET_ACCESS_TOKEN.action -> userInfoBridgeDispatcher.onGetAccessToken(callbackObj.id)
+                ActionType.SET_SCREEN_ORIENTATION.action -> screenBridgeDispatcher.onScreenRequest(callbackObj)
+            }
+        } catch (e: UninitializedPropertyAccessException) {
+            if (this::bridgeExecutor.isInitialized)
+                e.message?.let { bridgeExecutor.postError(callbackObj.id, it) }
```
minh-rakuten, line 127: This catch only happens when hostapp calls postMessage function so bridgeExecutor never be initialized, we don't need those executions.

There is no use case for miniapp triggers this error but as I mentioned we have this [issue](https://github.com/rakutentech/android-miniapp/pull/186#pullrequestreview-523035441) on `userInfoBridgeDispatcher`. Please make a fix for `userInfoBridgeDispatcher`.

khairul-alam-licon: > This catch only happens when hostapp calls postMessage function so bridgeExecutor never be initialized, we don't need those executions.

Ok, I have removed that try/catch block if we don't need to catch initialization error for `bridgeExecutor`. 

> There is no use case for miniapp triggers this error but as I mentioned we have this issue on userInfoBridgeDispatcher. Please make a fix for userInfoBridgeDispatcher, even though the catch already covers this error but we don't want the userInfoBridgeDispatcher is uninitialized when miniapp webview init.

Also, added an extra initialization for `userInfoBridgeDispatcher` inside `MiniAppMessageBridge.init`. Could you please review it when you'll have time? Thanks.

minh-rakuten (approved):
👍 
Just one more step to pass the unit test.

khairul-alam-licon:
Oh, I forgot to update unit test. I am updating.

khairul-alam-licon:
Thanks for the review.

PR #187, opened 2020-11-04
feat: Added exception types for some cases of server responses

corycaywood:
# Description
Added MiniAppHasNoPublishedVersionException for the case when the mini app id exists on the server but the server has no published versions, and added MiniAppNotFoundException for the case that the server returns 404.

## Links
MINI-1971

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkException.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkException.kt
@@ -15,6 +15,20 @@ open class MiniAppSdkException(message: String, cause: Throwable?) : Exception(m
 }
 
 /**
+ * Exception which is thrown when the server returns no published
+ * versions for the provided mini app ID.
+ */
+class MiniAppHasNoPublishedVersionException(appId: String) :
+    MiniAppSdkException("Server returned no published version info for the provided Mini App Id: $appId")
+
+/**
+ * Exception which is thrown when the provided mini app ID
+ * does not exist on the server.
+ */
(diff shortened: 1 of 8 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkException.kt
```diff
@@ -13,6 +13,20 @@ open class MiniAppSdkException(message: String, cause: Throwable?) : Exception(m
     constructor(message: String) : this(message, null)
 }
 
+/**
+ * Exception which is thrown when the server returns no published
+ * versions for the provided mini app ID.
+ */
+class MiniAppHasNoPublishedVersionException(appId: String)
+    : MiniAppSdkException("Server returned no published version info for the provided Mini App Id.")
```
khairul-alam-licon, line 21: Maybe we can append `appId` to this message.

khairul-alam-licon (approved):
Looks good 👍 
It seems CI failed due to detekt.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkException.kt
corycaywood: Thanks, I updated this.

khairul-alam-licon (approved):
👍

minh-rakuten (approved):
👍 👍 
Maybe we should add a document about our sdk exception.

corycaywood:
> 👍 👍
> Maybe we should add a document about our sdk exception.

Thanks, I added a note in the user guide.

minh-rakuten (approved):

PR #189, opened 2020-11-06
Add location permission to custom permissions management

khairul-alam-licon:
# Description
- Add a custom permission type for "rakuten.miniapp.device.location". 
- This new permission should also be available for permission management via `MiniApp.getCustomPermissions` and `MiniApp.setCustomPermissions`.
- Update the Demo App permissions management screen to contain this new permission.

## Links
- MINI-1954

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -44,24 +80,29 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
+        val miniAppId = miniAppCustomPermission.miniAppId
+        val allPermissions = prepareAllPermissionsToStore(miniAppId, supplied)
+        applyStoringPermissions(MiniAppCustomPermission(miniAppId, allPermissions))
+    }
+
+    /**
+     * Apply to be saved the grant results as json string to SharedPreferences.
+     * @param [miniAppCustomPermission] an object to contain the results per MiniApp.
+     */
+    @VisibleForTesting
```
minh-rakuten, line 99: Is it internal function or public one? You can remove this annotation if it can be public, otherwise should declare it as internal.

The same for other functions in this file.

minh-rakuten (commented):
There is a lack of spec implementation from mini-1954. I think it's okay to update in other PRs but let's update the description with the note to see which specs would be implemented later.

It's good to update the test cases now. The uncovered part of previous [PR](https://github.com/rakutentech/android-miniapp/pull/182) (`MiniAppSdkException` & `MiniAppMessageBridge`) should have test cases too.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
khairul-alam-licon: We don't need to make each function `internal` as the entire class is already `internal`.

minh-rakuten:
The order of the list has been changed when user interact with permissions. We should keep the consistence order.

khairul-alam-licon:
> It's good to update the test cases now. The uncovered part of previous PR (MiniAppSdkException & MiniAppMessageBridge) should have test cases too.

I want to address it in a separate PR.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
minh-rakuten: It makes sense but if you don't declare the function as internal, it means that other components in our sdk can call it freely. The internal here actually express that you want your function to be private but you make it internal for testing.

minh-rakuten (commented):
I think `Keeping consistence ordering of permissions` should be done in this PR.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -17,24 +17,60 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
+        return if (isDataExist(miniAppId)) {
+            try {
+                val cachedPermission: MiniAppCustomPermission = Gson().fromJson(
                     prefs.getString(miniAppId, ""),
                     object : TypeToken<MiniAppCustomPermission>() {}.type
                 )
+                val cachedPairs = cachedPermission.pairValues.toMutableList()
+
+                // detect any new change with comparing cached permissions and defaultDeniedList
+                // change means added new permission / removed existing permission from defaultDeniedList
```
corycaywood, line 40: Thanks for adding this - I didn't think of this case.

In the future, I think this won't be a problem after we add a permissions manifest, so each mini app will need to add a manifest which has a list of the permissions they want to use. And then we can get the list of permissions just by reading the manifest.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -17,24 +17,60 @@ internal class MiniAppCustomPermissionCache(context: Context) {
         "com.rakuten.tech.mobile.miniapp.custom.permissions.cache", Context.MODE_PRIVATE
     )
 
+    @VisibleForTesting
+    fun isDataExist(miniAppId: String) = prefs.contains(miniAppId)
```
corycaywood, line 21: Maybe this sounds more correct:
```suggestion
    fun doesDataExist(miniAppId: String) = prefs.contains(miniAppId)
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
khairul-alam-licon: Yes, I am having a similar thought. We can gradually improve the feature by having the permissions manifest.

khairul-alam-licon:
> I think Keeping consistence ordering of permissions should be done in this PR.

I am investigating the reason and will update this PR.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
khairul-alam-licon: I have added a `WantedPrivateButTesting` annotation to mark this kind of function to express the usage.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -161,17 +162,40 @@ abstract class MiniAppMessageBridge {
...
 
-            requestPermission(
-                MiniAppPermissionType.getValue(permissionParam.permission)
-            ) { isGranted -> onRequestPermissionsResult(
-                callbackId = callbackObj.id,
-                isGranted = isGranted
-            ) }
+            val permissionName = MiniAppPermissionType.getValue(permissionParam.permission)
+
+            if (permissionName.type == MiniAppPermissionType.LOCATION.type) {
```
khairul-alam-licon, line 167: @corycaywood I have added a solution for spec: `when a user tries to request location, we should first check that the "rakuten.miniapp.device.location" permission has been granted.`.  Do you think, we need to check it inside SDK? Or, I should move this implementation to HostApp?

Because, I was also reading a note in Jira: `This permission should be handled separately from the native Location permission which we handle via requestPermission`

p/s: I'll add unit test after confirming this fact.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
minh-rakuten: Wise solution.

minh-rakuten (dismissed):
Looks good to me.
Only test case left to solve.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
minh-rakuten: I tested and `WantedPrivateButTesting` does not behave like `VisibleForTesting`. It does not warn the developer to not use method publicly.

khairul-alam-licon: I guess, the annotation required to be stored to behave like `@VisibleForTesting`. So far, I am using `AnnotationRetention.SOURCE` policy, probably this is a reason. I haven't tried other policies yet. 

By the way, I just found that using `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` also may resolve our needs according to the [official guide](https://developer.android.com/studio/write/annotations). It's still visible to other classes, but by using it, I can have warning in other classes.  

e.g. 

Screen Shot 2020-11-13 at 15 08 29

@minh-rakuten What do you think?

minh-rakuten: I think you don't need to have own annotation. Just use `@VisibleForTesting` and declare `internal` for private functions.

khairul-alam-licon: Ok, I'll use `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` as officially recommended.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
corycaywood: @khairul-alam-licon Sorry, this wasn't very clear. I don't think we need to modify any of the current behavior of `onRequestPermission`. From the native SDK side, we should treat the device location permission request and the custom location permission request as completely separate and unrelated. The Host App also shouldn't have to worry about this.

I think we can add some functionality to the JS SDK to handle the interaction between the two. So I think the requirement for `when a user tries to request location, we should first check that the "rakuten.miniapp.device.location" permission has been granted.` can be handled in the JS SDK instead.

khairul-alam-licon: Okay, I am updating this part.

corycaywood (approved):
Looks good 👍

minh-rakuten (commented):
It's good to merge 👍 but MINI-1956 is not done yet so we cannot add this PR for the upcoming release.

CHANGELOG.md
```diff
@@ -8,11 +8,13 @@
 - **Change:** Added the default implementation for external link handler. Using [custom tab](https://developers.google.com/web/android/custom-tabs).
 - **Change:** `isTestMode` has been deprecated and replaced with `isPreviewMode` to adopt the Preview mode. See [this](miniapp/USERGUIDE.md#2-configure-sdk-settings-in-androidmanifestxml).
 - **Change:** `com.rakuten.tech.mobile.ras.AppId` has been deprecated, use `com.rakuten.tech.mobile.ras.ProjectId` instead. See [this](miniapp/USERGUIDE.md#2-configure-sdk-settings-in-androidmanifestxml).
+- **Feature:** Added `rakuten.miniapp.device.location` custom permission.
```
ts-leojoseph-r, line 11: It is `rakuten.miniapp.device.LOCATION` right ?

khairul-alam-licon: Ah yes, I haven't noticed it.

minh-rakuten (approved):
Let's update the bridge submodule and test one more time.

PR #190, opened 2020-11-06
refactor: default external link handler

minh-rakuten:
# Description
Default implementation of handling external link.
The link will be opened in [Custom Tab](https://developers.google.com/web/android/custom-tabs). There is no support for returning data back to sdk native webview at the moment.

## Links
MINI-1697

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -292,6 +292,8 @@ The mini app is loaded with the specific custom scheme and custom domain in mini
 
 In default, the external link will be opened in custom tab. See [this](https://developers.google.com/web/android/custom-tabs).
 
+HostApp also can implement their own way by passing `MiniAppNavigator` object to `MiniApp.create(appId: String, miniAppMessageBridge: MiniAppMessageBridge, miniAppNavigator: MiniAppNavigator)`.
+
 - Implement `MiniAppNavigator`.
 
 ```kotlin
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -290,8 +290,7 @@ override fun onBackPressed() {
 
 The mini app is loaded with the specific custom scheme and custom domain in mini app view.
 
-In default, the external link is also loaded in mini app view.
-It is possible for hostapp to load this external link with its own webview / browser.
+In default, the external link will be opened in custom tab. See [this](https://developers.google.com/web/android/custom-tabs).
 
 - Implement `MiniAppNavigator`.
 
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
@@ -31,23 +31,6 @@ abstract class MiniApp internal constructor() {
     abstract suspend fun listMiniApp(): List<MiniAppInfo>
 
     /**
-     * Creates a mini app.
-     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful.
-     * @param appId mini app id.
-     * @param miniAppMessageBridge the interface for communicating between host app & mini app
-     * @throws [MiniAppNotFoundException] when the specified mini app ID does not exist on the server
-     * @throws [MiniAppHasNoPublishedVersionException] when the specified mini app ID exists on the
-     * server but has no published versions
-     * @throws [MiniAppSdkException] when there is any other issue during fetching,
-     * downloading or creating the view.
-     */
-    @Throws(MiniAppNotFoundException::class, MiniAppHasNoPublishedVersionException::class, MiniAppSdkException::class)
-    abstract suspend fun create(
-        appId: String,
-        miniAppMessageBridge: MiniAppMessageBridge
-    ): MiniAppDisplay
-
-    /**
      * Same as [create(String, MiniAppMessageBridge)].
      * Use this to control external url loader.
      * @param miniAppNavigator allow host app to handle specific urls such as external link.
@@ -56,7 +39,7 @@ abstract class MiniApp internal constructor() {
     abstract suspend fun create(
         appId: String,
         miniAppMessageBridge: MiniAppMessageBridge,
-        miniAppNavigator: MiniAppNavigator
+        miniAppNavigator: MiniAppNavigator? = null
     ): MiniAppDisplay
 
     /**
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
@@ -31,9 +31,16 @@ abstract class MiniApp internal constructor() {
     abstract suspend fun listMiniApp(): List<MiniAppInfo>
 
     /**
-     * Same as [create(String, MiniAppMessageBridge)].
-     * Use this to control external url loader.
+     * Creates a mini app.
+     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful.
+     * @param appId mini app id.
+     * @param miniAppMessageBridge the interface for communicating between host app & mini app.
      * @param miniAppNavigator allow host app to handle specific urls such as external link.
+     * @throws [MiniAppNotFoundException] when the specified mini app ID does not exist on the server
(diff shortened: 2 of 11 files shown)


corycaywood (approved):
Looks good 👍  I left a couple of suggestions.

@Climbatize @ts-leojoseph-r  Do you think we should do something similar on iOS? Basically make the SDK's default secondary webview implementation use `SFSafariViewController` and then move the custom `WKWebview` implementation back into the demo app. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -94,6 +97,18 @@ internal class MiniAppWebView(
         destroy()
     }
 
+    @VisibleForTesting
+    internal fun setupMiniAppNavigator() {
+        if (miniAppNavigator == null) {
```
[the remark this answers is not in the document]
minh-rakuten: I agree that we should omit `MiniApp.create(appId: String, miniAppMessageBridge: MiniAppMessageBridge)`.

I have updated the following with `null` as the default value for `miniAppNavigator`.
```
abstract suspend fun create(
        appId: String,
        miniAppMessageBridge: MiniAppMessageBridge,
        miniAppNavigator: MiniAppNavigator? = null
): MiniAppDisplay
```

so I think we can omit immediately and no need to change to deprecated status.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -30,23 +30,6 @@ abstract class MiniApp internal constructor() {
...
-     * downloading or creating the view.
-     */
-    @Throws(MiniAppNotFoundException::class, MiniAppHasNoPublishedVersionException::class, MiniAppSdkException::class)
-    abstract suspend fun create(
-        appId: String,
-        miniAppMessageBridge: MiniAppMessageBridge
-    ): MiniAppDisplay
-
     /**
      * Same as [create(String, MiniAppMessageBridge)].
```
khairul-alam-licon, line 34: This line now looks redundant because `create(String, MiniAppMessageBridge)` doesn't exist.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
```diff
@@ -42,7 +42,7 @@ class MiniAppDisplayViewModel constructor(
     ) = viewModelScope.launch(Dispatchers.IO) {
         try {
             _isLoading.postValue(true)
-            miniAppDisplay = miniapp.create(appId, miniAppMessageBridge, miniAppNavigator)
```
khairul-alam-licon, line 45: Then, should we remove the sample app implementation of `miniAppNavigator` including the secondary webview? Because it's to upto HostApp if they wants to implement secondary webview or other technology with providing non-null `miniAppNavigator`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -30,23 +30,6 @@ abstract class MiniApp internal constructor() {
...
-    /**
-     * Creates a mini app.
-     * The mini app is downloaded, saved and provides a [MiniAppDisplay] when successful.
-     * @param appId mini app id.
-     * @param miniAppMessageBridge the interface for communicating between host app & mini app
-     * @throws [MiniAppNotFoundException] when the specified mini app ID does not exist on the server
-     * @throws [MiniAppHasNoPublishedVersionException] when the specified mini app ID exists on the
-     * server but has no published versions
-     * @throws [MiniAppSdkException] when there is any other issue during fetching,
-     * downloading or creating the view.
```
khairul-alam-licon, line 42: Do we need to remove those information for `@throws`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -30,23 +30,6 @@ abstract class MiniApp internal constructor() {
...
-     */
-    @Throws(MiniAppNotFoundException::class, MiniAppHasNoPublishedVersionException::class, MiniAppSdkException::class)
-    abstract suspend fun create(
-        appId: String,
-        miniAppMessageBridge: MiniAppMessageBridge
-    ): MiniAppDisplay
-
     /**
      * Same as [create(String, MiniAppMessageBridge)].
      * Use this to control external url loader.
```
khairul-alam-licon, line 35: We could specify in CHANGELOG and/or USERGUIDE i.e. when `miniAppNavigator` is `null`, MiniApp will invoke default implementation from SDK.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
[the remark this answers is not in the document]
khairul-alam-licon: I think, we can keep `create(String, MiniAppMessageBridge)` and marked it as just deprecated instead of directly delete it in earlier. It will be nicer to give a deprecation warning if it's possible. Also, need to describe this kind thing in CHANGELOG.

khairul-alam-licon (commented):
Going good 👍 I have left few comments.

ts-leojoseph-r:
> @Climbatize @ts-leojoseph-r Do you think we should do something similar on iOS? Basically make the SDK's default secondary webview implementation use `SFSafariViewController` and then move the custom `WKWebview` implementation back into the demo app. What do you think?

That's good idea, however, we should also think of a list of features that we are planning to support in the external webview.  If the external webview features are limited then I guess we can go ahead with the approach you are saying but if we plan to support similar features like the MiniAppWebView then I guess we need to think.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
minh-rakuten: The miniAppNavigator in demo app and the default one are different. We should keep the miniAppNavigator in our Demo app to show user how to send data back.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
minh-rakuten: good idea.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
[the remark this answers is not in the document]
minh-rakuten: The users do not see the change. The implementation of default null and `create(String, MiniAppMessageBridge)` are the same. That's why I think there is no need for deprecation warning because there is no deprecated method.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
minh-rakuten: ah let me update those information.

miniapp/USERGUIDE.md
```diff
@@ -290,8 +290,7 @@ override fun onBackPressed() {
 
 The mini app is loaded with the specific custom scheme and custom domain in mini app view.
 
-In default, the external link is also loaded in mini app view.
-It is possible for hostapp to load this external link with its own webview / browser.
```
khairul-alam-licon, line 294: I am not sure about the better place to write but it's good to mention like following:

"HostApp also can implement their own way by passing `MiniAppNavigator` object using `MiniApp.create(appId: String, miniAppMessageBridge: MiniAppMessageBridge, miniAppNavigator: MiniAppNavigator)` ". There is a sample implementation of `MiniAppNavigator` in [here](https://github.com/rakutentech/android-miniapp/blob/master/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/WebViewActivity.kt)"

khairul-alam-licon (approved):
Looks good 👍

miniapp/USERGUIDE.md
minh-rakuten: let me update it.

PR #192, opened 2020-11-09
fix: support back pressed operation for default permission UI 

khairul-alam-licon:
# Description
Previously, there was no support for back pressed when opening default permission UI. It should be the same operation while users press "Close" button. This PR added the fix for back button pressed.

## Links
- MINI-2033

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/ui/MiniAppCustomPermissionWindow.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/ui/MiniAppCustomPermissionWindow.kt
@@ -1,6 +1,7 @@
 package com.rakuten.tech.mobile.miniapp.permission.ui
 
 import android.app.Activity
+import android.view.KeyEvent
 import android.view.LayoutInflater
 import android.view.View
 import android.widget.TextView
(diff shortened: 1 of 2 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/ui/MiniAppCustomPermissionWindow.kt
```diff
@@ -107,8 +108,19 @@ internal class MiniAppCustomPermissionWindow(
 
         customPermissionLayout.findViewById<TextView>(R.id.permissionCloseWindow)
             .setOnClickListener {
-                customPermissionBridgeDispatcher.sendCachedCustomPermissions()
-                customPermissionAlertDialog.dismiss()
+                onNoPermissionsSaved()
             }
+
+        customPermissionAlertDialog.setOnKeyListener { _, keyCode, event ->
+            if (keyCode == KeyEvent.KEYCODE_BACK && event.action == KeyEvent.ACTION_DOWN) {
```
corycaywood, line 115: It seems a little strange to trigger this on `ACTION_DOWN`. Usually I expect a button to trigger only after I've released it, so on `ACTION_UP`.  What do you think @minh-rakuten @khairul-alam-licon ?

khairul-alam-licon: Yes, `KeyEvent.ACTION_UP` makes more sense in this usecase. I have updated this part.

corycaywood (approved):
Looks good 👍

minh-rakuten (approved):
While I'm playing with permission dialog, I have this crash:
```
com.rakuten.tech.mobile.miniapp.testapp.debug E/Parcel: Class not found when unmarshalling: androidx.fragment.app.FragmentManagerState
    java.lang.ClassNotFoundException: androidx.fragment.app.FragmentManagerState
```

@khairul-alam-licon @corycaywood The problem is I cannot reproduce so we should notice this.

khairul-alam-licon:
> com.rakuten.tech.mobile.miniapp.testapp.debug E/Parcel: Class not found when unmarshalling: androidx.fragment.app.FragmentManagerState
    java.lang.ClassNotFoundException: androidx.fragment.app.FragmentManagerState

This is an issue I have encountered in other branches as well. It's not related to this PR's change, but we should address it in another PR.

PR #193, opened 2020-11-10
sample: external webview update

minh-rakuten:
# Description
Turn on some setting flags on external webview to cover some errors.

Why: The hostapp is responsible for external webview implementation but we don't want our clients & partners see errors while testing in our demo app.

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@
 - **Feature:** Mini App can call media execution play/pause programmatically.
 - **Change:** Added the default implementation for external link handler. Using [custom tab](https://developers.google.com/web/android/custom-tabs).
 
+**Sample App**
+- **Change:** Update setting of external webview.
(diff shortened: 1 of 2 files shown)


khairul-alam-licon:
Just a suggestion: Since we are handling two webviews, so we can have a SDK's function for applying common settings in both side.

```
  fun applyCommonWebViewSettings(settings: WebSettings) {
        settings.javaScriptEnabled = true
        settings.allowUniversalAccessFromFileURLs = true
        settings.domStorageEnabled = true
        settings.databaseEnabled = true
    }
```

And, we can provide it as a userguide for HostApp developers. Of course if they want to inherit the common web settings from the primary webview.

khairul-alam-licon (approved):
Looks good 👍

corycaywood (approved):
LGTM 👍

PR #196, opened 2020-11-12
Update docs and changelog for v2.5.0

corycaywood:
# Description
Updated changelog and also added some additional information about throwing errors to a mini app.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
@@ -69,7 +69,10 @@ abstract class MiniAppMessageBridge {
     @VisibleForTesting
     internal fun createBridgeExecutor(webViewListener: WebViewListener) = MiniAppBridgeExecutor(webViewListener)
 
-    /** Get provided id of mini app for any purpose. **/
+    /**
+     * Get provided id of mini app for any purpose.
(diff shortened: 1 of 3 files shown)


minh-rakuten (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -69,7 +69,10 @@ abstract class MiniAppMessageBridge {
     @VisibleForTesting
     internal fun createBridgeExecutor(webViewListener: WebViewListener) = MiniAppBridgeExecutor(webViewListener)
 
-    /** Get provided id of mini app for any purpose. **/
+    /**
+     * Get provided id of mini app for any purpose.
+     * You can also throw an [Exception] from this method to pass an error message to the mini app.
```
khairul-alam-licon, line 74: Is `getUniqueId()` throwing any exception?

khairul-alam-licon (approved):
Looks good 👍 
Just a qus.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
corycaywood: There is a case where the Host App wants to throw an error to the mini app when they fail to generate the Unique ID. For example, the Unique ID could be based on the User's login and they need to fetch it from an API, so the API request could fail.

PR #197, opened 2020-11-13
fix: redirecting to default browser when window.open anchor

khairul-alam-licon:
# Description
This PR adds a change for secondary webview in sample app, something like redirecting to default browser when there is `window.open` / `_blank` anchor href. Since, SDK has a default implementation for external url handling, so this change doesn't affect anything in sample app. But still we can test this PR by using `miniapp.create(appId, miniAppMessageBridge, miniAppNavigator)` inside `MiniAppDisplayViewModel`.

Note: There is a way to detect `window.open` distinctly in `WebChromeClient.onCreateWindow()`, but it's not reliable solution, so haven't included it.

## Links
- MINI-1928

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/component/SampleExternalWebView.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/component/SampleExternalWebView.kt
@@ -2,6 +2,8 @@ package com.rakuten.tech.mobile.testapp.ui.component
 
 import android.annotation.SuppressLint
 import android.content.Context
+import android.os.Message
+import android.webkit.WebChromeClient
 import android.webkit.WebView
 import android.webkit.WebViewClient
 import com.rakuten.tech.mobile.miniapp.navigator.MiniAppExternalUrlLoader
@@ -15,7 +17,9 @@ class SampleExternalWebView(context: Context, url: String, sampleWebViewClient: 
         settings.domStorageEnabled = true
         settings.databaseEnabled = true
         settings.mediaPlaybackRequiresUserGesture = false
+        settings.setSupportMultipleWindows(true)
         webViewClient = sampleWebViewClient
+        webChromeClient = SampleWebChromeClient(context)
         loadUrl(url)
     }
 }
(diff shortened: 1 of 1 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/component/SampleExternalWebView.kt
```diff
@@ -25,3 +29,18 @@ class SampleWebViewClient(private val miniAppExternalUrlLoader: MiniAppExternalU
...
+
+class SampleWebChromeClient(val context: Context) : WebChromeClient() {
+
+    override fun onCreateWindow(
+        view: WebView?,
+        isDialog: Boolean,
+        isUserGesture: Boolean,
+        resultMsg: Message?
+    ): Boolean {
+        // redirect to default browser when there is window.open / _blank type anchor
```
minh-rakuten, line 41: so finally we decided to make this no action or open default browser? @corycaywood

corycaywood: I think probably this is okay - to me it seems like the more reasonable behavior for `window.open` and `_blank` to be sent to the default browser. 

But we would also need to change iOS to match this. Is this possible @Climbatize @ts-leojoseph-r ?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/component/SampleExternalWebView.kt
```diff
@@ -25,3 +29,18 @@ class SampleWebViewClient(private val miniAppExternalUrlLoader: MiniAppExternalU
...
+        (resultMsg?.obj as WebView.WebViewTransport).webView = WebView(context)
```
corycaywood, line 42: Maybe we should also check `isUserGesture`? The docs say this can prevent unwanted pop-ups:
```
if (isUserGesture) {
 // ...
} else {
  return false
}
```

What do you think?

corycaywood (approved):
Looks good 👍

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/component/SampleExternalWebView.kt
khairul-alam-licon: Yes, it's okay to check with `isUserGesture`, I have updated this part.

minh-rakuten (approved):
👍 👍

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/component/SampleExternalWebView.kt
Climbatize: I missed your mention. I think it's quite possible, at least for blank targets

PR #200, opened 2020-11-16
api: remove hostapp version

minh-rakuten:
# Description
There is no usage of hostapp version so this PR removes it in sdk and sample app.

## Links
MINI-2028

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
 **SDK**
 - **Feature:** Mini App can call media execution play/pause programmatically.
 - **Change:** Added the default implementation for external link handler. Using [custom tab](https://developers.google.com/web/android/custom-tabs).
-- **Removed:** The SDK setting `Host App Version` has been removed. There is no usage of it.
 - **Change:** `isTestMode` has been deprecated and replaced with `isPreviewMode` to adopt the Preview mode. See [this](miniapp/USERGUIDE.md#2-configure-sdk-settings-in-androidmanifestxml).
 
 **Sample App**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **SDK**
 - **Feature:** Mini App can call media execution play/pause programmatically.
 - **Change:** Added the default implementation for external link handler. Using [custom tab](https://developers.google.com/web/android/custom-tabs).
+- **Removed:** The SDK setting `Host App Version` has been removed. There is no usage of it.
 
 **Sample App**
 - **Change:** Update setting of external webview.
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -43,7 +43,6 @@ The SDK is configured via manifest meta-data, the configurable values are:
 
 **Note:**  
 * We don't currently host a public API, so you will need to provide your own Base URL for API requests.
-* All meta-data values must be string values, including the value for `com.rakuten.tech.mobile.miniapp.HostAppVersion`. For example it could be set to the string value `1.0.0`, but if you need to use a number value such as `1.0` or `1`, then you must declare the value in your string resources (`res/values/strings.xml`) and reference the string ID in the manifest, for example `@string/app_version`.
 * The host app info is the string which is appended to user-agent of webview. It should be a meaningful keyword such as host app name to differentiate other host apps.
 
 In your `AndroidManifest.xml`:
@@ -66,11 +65,6 @@ In your `AndroidManifest.xml`:
         <meta-data
             android:name="com.rakuten.tech.mobile.ras.ProjectSubscriptionKey"
             android:value="your_subscription_key" />
-
-        <!-- Version of your app - used to determine feature compatibility for Mini App -->
-        <meta-data
-            android:name="com.rakuten.tech.mobile.miniapp.HostAppVersion"
-            android:value="your_app_version" />
 
         <!-- Optional User Agent Information relating to the host app -->
         <meta-data
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -39,7 +39,6 @@ The SDK is configured via manifest meta-data, the configurable values are:
 | Base URL                     | String  | `com.rakuten.tech.mobile.miniapp.BaseUrl`              | ❌         | 🚫        |
 | App ID                       | String  | `com.rakuten.tech.mobile.ras.AppId`                    | ❌         | 🚫        |
 | RAS Project Subscription Key | String  | `com.rakuten.tech.mobile.ras.ProjectSubscriptionKey`   | ❌         | 🚫        |
-| Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
 | Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
 
 **Note:**  
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
@@ -5,7 +5,6 @@ package com.rakuten.tech.mobile.miniapp
  * @property baseUrl Base URL used for retrieving a Mini App.
  * @property rasAppId App ID for the Platform API.
  * @property subscriptionKey Subscription Key for the Platform API.
- * @property hostAppVersionId Version of the host app, used to determine feature compatibility for Mini App.
  * @property hostAppUserAgentInfo User Agent information from Host App.
  * @property isTestMode Whether the sdk is making use of the Test API Endpoints for under "Testing" mini apps.
  */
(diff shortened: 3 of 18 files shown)


miniapp/USERGUIDE.md
```diff
@@ -39,7 +39,6 @@ The SDK is configured via manifest meta-data, the configurable values are:
 | Base URL                     | String  | `com.rakuten.tech.mobile.miniapp.BaseUrl`              | ❌         | 🚫        |
 | App ID                       | String  | `com.rakuten.tech.mobile.ras.AppId`                    | ❌         | 🚫        |
 | RAS Project Subscription Key | String  | `com.rakuten.tech.mobile.ras.ProjectSubscriptionKey`   | ❌         | 🚫        |
-| Host App Version             | String  | `com.rakuten.tech.mobile.miniapp.HostAppVersion`       | ❌         | 🚫        |
 | Host App User Agent Info     | String  | `com.rakuten.tech.mobile.miniapp.HostAppUserAgentInfo` | ✅         | 🚫        |
 
 **Note:**
```
khairul-alam-licon, line 44: There is a note written for Host App Version, something like `All meta-data values must be string values, including the value for com.rakuten.tech.mobile.miniapp.HostAppVersion...`, is it okay to remove that line too?

khairul-alam-licon: Also, removing the following line from USERGUIDE?

```
<meta-data
   android:name="com.rakuten.tech.mobile.miniapp.HostAppVersion"
   android:value="your_app_version" />
```

miniapp/src/test/java/com/rakuten/tech/mobile/miniapp/api/ApiClientSpec.kt
```diff
@@ -136,8 +136,7 @@ open class ApiClientSpec {
         ApiClient(
             baseUrl = TEST_URL_HTTPS_2,
             rasAppId = TEST_HA_ID_APP,
-            subscriptionKey = TEST_HA_SUBSCRIPTION_KEY,
-            hostAppVersionId = TEST_HA_ID_VERSION
+            subscriptionKey = TEST_HA_SUBSCRIPTION_KEY
         )
     }
```
khairul-alam-licon, line 142: Also, we can remove-

```
private fun createApiClient(
        ...
        hostAppVersionId: String = TEST_HA_ID_VERSION,
        ...
    )
```

and, finally remove `TEST_HA_ID_VERSION` from `TestData`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
```diff
@@ -13,11 +13,10 @@ data class MiniAppSdkConfig(
     val baseUrl: String,
     val rasAppId: String,
     val subscriptionKey: String,
-    val hostAppVersionId: String,
```
khairul-alam-licon, line 16: So, we can remove ` * @property hostAppVersionId Version of the host app, used to determine feature compatibility for Mini App.` comment in this file?

khairul-alam-licon (commented):
Looking good, do we need to mention this change in CHANGELOG as well?

minh-rakuten:
@khairul-alam-licon Thank you for figuring out. I have removed them.

khairul-alam-licon (dismissed):
Looks good to me 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
```diff
@@ -5,19 +5,17 @@ package com.rakuten.tech.mobile.miniapp
...
  * @property subscriptionKey Subscription Key for the Platform API.
- * @property hostAppVersionId Version of the host app, used to determine feature compatibility for Mini App.
  * @property hostAppUserAgentInfo User Agent information from Host App.
  * @property isTestMode Whether the sdk is making use of the Test API Endpoints for under "Testing" mini apps.
  */
 data class MiniAppSdkConfig(
     val baseUrl: String,
     val rasAppId: String,
     val subscriptionKey: String,
-    val hostAppVersionId: String,
```
corycaywood, line 16: I think we wanted to keep the `hostAppVersionId` setting, but just remove it from the API requests. This is because we might want to use it in the future for returning different versions of a mini app based on the version of the Host App (that was the settings original purpose, but the functionality was never added).

I guess we could remove it, but this is a breaking change since `MiniAppSdkConfig` is public. So we would need to just deprecate this parameter instead, so this gets a little complicated.

What do you think, should we deprecate it or just leave the setting here and not use it for anything?

minh-rakuten: We can leave this setting here. This PR implementation causes no impact whether the user defines the manifest of hostapp version or not.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt
```diff
@@ -76,9 +70,8 @@ class MiniappSdkInitializer : ContentProvider() {
                 baseUrl = manifestConfig.baseUrl(),
                 rasAppId = manifestConfig.rasAppId(),
                 subscriptionKey = manifestConfig.subscriptionKey(),
-                hostAppVersionId = manifestConfig.hostAppVersion(),
                 hostAppUserAgentInfo = manifestConfig.hostAppUserAgentInfo(),
-                isTestMode = manifestConfig.isPreviewMode() || manifestConfig.isTestMode()
```
khairul-alam-licon, line 81: Thanks for fixing the conflicts, at this area, I think we kept [this line in master branch](https://github.com/rakutentech/android-miniapp/blob/e7dbff0de08ae436102e03752bc6d2d50c00eda0/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt#L81).

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 **SDK**
 - **Feature:** Mini App can call media execution play/pause programmatically.
 - **Change:** Added the default implementation for external link handler. Using [custom tab](https://developers.google.com/web/android/custom-tabs).
+- **Removed:** The SDK setting `Host App Version` has been removed. There is no usage of it.
```
khairul-alam-licon, line 8: Should we update this line as we have `hostAppVersionId` in `MiniAppSdkConfig`?

minh-rakuten: I have removed this line. I guess it's the most proper way 😆.

khairul-alam-licon: Okay, I think that's should be fine 👌

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt
minh-rakuten: why do we use `isTestMode` instead `isPreviewMode`?

khairul-alam-licon: I just checked that there is no difference between:
```
`isTestMode = manifestConfig.isPreviewMode() || manifestConfig.isTestMode()` 

// or

`isPreviewMode = manifestConfig.isPreviewMode() || manifestConfig.isTestMode()`. 
```

But, both are breaking changes when the HostApp _(by mistake)_ define both in `gradle.properties` configuration. 
I just checked this part, actually the more appropriate solution I think:

```
       MiniApp.init(
            context = context,
            miniAppSdkConfig = MiniAppSdkConfig(
                ...
              isTestMode = manifestConfig.isTestMode(),
              isPreviewMode = manifestConfig.isPreviewMode()
            )
        )
```

We can merge this PR as it is, then I'll investigate it and make a separate PR if it needs the new solution.

@minh-rakuten @corycaywood What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
```diff
@@ -13,18 +13,18 @@ data class MiniAppSdkConfig(
...
-    internal val key = "$baseUrl-$isPreviewMode-$rasAppId-$subscriptionKey-$hostAppVersionId"
+    internal val key = "$baseUrl-$isPreviewMode-$rasAppId-$subscriptionKey"
 
     @Deprecated("Use isPreviewMode instead of isTestMode")
     constructor(
         baseUrl: String,
         rasAppId: String,
         subscriptionKey: String,
-        hostAppVersionId: String,
+        hostAppVersionId: String = "",
```
khairul-alam-licon, line 27: Do we need to add it in secondary constructor? We haven't released this constructor yet for any existing client. I think, we could remove it. But, with the primary constructor, we can add a `@Deprecated` annotation e.g.

```
data class MiniAppSdkConfig(
    ...
    @Deprecated("It has been removed from api endpoints")
    val hostAppVersionId: String = "",
    ...
) ...
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt
khairul-alam-licon: This part is still using the primary constructor of `MiniAppSdkConfig`, so `isPreviewMode` was appropriate to use in this case.
Minh was right for the change.

Sorry, Project ID PR wasn't merged yet which is using the secondary constructor for backward compatibility. So, confusion arises.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
minh-rakuten: No, the only change is I put the default value so nothing is deprecated here.

minh-rakuten:
@khairul-alam-licon @corycaywood I have pushed the merge from master.

corycaywood (approved):
Looks good 👍

khairul-alam-licon (approved):
Looks good, it can be merged after fixing the conflicts successfully. 
This PR faces lots of conflicts, sorry about that.

PR #205, opened 2020-11-20
fix: show preview of granted permissions on permissions screen

khairul-alam-licon:
# Description
On the Permissions screen where each Mini App is listed, add a comma separated list showing the granted permissions for each Mini App. See the attached screenshot of the iOS Demo App. 

## Links
- MINI-1777

# Checklist


testapp/build.gradle
```diff
@@ -104,6 +104,7 @@ dependencies {
     implementation "androidx.core:core-ktx:$androidx_coreKtx"
     implementation "androidx.lifecycle:lifecycle-extensions:$androidx_lifecycle"
     implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:$androidx_lifecycle"
+    implementation "androidx.activity:activity:$androidx_activity"
```
minh-rakuten, line 107: Why do we need this dependency and why we choose `1.2.0-alpha04`?

minh-rakuten (commented):
There is a crash when tapping on Permission setting.
```
Caused by: java.lang.IllegalArgumentException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkParameterIsNotNull, parameter type
        at com.rakuten.tech.mobile.testapp.ui.permission.PermissionsUtilsKt.toReadableName(Unknown Source:2)```

testapp/build.gradle
khairul-alam-licon: I am using `registerForActivityResult` instead of `startActivityForResult` by using this dependency. See [this](https://developer.android.com/training/basics/intents/result).
I'll update it to current stable version.

khairul-alam-licon:
> There is a crash when tapping on Permission setting.

@minh-rakuten Can you tell more info, like the reproducing steps?

corycaywood (approved):
Looks good 👍

testapp/build.gradle
minh-rakuten: I think we should avoid using alpha or beta version because there could be bugs from google. However, if it is required then it's okay.

minh-rakuten (approved):
👍 👍

PR #208, opened 2020-11-24
refactor error messages with project id when `MiniAppNotFoundException` exception

khairul-alam-licon:
# Description
Refactor messages with specifying project id when `MiniAppNotFoundException` exception. 

## Links
MINI-2175

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
@@ -111,7 +111,7 @@ internal class RetrofitRequestExecutor(
         }
     }
 
-    @Throws(MiniAppSdkException::class)
+    @Throws(MiniAppSdkException::class, MiniAppNotFoundException::class)
     @Suppress("MagicNumber", "ThrowsCount")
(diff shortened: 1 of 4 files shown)


corycaywood (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -111,7 +111,7 @@ internal class RetrofitRequestExecutor(
         }
     }
 
-    @Throws(MiniAppSdkException::class)
+    @Throws(MiniAppSdkException::class, MiniAppNotFoundException::class)
```
minh-rakuten, line 114: `MiniAppNotFoundException` is `MiniAppSdkException` so I don't think we need to add it here.

khairul-alam-licon: I have followed the way like `MiniApp.create` where are keeping both. I think, the main reason is distinguish both as separate exception while providing error message.

minh-rakuten (approved):
👍

corycaywood (approved):

PR #211, opened 2020-11-25
feat: add an option to load mini app by url (MINI-1981)

imaginaris:
# Description
URL loading available when preview mode is off (not sure if that's correct).
I'll write more tests once the main part is accepted.

## Links
MINI-1981
SDKCF-3025

# Checklist
- [X] I have read the [contributing guidelines](../.github/CONTRIBUTING.md).
- [ ] I have completed the [SDK Development Learning Path](../.github/CONTRIBUTING.md#SDK-Development-Learning-Path) (if submitting a feature PR)
- [/] I wrote/updated tests for new/changed code
- [X] I removed all sensitive data before every commit, including API endpoints and keys
- [X] I ran `./gradlew check` without errors

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -14,6 +14,10 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import kotlinx.coroutines.Job
 import java.io.File
+import java.io.IOException
+import java.net.ConnectException
+import java.net.HttpURLConnection
+import java.net.URL
 
 @Suppress("SwallowedException", "TooManyFunctions")
 internal class MiniAppDownloader(
@@ -23,6 +27,9 @@ internal class MiniAppDownloader(
     private val verifier: CachedMiniAppVerifier,
     private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
 ) : UpdatableApiClient {
+
+    @Suppress("MagicNumber")
+    private val validHTTPResponseCodes = 100..399
 
     suspend fun getMiniApp(appId: String): Pair<String, MiniAppInfo> = try {
         val miniAppInfo = apiClient.fetchInfo(appId)
@@ -35,6 +42,24 @@ internal class MiniAppDownloader(
         onGetMiniApp(miniAppInfo)
     } catch (netError: MiniAppNetException) {
         onNetworkError(miniAppInfo)
+    }
+
+    @Suppress("ThrowsCount")
+    fun validateHttpAppUrl(url: String) {
+        var connection: HttpURLConnection? = null
+        try {
+            connection = URL(url).openConnection() as HttpURLConnection
+            val code = connection.responseCode
+            if (code !in validHTTPResponseCodes) {
+                throw MiniAppSdkException("Invalid URL error")
+            }
+        } catch (netError: ConnectException) {
+            throw MiniAppNotFoundException("Invalid URL error")
+        } catch (ioError: IOException) {
+            throw sdkExceptionForInternalServerError()
+        } finally {
+            connection?.disconnect()
+        }
     }
 
     private suspend fun onGetMiniApp(miniAppInfo: MiniAppInfo): Pair<String, MiniAppInfo> {
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
@@ -4,10 +4,10 @@ import androidx.annotation.VisibleForTesting
 import com.rakuten.tech.mobile.miniapp.api.ApiClient
 import com.rakuten.tech.mobile.miniapp.api.ApiClientRepository
 import com.rakuten.tech.mobile.miniapp.display.Displayer
-import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermission
-import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
 import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageBridge
 import com.rakuten.tech.mobile.miniapp.navigator.MiniAppNavigator
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermission
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
 
 internal class RealMiniApp(
     private val apiClientRepository: ApiClientRepository,
@@ -66,6 +66,23 @@ internal class RealMiniApp(
             displayer.createMiniAppDisplay(
                 basePath,
                 miniAppInfo,
+                miniAppMessageBridge,
+                miniAppNavigator,
+                miniAppCustomPermissionCache
+            )
+        }
+    }
+
+    override suspend fun createWithUrl(
+        appUrl: String,
+        miniAppMessageBridge: MiniAppMessageBridge,
+        miniAppNavigator: MiniAppNavigator?
+    ): MiniAppDisplay = when {
+        appUrl.isBlank() -> throw sdkExceptionForInvalidArguments()
+        else -> {
+            miniAppDownloader.validateHttpAppUrl(appUrl)
+            displayer.createMiniAppDisplay(
+                appUrl,
                 miniAppMessageBridge,
                 miniAppNavigator,
                 miniAppCustomPermissionCache
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/Displayer.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/Displayer.kt
@@ -24,4 +24,18 @@ internal class Displayer(private val context: Context, private val hostAppUserAg
         hostAppUserAgentInfo = hostAppUserAgentInfo,
         miniAppCustomPermissionCache = miniAppCustomPermissionCache
     )
+
+    fun createMiniAppDisplay(
+        appUrl: String,
+        miniAppMessageBridge: MiniAppMessageBridge,
+        miniAppNavigator: MiniAppNavigator?,
+        miniAppCustomPermissionCache: MiniAppCustomPermissionCache
+    ): MiniAppDisplay = RealMiniAppDisplay(
+        context = context,
+        appUrl = appUrl,
+        miniAppMessageBridge = miniAppMessageBridge,
+        miniAppNavigator = miniAppNavigator,
+        hostAppUserAgentInfo = hostAppUserAgentInfo,
+        miniAppCustomPermissionCache = miniAppCustomPermissionCache
+    )
 }
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
@@ -1,0 +1,45 @@
+package com.rakuten.tech.mobile.miniapp.display
+
+import android.content.Context
+import com.rakuten.tech.mobile.miniapp.MiniAppInfo
+import com.rakuten.tech.mobile.miniapp.MiniAppScheme
+import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageBridge
+import com.rakuten.tech.mobile.miniapp.navigator.MiniAppNavigator
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
+import java.util.*
+
+internal class MiniAppHttpWebView(
+    context: Context,
+    miniAppTitle: String,
+    val appUrl: String,
+    miniAppMessageBridge: MiniAppMessageBridge,
+    miniAppNavigator: MiniAppNavigator?,
+    hostAppUserAgentInfo: String,
+    miniAppWebChromeClient: MiniAppWebChromeClient = MiniAppWebChromeClient(context, miniAppTitle),
+    miniAppCustomPermissionCache: MiniAppCustomPermissionCache
+) : MiniAppWebView(
+    context,
+    "",
+    MiniAppInfo.empty(),
+    miniAppMessageBridge,
+    miniAppNavigator,
+    hostAppUserAgentInfo,
+    miniAppWebChromeClient,
+    miniAppCustomPermissionCache
+) {
+    init {
+        miniAppScheme = MiniAppScheme.schemeWithCustomUrl(appUrl)
+        miniAppId = UUID.randomUUID().toString() // some id is needed to handle custom permissions
+        commonInit()
+    }
+
+    override fun getMiniAppWebViewClient(): MiniAppWebViewClient = MiniAppWebViewClient(
+        context,
+        null,
+        miniAppNavigator!!,
+        externalResultHandler,
+        miniAppScheme
+    )
+
+    override fun getLoadUrl(): String = appUrl
+}
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
@@ -1,12 +1,13 @@
 package com.rakuten.tech.mobile.miniapp.display
 
 import android.content.Context
+import androidx.annotation.VisibleForTesting
 import com.rakuten.tech.mobile.miniapp.MiniAppInfo
 import com.rakuten.tech.mobile.miniapp.MiniAppScheme
 import com.rakuten.tech.mobile.miniapp.js.MiniAppMessageBridge
 import com.rakuten.tech.mobile.miniapp.navigator.MiniAppNavigator
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
-import java.util.*
+import java.util.UUID
 
 internal class MiniAppHttpWebView(
     context: Context,
@@ -42,4 +43,14 @@ internal class MiniAppHttpWebView(
     )
 
     override fun getLoadUrl(): String = appUrl
+
+    override fun onDetachedFromWindow() {
+        super.onDetachedFromWindow()
+        miniAppCustomPermissionCache.removeId(miniAppId)
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun callOnDetached() {
+        onDetachedFromWindow()
+    }
 }
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
@@ -22,18 +22,19 @@ private const val SUB_DOMAIN_PATH = "miniapp"
 private const val MINI_APP_INTERFACE = "MiniAppAndroid"
 
 @SuppressLint("SetJavaScriptEnabled")
-internal class MiniAppWebView(
+internal open class MiniAppWebView(
     context: Context,
     val basePath: String,
     val miniAppInfo: MiniAppInfo,
     val miniAppMessageBridge: MiniAppMessageBridge,
     var miniAppNavigator: MiniAppNavigator?,
     val hostAppUserAgentInfo: String,
-    val miniAppWebChromeClient: MiniAppWebChromeClient = MiniAppWebChromeClient(context, miniAppInfo),
+    val miniAppWebChromeClient: MiniAppWebChromeClient = MiniAppWebChromeClient(context, miniAppInfo.displayName),
     val miniAppCustomPermissionCache: MiniAppCustomPermissionCache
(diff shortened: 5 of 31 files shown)


miniapp/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <domain-config cleartextTrafficPermitted="true">
+	<domain includeSubdomains="true">10.0.2.2</domain> <!--localhost-->
```
imaginaris, line 4: I didn't want to allow every url so if someone uses other http address than localhost, they'd need to put the address in this file.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -68,17 +89,31 @@ internal class RealMiniAppDisplay(
...
-                context = context,
-                basePath = basePath,
-                miniAppInfo = miniAppInfo,
-                miniAppMessageBridge = miniAppMessageBridge,
-                miniAppNavigator = miniAppNavigator,
-                hostAppUserAgentInfo = hostAppUserAgentInfo,
-                miniAppCustomPermissionCache = miniAppCustomPermissionCache
-            )
+            if (miniAppUrl != null) {
+                miniAppWebView = MiniAppWebView(
```
minh-rakuten, line 96: This can be replace with `MiniAppHttpWebView`. See [here](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt).

miniapp/src/main/res/xml/network_security_config.xml
minh-rakuten: This one is gud but the spec declares that http is allowance. @corycaywood What do you think?

testapp/src/main/AndroidManifest.xml
```diff
@@ -5,6 +5,7 @@
 
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
+    <uses-permission android:name="android.permission.INTERNET" />
```
minh-rakuten, line 8: The demo app does not need this. MiniApp SDK can access internet itself.

minh-rakuten (commented):
Welcome and thank you for android miniapp contribution 🎉 🎉 

It looks good. Here are some concerns.

1/ URL loading available when preview mode is off -> it's okay for now. The miniapp team will update it later.

2/ To keep the miniapp webview maintainable and simple, let's have a separated class for loading with http url. I have push [this](https://github.com/minh-rakuten/android-miniapp/commit/03e78f86f5130608ce4f94e210b47ae6f4beec72?branch=03e78f86f5130608ce4f94e210b47ae6f4beec72&diff=unified#diff-a8a5ca1f1ff16520549361cd01bf8d10fc2543d3a79f4f71075b7c2c19150a3cR96) based on your PR. Please feel free to consider it as suggestion or merge it. There are three changes file: [MiniAppWebView](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt), [MiniAppHttpWebView](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt), [RealMiniAppDisplay](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt).

3/ Minor comments below.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
```diff
@@ -63,6 +63,31 @@ class MiniAppDisplayViewModel constructor(
...
+    ) = viewModelScope.launch(Dispatchers.IO) {
+        try {
+            _isLoading.postValue(true)
+            miniAppDisplay = miniapp.createWithUrl(appUrl, miniAppMessageBridge)
+            hostLifeCycle?.addObserver(miniAppDisplay)
+            _miniAppView.postValue(miniAppDisplay.getMiniAppView(context))
+        } catch (e: MiniAppSdkException) {
+            e.printStackTrace()
+            when (e) {
+                is MiniAppHasNoPublishedVersionException ->
```
khairul-alam-licon, line 80: Do we need to catch this exception? Because `MiniApp.createWithUrl` doesn't throw this one.

khairul-alam-licon (commented):
Looking nice in the first place 👍 
I was confused in some points while reading few variable names as we can accept a normal HTTP url here.  I have written my thoughts. 

Also, can we include the documentation in changelog and userguide? Using a separate PR is totally fine for the user-documentation.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
```diff
@@ -63,6 +63,31 @@ class MiniAppDisplayViewModel constructor(
...
 
+    fun obtainMiniAppDisplayUrl(
+        context: Context,
+        appUrl: String,
+        miniAppMessageBridge: MiniAppMessageBridge,
+        miniAppNavigator: MiniAppNavigator
+    ) = viewModelScope.launch(Dispatchers.IO) {
+        try {
+            _isLoading.postValue(true)
+            miniAppDisplay = miniapp.createWithUrl(appUrl, miniAppMessageBridge)
```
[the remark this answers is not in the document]
imaginaris: True. I missed that 👍

imaginaris:
> 2/ To keep the miniapp webview maintainable and simple, let's have a separated class for loading with http url. I have push [this](https://github.com/minh-rakuten/android-miniapp/commit/03e78f86f5130608ce4f94e210b47ae6f4beec72?branch=03e78f86f5130608ce4f94e210b47ae6f4beec72&diff=unified#diff-a8a5ca1f1ff16520549361cd01bf8d10fc2543d3a79f4f71075b7c2c19150a3cR96) based on your PR. Please feel free to consider it as suggestion or merge it. There are three file changes: [MiniAppWebView](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt), [MiniAppHttpWebView](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt), [RealMiniAppDisplay](https://github.com/minh-rakuten/android-miniapp/blob/03e78f86f5130608ce4f94e210b47ae6f4beec72/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt).

Thank you @minh-rakuten. It's very nice of you. Subclassing is definitely better idea. I'll add your changes to this PR

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/navigator/MiniAppExternalUrlLoader.kt
```diff
@@ -10,14 +10,25 @@ import com.rakuten.tech.mobile.miniapp.MiniAppScheme
  * @param miniAppId The id of loading mini app.
  * @param activity The Activity contains webview. Pass the activity if you want to auto finish
  * the Activity with current external loading url as result data.
+ * @param customMiniAppUrl The url that was used to load the Mini App.
  **/
-class MiniAppExternalUrlLoader(miniAppId: String, private val activity: Activity? = null) {
+class MiniAppExternalUrlLoader(
+    miniAppId: String,
+    private val activity: Activity? = null,
+    private val customMiniAppUrl: String? = null
```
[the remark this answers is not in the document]
imaginaris: Ok. Done. I'm not sure if it looks better now

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -24,6 +28,9 @@ internal class MiniAppDownloader(
     private val coroutineDispatcher: CoroutineDispatcher = Dispatchers.IO
 ) : UpdatableApiClient {
 
+    @Suppress("MagicNumber")
+    private val validHTTPReponseCodes = 100..399
```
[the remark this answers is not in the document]
imaginaris: 😮

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -37,6 +44,21 @@ internal class MiniAppDownloader(
...
+            val connection = URL(url).openConnection() as HttpURLConnection
+            val code = connection.responseCode
+            if (code !in validHTTPReponseCodes) {
+                throw MiniAppNotFoundException("Invalid URL error")
+            }
+        } catch (netError: ConnectException) {
+            throw MiniAppNotFoundException("Invalid URL error")
+        } catch (ioError: IOException) {
+            throw sdkExceptionForInternalServerError()
+        }
```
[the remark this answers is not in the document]
imaginaris: Probably. I'll put it just in case

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -73,6 +73,23 @@ internal class RealMiniApp(
...
 
+    override suspend fun createWithUrl(
+        appUrl: String,
+        miniAppMessageBridge: MiniAppMessageBridge,
+        miniAppNavigator: MiniAppNavigator?
+    ): MiniAppDisplay = when {
+        appUrl.isBlank() -> throw sdkExceptionForInvalidArguments()
+        else -> {
+            miniAppDownloader.validateMiniAppUrl(appUrl)
+            displayer.createMiniAppDisplay(
```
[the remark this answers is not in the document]
imaginaris: but that method is already throwing in case of error

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -29,11 +31,40 @@ internal class MiniAppWebView(
...
+            miniAppMessageBridge,
+            miniAppNavigator,
+            hostAppUserAgentInfo,
+            miniAppWebChromeClient,
+            miniAppCustomPermissionCache) {
+
+        this.miniAppUrl = miniAppUrl
+        this.miniAppTitle = miniAppTitle
+        miniAppScheme = MiniAppScheme.schemeWithCustomUrl(miniAppUrl)
+        miniAppId = "custom${Random.nextInt(0, Int.MAX_VALUE)}" // some id is needed to handle permissions
```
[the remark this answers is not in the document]
imaginaris: it's about handling permissions. empty id didn't work and url loaded miniapps don't have any metadata

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -68,17 +89,31 @@ internal class RealMiniAppDisplay(
...
+                    context = context,
+                    miniAppTitle = "Mini app",
```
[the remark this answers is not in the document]
imaginaris: Correct

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/Displayer.kt
```diff
@@ -24,4 +24,18 @@ internal class Displayer(private val context: Context, private val hostAppUserAg
         hostAppUserAgentInfo = hostAppUserAgentInfo,
         miniAppCustomPermissionCache = miniAppCustomPermissionCache
     )
+
+    fun createMiniAppDisplay(
+        miniAppUrl: String,
```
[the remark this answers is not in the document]
imaginaris: Ok. I changed all occurrences to `appUrl`.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayViewModel.kt
imaginaris: Oh I missed that one.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
```diff
@@ -0,0 +1,45 @@
...
+    MiniAppInfo.empty(),
+    miniAppMessageBridge,
+    miniAppNavigator,
+    hostAppUserAgentInfo,
+    miniAppWebChromeClient,
+    miniAppCustomPermissionCache
+) {
+    init {
+        miniAppScheme = MiniAppScheme.schemeWithCustomUrl(appUrl)
+        miniAppId = "custom${Random.nextInt(0, Int.MAX_VALUE)}" // some id is needed to handle permissions
```
khairul-alam-licon, line 32: Could you help us to understand what kind of permissions need to handle?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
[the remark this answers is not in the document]
khairul-alam-licon: Yeah, in this case we can rely on current implementation. Currently `createWithUrl` returns non-null `MiniAppDisplay`, so we need to do some refactoring to adopt my previous comment. But, I think for now you can ignore this suggestion.

miniapp/src/main/res/xml/network_security_config.xml
corycaywood: I think we shouldn't set the `network-security-config` on the SDK side. We should require the Host App to control this.

So probably it's better to move this file to `testapp`, and then we can set it to allow clear text for all domains. The demo app is only used for testing, so I think this should be okay.

@minh-rakuten @imaginaris Do you think that makes sense?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
corycaywood: @khairul-alam-licon I think this is so the custom permissions will be cached to a random ID. Is that right @imaginaris ?

We don't need the permissions to be cached for URLs. But I think this solution is okay because each time a URL is launched it will be assigned a different ID to use for caching. Trying to disable caching would probably need some more major refactoring. 

But a disadvantage is that we end up with a lot of permissions cached for random IDs which we don't need. Is it possible to delete these from the cache when the mini app is closed? For example something like this:

```
    override fun onDetachedFromWindow() {
        super.onDetachedFromWindow()
        
        miniAppCustomPermissionCache.removeId(miniAppId) // you would need to add a method like this
    }
```

miniapp/src/main/res/xml/network_security_config.xml
minh-rakuten: If we want to make it available for every domain then I think using `android:usesCleartextTraffic="true"` in `testapp` is enough.

testapp/src/main/res/values/strings.xml
```diff
@@ -21,7 +22,7 @@
     <string name="userdata_warning_empty_contact_list">No contact available!</string>
 
     <string name="action_display">Display</string>
-    <string name="action_go_input">Display Miniapp From ID</string>
+    <string name="action_go_input">Display Miniapp From ID or URL</string>
```
corycaywood, line 25: It's a little confusing that this is only available when "Preview mode" is turned off. Could we also display this button while the Demo App has "Preview mode" enabled but then change the text to `Display Miniapp from URL` and hide the "App ID" field and button on `mini_app_input_activity.xml`?

This would be a little more complex to do. What do you think?

@minh-rakuten @khairul-alam-licon Any other suggestions for how to handle this?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
minh-rakuten: @corycaywood Loading by url does not proceed with `MiniAppDownloader` and `MiniAppStatus` does not cache anything.
We need a dummy miniapp id to ensure other parts are working properly (such as retrieve access token).
-> @imaginaris  solution is okay.

khairul-alam-licon: I have another concern when users launch a same url twice e.g. `https://www.google.com` it will produce different ids for the same url. Should we care about it? 

Also, if we need to generate this random id, so we could probably match with the format what backend provides.

e.g. `miniAppId = UUID.randomUUID().toString()`

testapp/src/main/res/values/strings.xml
khairul-alam-licon: if loading a MiniApp by url has no dependency with preview/published mode, then we can use a separate place for this option. I am not good at suggesting UX design, but the simplest thing for now I can say is adding a separate button on bottom of this UI.

minh-rakuten: Yeah I intended that we will go that way in separated PR. It's okay to implement it in this PR too.
I think the label can be changed into `Display Miniapp by Input`, it would go well with both cases.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
minh-rakuten:
```suggestion
        miniAppId = UUID.fromString(appUrl).toString()
```

miniapp/src/main/res/xml/network_security_config.xml
imaginaris: Got it!

testapp/src/main/res/values/strings.xml
imaginaris: I think I'll go with @corycaywood 's solution. Should be fairly easy to implement

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
khairul-alam-licon: Ah, I just realized when I try a normal url e.g. `https://www.google.com/` using `UUID.fromString(appUrl).toString()`, it returns the following error:

```
java.lang.IllegalArgumentException: Invalid UUID string: https://www.google.com/
```
So, maybe we can attempt with `UUID.randomUUID().toString()` instead.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppHttpWebView.kt
```diff
@@ -0,0 +1,45 @@
...
+    MiniAppInfo.empty(),
+    miniAppMessageBridge,
+    miniAppNavigator,
+    hostAppUserAgentInfo,
+    miniAppWebChromeClient,
+    miniAppCustomPermissionCache
+) {
+    init {
+        miniAppScheme = MiniAppScheme.schemeWithCustomUrl(appUrl)
+        miniAppId = UUID.fromString(appUrl).toString() // some id is needed to handle custom permissions
```
minh-rakuten, line 32: oh I'm sorry. `UUID.fromString(appUrl).toString()` isn't working as I expected 🙇  so we can keep your old code `"custom${Random.nextInt(0, Int.MAX_VALUE)}"` or `UUID.randomUUID().toString()`.

khairul-alam-licon: `UUID.randomUUID().toString()` +1

imaginaris: Ok. Fixed

minh-rakuten (commented):
When this [suggestion](https://github.com/rakutentech/android-miniapp/pull/211#discussion_r533095159) would be resolved, I think it's ready for unit test.

khairul-alam-licon (approved):
Looks good to me.
It'll be nice while working on unit test and the related documentation of this new feature 🎉 
Thank you for your work 👍

imaginaris:
I implemented solution suggested by @corycaywood 
Let me know if there is still anything missing.

corycaywood (approved):
Nice, looks good! 👍

minh-rakuten (approved):
👍 👍

PR #216, opened 2020-12-03
refactor: custom permission sort

minh-rakuten:
# Description
Refactor the sort order of custom permission.

## Links
MINI-2188. This ticket had been resolved but the fact that the problem has not been observed properly.

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
@@ -96,6 +96,11 @@ internal class MiniAppCustomPermissionCache(context: Context) {
         applyStoringPermissions(MiniAppCustomPermission(miniAppId, allPermissions))
     }
 
+    fun removeId(miniAppId: String) {
+        prefs.edit().remove(miniAppId).apply()
+    }
+
+    @SuppressWarnings("PrintStackTrace")
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun applyStoringPermissions(miniAppCustomPermission: MiniAppCustomPermission) {
         val jsonToStore: String = Gson().toJson(sortedByDefault(miniAppCustomPermission))
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
@@ -96,44 +96,17 @@ internal class MiniAppCustomPermissionCache(context: Context) {
         applyStoringPermissions(MiniAppCustomPermission(miniAppId, allPermissions))
     }
 
-    @SuppressWarnings("PrintStackTrace")
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun applyStoringPermissions(miniAppCustomPermission: MiniAppCustomPermission) {
-        try {
-            val jsonToStore: String = Gson().toJson(orderByDefaultList(miniAppCustomPermission))
-            prefs.edit().putString(miniAppCustomPermission.miniAppId, jsonToStore).apply()
-        } catch (e: Exception) {
-            e.printStackTrace()
-        }
+        val jsonToStore: String = Gson().toJson(sortedByDefault(miniAppCustomPermission))
+        prefs.edit().putString(miniAppCustomPermission.miniAppId, jsonToStore).apply()
     }
 
-    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
-    fun orderByDefaultList(miniAppCustomPermission: MiniAppCustomPermission): MiniAppCustomPermission {
-        val miniAppId = miniAppCustomPermission.miniAppId
-        val defaultTypesOrder = mutableListOf<MiniAppCustomPermissionType>()
-        defaultDeniedList(miniAppId).pairValues.forEach {
-            defaultTypesOrder.add(it.first)
-        }
-
-        val currentTypesOrder = mutableListOf<MiniAppCustomPermissionType>()
-        miniAppCustomPermission.pairValues.forEach {
-            currentTypesOrder.add(it.first)
-        }
-
-        val expectedTypesOrder = currentTypesOrder.map {
-            defaultTypesOrder.indexOf(it)
-        }.sorted().map { value -> defaultTypesOrder[value] }
-
-        val orderedPair =
-            mutableListOf<Pair<MiniAppCustomPermissionType, MiniAppCustomPermissionResult>>()
-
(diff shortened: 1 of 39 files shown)


khairul-alam-licon:
> MINI-2188. This ticket had been resolved but the fact that the problem has not been observed properly.

What is the other problem?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -96,44 +96,17 @@ internal class MiniAppCustomPermissionCache(context: Context) {
         applyStoringPermissions(MiniAppCustomPermission(miniAppId, allPermissions))
     }
 
-    @SuppressWarnings("PrintStackTrace")
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun applyStoringPermissions(miniAppCustomPermission: MiniAppCustomPermission) {
-        try {
-            val jsonToStore: String = Gson().toJson(orderByDefaultList(miniAppCustomPermission))
-            prefs.edit().putString(miniAppCustomPermission.miniAppId, jsonToStore).apply()
-        } catch (e: Exception) {
```
khairul-alam-licon, line 105: Is there any reason to remove this exception?

minh-rakuten:
> > MINI-2188. This ticket had been resolved but the fact that the problem has not been observed properly.
> 
> What is the other problem?

MINI-2188 has not been fixed. The ticket was resolved as "Won't Fix"

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
minh-rakuten: Actually, why there is the try catch exception here?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -96,44 +96,17 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
-            e.printStackTrace()
-        }
+        val jsonToStore: String = Gson().toJson(sortedByDefault(miniAppCustomPermission))
+        prefs.edit().putString(miniAppCustomPermission.miniAppId, jsonToStore).apply()
     }
 
-    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
khairul-alam-licon, line 110: Why removing this line?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -96,44 +96,17 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
-            miniAppCustomPermission.pairValues.find {
-                it.first == type
-            }?.let { orderedPair.add(index, it) }
-        }
-
-        return MiniAppCustomPermission(miniAppId, orderedPair)
+    // Sort the `pairValues` by ordinal of [MiniAppCustomPermissionType].
+    @VisibleForTesting
+    internal fun sortedByDefault(miniAppCustomPermission: MiniAppCustomPermission): MiniAppCustomPermission {
+        val sortedPairValues = miniAppCustomPermission.pairValues.sortedBy { it.first.ordinal }
```
khairul-alam-licon, line 108: In previous implementation, I wanted to use the ordering from `defaultDeniedList`, the expected order would be inside this class. But, using `MiniAppCustomPermissionType` is better as it removes some boilerplate code 👍 As well as the related unit test has been passed.

P/S: I don't think, we need to use `internal` here, this entire class is already `internal`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
khairul-alam-licon: I think, it's good to have an exception catchup as there could be a case as documentation. It's not obvious but code quality matters.

Screen Shot 2020-12-03 at 18 28 23

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
minh-rakuten: because you're using it [here](https://github.com/rakutentech/android-miniapp/blob/11bcfa688a5a137a79c1801f73c74419806ca480/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt#L198).

khairul-alam-licon:
> MINI-2188. This ticket had been resolved but the fact that the problem has not been observed properly.

> What is the other problem?

> @khairul-alam-licon MINI-2188 has not been fixed. The ticket was resolved as "Won't Fix"

Yes, the ticket was issued before merging Location custom permission. The ordering was already fixed in Location PR. So, this PR is just refactoring this logic. Sorry, I asked because the PR's description was confusing for me.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
minh-rakuten: I think if there is error on json parsing, it should be addressed because there must be no error here.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -96,44 +96,17 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
-
-        expectedTypesOrder.forEachIndexed { index, type ->
-            miniAppCustomPermission.pairValues.find {
-                it.first == type
-            }?.let { orderedPair.add(index, it) }
-        }
-
-        return MiniAppCustomPermission(miniAppId, orderedPair)
+    // Sort the `pairValues` by ordinal of [MiniAppCustomPermissionType].
+    @VisibleForTesting
```
khairul-alam-licon, line 106: `sortedByDefault` function (formerly, `orderByDefaultList`) is using only in `MiniAppCustomPermissionCache`. So, we can make just private. But, to write unit test, we need to use `@VisibleForTesting`.  And, I have used `@VisibleForTesting(otherwise = VisibleForTesting.NONE)` previously to restrict it only for unit test. 

Screen Shot 2020-12-03 at 18 45 43

khairul-alam-licon (approved):
👍 👍

corycaywood (approved):
LGTM 👍

PR #218, opened 2020-12-08
move app name from sdk to hostapp

khairul-alam-licon:
# Description
- Remove "app_name" resource from `miniapp/src/main/res/values/strings.xml`.

## Links
- MINI-2227

# Checklist

### DIFF
--- a/testapp/build.gradle
+++ b/testapp/build.gradle
@@ -32,14 +32,24 @@ android {
         }
     }
 
-    def appName = "Mini App Sample"
+    def defaultAppName = "Mini App Sample"
+    def debugManifestPlaceholders = [
+            baseUrl                 : property("MINIAPP_SERVER_BASE_URL") ?: "https://www.example.com/",
+            isPreviewMode           : property("IS_PREVIEW_MODE") ?: true,
+            hostAppUserAgentInfo    : property("HOST_APP_UA_INFO") ?: "MiniApp Demo App/$project.version",
+            projectId               : property("HOST_PROJECT_ID") ?: "test-host-project-id",
+            subscriptionKey         : property("HOST_APP_SUBSCRIPTION_KEY") ?: "test-subs-key",
+            adMobAppId              : property("ADMOB_APP_ID") ?: "",
+            appName                 : defaultAppName + " DEBUG"
+    ]
     def stagingManifestPlaceholders = [
             baseUrl                 : property("MINIAPP_SERVER_BASE_URL") ?: "https://www.example.com/",
             isPreviewMode           : property("IS_PREVIEW_MODE") ?: true,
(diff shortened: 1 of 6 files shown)


testapp/src/main/AndroidManifest.xml
```diff
@@ -10,7 +10,7 @@
         android:name="com.rakuten.tech.mobile.testapp.SampleApplication"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
-        android:label="@string/app_name"
+        android:label="@string/hostapp_name"
```
corycaywood, line 13: But isn't this going to break the title on the different build variants? For example, all of them will be named "Mini App Sample" instead of "Mini App Sample DEBUG" and "Mini App Sample STG". See https://github.com/rakutentech/android-miniapp/blob/master/testapp/build.gradle#L59

Actually, couldn't you just completely delete `app_name` (and `hostapp_name`) from `strings.xml` because it's being set in `build.gradle`?

khairul-alam-licon: Yeah, I agree. I have figured out something from https://developer.android.com/studio/build/gradle-tips that we can utilize `manifestPlaceholders`. I have updated with it.

testapp/build.gradle
```diff
@@ -32,22 +32,33 @@ android {
...
+    ]
     def stagingManifestPlaceholders = [
             baseUrl                 : property("MINIAPP_SERVER_BASE_URL") ?: "https://www.example.com/",
             isPreviewMode           : property("IS_PREVIEW_MODE") ?: true,
             hostAppUserAgentInfo    : property("HOST_APP_UA_INFO") ?: "MiniApp Demo App/$project.version",
             projectId               : property("HOST_PROJECT_ID") ?: "test-host-project-id",
             subscriptionKey         : property("HOST_APP_SUBSCRIPTION_KEY") ?: "test-subs-key",
-            adMobAppId              : property("ADMOB_APP_ID") ?: ""
+            adMobAppId              : property("ADMOB_APP_ID") ?: "",
+            appName                 : defaultAppName + " STG"
```
corycaywood, line 52: Looks like `debugManifestPlaceholders` and `stagingManifestPlaceholders` are exactly the same except for the `appName`. So could you just clone it instead of duplicating the properties?

```
def stagingManifestPlaceholders = debugManifestPlaceholders.clone()
stagingManifestPlaceholders.appName = defaultAppName + " STG"
```

corycaywood (approved):
Looks good 👍  Just left one suggestion.

testapp/build.gradle
khairul-alam-licon: Thanks, nice idea, I have updated it.

minh-rakuten (approved):
👍 👍

PR #219, opened 2020-12-09
fix: disable google backup for shared preferences

khairul-alam-licon:
# Description
`android:allowBackup="true"` is using for Google backup which can be seen in Google drive. Data of AppSettings is storing in Shared preferences which is [one of the files to backup](https://developer.android.com/guide/topics/data/autobackup.html#Files). There is might unexpected behavior when we change these settings in new releases of the Demo App. In this case, we can restrict the rule for Shared preferences not to do auto-backup in Google drive. This PR aims to bring a solution for this. 

However, I was checking [the document ](https://developer.android.com/guide/topics/data/autobackup#XMLSyntax)if there is something to specify the data from AppSettings, I think, they only support files/folders to exclude. 

```
<exclude> - Specifies a file or folder to exclude during backup. Here are some files that are typically excluded from backup:
```

So, I have excluded all shared preferences data in the rule.

## Links
- MINI-2258

# Checklist

### DIFF
--- a/testapp/src/main/res/xml/data_backup_rules.xml
+++ b/testapp/src/main/res/xml/data_backup_rules.xml
@@ -1,4 +1,0 @@
-<?xml version="1.0" encoding="utf-8"?>
-<full-backup-content>
-  <exclude domain="sharedpref" path="."/>
-</full-backup-content>
--- a/testapp/src/main/res/xml/data_backup_rules.xml
+++ b/testapp/src/main/res/xml/data_backup_rules.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+  <exclude domain="sharedpref" path="."/>
+</full-backup-content>
--- a/testapp/src/main/AndroidManifest.xml
+++ b/testapp/src/main/AndroidManifest.xml
@@ -8,8 +8,7 @@
 
     <application
         android:name="com.rakuten.tech.mobile.testapp.SampleApplication"
-        android:allowBackup="true"
-        android:fullBackupContent="@xml/data_backup_rules"
+        android:allowBackup="false"
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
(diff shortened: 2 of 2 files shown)


testapp/src/main/res/xml/data_backup_rules.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
+  <exclude domain="sharedpref" path="."/>
```
corycaywood, line 3: Excluding all shared preferences will also exclude the settings from the SDK side - like caching which mini apps have already been downloaded. So in that case, the `files` folder which includes the mini app assets would still be backed up to google cloud, but the shared preferences setting which caches that the mini app has been downloaded would not be backed up. So then the SDK would re-download the mini app, even though it's already been downloaded.

Could you specify just the file used by the sample app for the `path' here? i.e. `com.rakuten.tech.mobile.miniapp.sample.settings.xml`

Otherwise, we should just completely disable Google Backup for the App in order to avoid strange behavior like mentioned above.

khairul-alam-licon: Yeah, I think `com.rakuten.tech.mobile.miniapp.sample.settings.xml` will work. However, I think, there is no good way to explore which data has been backed up, I was trying to find the answer in https://support.google.com/drive/answer/6305834.
Looks like the backup data will expire after a certain period, it depends on user when they like to backup their data. It means, from our side, it's challenging to rely on this backup data even if we want to use it in future. 

So, I am thinking if we should completely disable the Google backup for now.

corycaywood: Okay, I think probably it's fine just to completely disable Google backup.

khairul-alam-licon: Yeah, I have updated this PR.

corycaywood (approved):
LGTM 👍

minh-rakuten (approved):
The problem of ticket MINI-2258 is not obvious to us. I think we should take a deeper look to detect the root cause.

corycaywood:
> The problem of ticket MINI-2258 is not obvious to us. I think we should take a deeper look to detect the root cause.

Yeah, I'm not really sure either. I only know that the latest demo app didn't work for some people if they had google backup enabled.

It will be useful once we integrate crash reporting because then we can investigate things like this better.

PR #221, opened 2020-12-11
fix: retrieve location correctly in geo location flow

khairul-alam-licon:
# Description
Fix issue when location was showing during custom permission is denied but device permission is allowed. It seems that `onGeolocationPermissionsShowPrompt` has an intention to check permission state for `android.permission.ACCESS_FINE_LOCATION`, so we should add a barricade of custom permission while invoking the permission which affects Geolocation API. 

Initially, I intended to show dialogs for both device and custom permission before executing `GeolocationPermissions.Callback?.invoke(origin, true, false)`, but it seems the checkup of JS SDK is already enough. So, in Android side, I have used the value of custom permission under `onGeolocationPermissionsShowPrompt` at runtime.

## Links
- MINI-2314

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
@@ -2,7 +2,6 @@ package com.rakuten.tech.mobile.miniapp.display
 
 import android.app.Activity
 import android.content.Context
-import android.content.pm.PackageManager
 import android.view.View
 import android.view.ViewGroup
 import android.webkit.GeolocationPermissions
@@ -12,7 +11,6 @@ import android.webkit.WebChromeClient
 import android.webkit.WebView
 import android.widget.FrameLayout
 import androidx.annotation.VisibleForTesting
-import androidx.core.app.ActivityCompat
 import com.rakuten.tech.mobile.miniapp.MiniAppInfo
 import com.rakuten.tech.mobile.miniapp.js.DialogType
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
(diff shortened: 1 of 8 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -38,7 +45,19 @@ internal class MiniAppWebChromeClient(
...
     ) {
-        callback?.invoke(origin, true, false)
+        val hasCustomPermission =
+            miniAppCustomPermissionCache.readPermissions(miniAppInfo.id)
+                .pairValues.find {
+                    it.first == MiniAppCustomPermissionType.LOCATION
+                }?.let {
+                    it.second == MiniAppCustomPermissionResult.ALLOWED
+                }
+        val hasDevicePermission = ActivityCompat.checkSelfPermission(
```
corycaywood, line 55: Do we need to have this check for the device permission? I think this is already handled by the android system. Device permissions was already working correctly without this.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -38,7 +45,19 @@ internal class MiniAppWebChromeClient(
         origin: String?,
         callback: GeolocationPermissions.Callback?
     ) {
-        callback?.invoke(origin, true, false)
+        val hasCustomPermission =
+            miniAppCustomPermissionCache.readPermissions(miniAppInfo.id)
```
corycaywood, line 49: It seems like we're duplicating code in several places for checking if a permission is granted (for example in `CustomPermissionBridgeDispatcher`, in `UserInfoBridgeDispatcher`, and right here).

Maybe we should add something like `isGranted(permission: MiniAppCustomPermissionType)` to `MiniAppCustomPermissionCache`? I guess this could be a refactoring improvement in another PR.

corycaywood (commented):
Looking good 👍  See my comment below about the native permission.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
khairul-alam-licon: Thanks for the concern, we can skip this also. Initially, I thought we should check this but it's not obvious in this situation, so I'll update this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
khairul-alam-licon: Right, I have added something like `hasPermission(...)` to `MiniAppCustomPermissionCache` for this cleanup.
It can be seen in https://github.com/rakutentech/android-miniapp/pull/222

corycaywood (approved):
Looks good! 👍

minh-rakuten (approved):
👍 👍

PR #223, opened 2020-12-16
feat: allow profile photo and user name to be retrieved asynchronously

khairul-alam-licon:
# Description
Change `MiniAppMessageBridge.getProfilePhoto` and `MiniAppMessageBridge.getUserName` to use a callback so that the photo and name can be retrieved/generated asynchronously. Deprecate the old synchronous versions of these methods.

## Links
- MINI-2120

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -266,36 +266,38 @@ val miniAppMessageBridge = object: MiniAppMessageBridge() {
 val userInfoBridgeDispatcher = object : UserInfoBridgeDispatcher() {
 
     override fun getUserName(
-        onSuccess: (name: String) -> Unit,
-        onError: (message: String) -> Unit
+        callback: (result: Result<String?>) -> Unit
     ) {
         val name: String = ""
         // Check if there is any valid username in HostApp
         // .. .. ..
-        if (name is valid)
-            onSuccess(name) // allow miniapp to get the name.
-        else
-            onError(message) // reject miniapp to get the name with message explanation.
+        val result: Result<String> = run {
+            if (name is valid) success(name) // allow miniapp to get the name.
+            else failure(Exception("User name is not found.")) // reject miniapp to get the name with message explanation.
+        }
+
+        callback.invoke(result) // invoke the Result callback
     }
 
     override fun getProfilePhoto(
-        onSuccess: (photoUrl: String) -> Unit,
-        onError: (message: String) -> Unit
+            callback: (result: Result<String?>) -> Unit
     ) {
         val photoUrl: String = ""
         // Check if there is any valid photo url in HostApp
         // .. .. ..
-        if (photoUrl is valid)
-            onSuccess(photoUrl) // allow miniapp to get the photo url.
-        else
-            onError(message) // reject miniapp to get the photo url with message explanation.
+        val result: Result<String> = run {
+            if (photoUrl is valid) success(name) // allow miniapp to get the photo url.
+            else failure(Exception("User name is not found.")) // reject miniapp to get the photo url with message explanation.
+        }
+
+        callback.invoke(result) // invoke the Result callback
     }
 
     override fun getAccessToken(
         miniAppId: String,
         onSuccess: (tokenData: TokenData) -> Unit,
         onError: (message: String) -> Unit
-    ){
+    ) {
         var allowToken: Boolean = false
         // Check if you want to allow this Mini App ID to use the Access Token
         // .. .. ..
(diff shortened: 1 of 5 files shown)


miniapp/USERGUIDE.md
```diff
@@ -264,20 +264,49 @@ val miniAppMessageBridge = object: MiniAppMessageBridge() {
...
 
 val userInfoBridgeDispatcher = object : UserInfoBridgeDispatcher() {
+
+    // note: it has been deprecated now.
     override fun getUserName(): String {
         var name: String = ""
         // Implementation details to get user name
         // .. .. ..
         return name
     }
```
minh-rakuten, line 274: It's okay to completely remove the deprecated example.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridgeDispatcher.kt
```diff
@@ -22,16 +22,46 @@ abstract class UserInfoBridgeDispatcher {
...
     open fun getUserName(): String =
         throw MiniAppSdkException("The `UserInfoBridgeDispatcher.getUserName` $NO_IMPL")
 
+    /**
+     * Get user name from host app.
+     * You can also throw an [Exception] from this method to pass an error message to the mini app.
+     */
+    open fun getUserName(
+        onSuccess: (name: String) -> Unit,
+        onError: (message: String) -> Unit
```
minh-rakuten, line 38: We can use a single callback with [Result](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/).

minh-rakuten (commented):
It's good but we can improve it by using single callback. We had a [discussion](https://github.com/rakutentech/android-miniapp/pull/166#discussion_r501401190) about this. I think this scenario can apply that suggestion.

corycaywood (approved):
Looks good 👍  I left a suggestion below.

minh-rakuten (approved):
LGTM 👍 
Only unit test left.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridgeDispatcher.kt
```diff
@@ -22,16 +22,44 @@ abstract class UserInfoBridgeDispatcher {
...
+    )
     open fun getProfilePhoto(): String =
         throw MiniAppSdkException("The `UserInfoBridgeDispatcher.getProfilePhoto` $NO_IMPL")
 
+    /**
+     * Get profile photo url from host app.
+     * You can send profile photo url or throw an [Exception] from this method by passing a [Result].
+     */
+    open fun getProfilePhoto(
+        callback: (result: Result<String?>) -> Unit
```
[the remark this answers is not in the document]
khairul-alam-licon: Yeah, we can do that. But it seems like we are intending to add many deprecated functions at a once. I think, if we write this asynchronous username and profile photo function without using `Result` to keep consistency with others, then it will be user friendly. 
Then, we can add `Result` in a separate ticket, of course it will be nice if we can completely remove the deprecated functions meanwhile. 

@corycaywood @minh-rakuten What do you think?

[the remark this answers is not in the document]
khairul-alam-licon: > Then, we can add Result in a separate ticket, of course it will be nice if we can completely remove the deprecated functions meanwhile.

Or, I can address it in this same PR as @corycaywood suggested before. So, I have updated this PR.

I want to address `shareContent` with using `Result` in a separate PR because it has default implementation. So, need to think differently.

khairul-alam-licon:
I haven't written duplicate test lines for both old and new functions. So, it affects the coverage, but after removing the deprecated functions hopefully the coverage will be improved.

miniapp/USERGUIDE.md
```diff
@@ -264,44 +264,57 @@ val miniAppMessageBridge = object: MiniAppMessageBridge() {
...
+
+    override fun getUserName(
+        onSuccess: (userName: String) -> Unit,
+        onError: (message: String) -> Unit
+    ) {
+        val name: String = ""
+        // Check if there is any valid username in HostApp
         // .. .. ..
-        return name
+        if (name is valid)
```
corycaywood, line 275: It seems like this could be confusing to developers because `is` is a valid operator in Kotlin but it wouldn't be used in a context like this. So the code would be broken if they just copied and pasted it like this. Maybe we can just make `name` be nullable here and add a comment?
```
val name: String?
 // Check if there is any valid username in HostApp
// .. .. ..
if (name) // Check if name is valid
```

corycaywood (approved):
LGTM 👍  Just a suggestion on the user guide.

miniapp/USERGUIDE.md
khairul-alam-licon: Ah right, I didn't notice like this way. I have updated this part.

minh-rakuten (approved):
👍 👍

PR #227, opened 2020-12-18
docs: Add Java example to user guide

corycaywood:
# Description
Added example for using in a Java project. Also added info that we don't specifically support usage of the SDK in Java.

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -612,6 +612,80 @@ dependency {
 ```
 </details>
 
+<details><summary markdown="span"><b>How can I use this SDK in a Java project?</b>
+</summary>
+
+We don't support usage of the Mini App SDK in a Java project. This is because this SDK uses Kotlin specific features such as `suspend` functions and `Coroutines`.
+
+However, it is possible to use this SDK in a Java project if you wrap the calls to `suspend` functions into something that Java can use such as a `Future` or a normal callback interface.
+
+To do this, you will need to integrate Kotlin and Kotlin Coroutines into your project:
+
+```groovy
+// In your top level "build.gradle" file:
+buildscript {
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.50"
+    }
+}
+
+// In your project "myApp/build.gradle" file
+apply plugin: 'kotlin-android'
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.50"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.3.0" // Needed if you want to use CoroutineScope.future
+}
+```
+
+Then, you will need to create wrapper functions for the Mini App SDK functionality which you wish to use.
+
+If your minimum Android API level is 24 or higher, then you can use Java [`CompletableFuture`](https://developer.android.com/reference/java/util/concurrent/CompletableFuture):
+
+```kotlin
+// MiniAppAsync.kt
+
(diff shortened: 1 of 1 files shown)


miniapp/USERGUIDE.md
```diff
@@ -612,6 +612,80 @@ dependency {
...
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.50"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.3.0"
+}
+```
+
+Then, you will need to create wrapper functions for the Mini App SDK functionality which you wish to use.
+
+If your minimum SDK level is 24 or higher, then you can use Java `CompletableFuture`:
```
khairul-alam-licon, line 643: Also, I think we can link https://developer.android.com/reference/java/util/concurrent/CompletableFuture here.

miniapp/USERGUIDE.md
```diff
@@ -612,6 +612,80 @@ dependency {
...
+fun createMiniAppAsync(
+    appId: String,
+    miniAppMessageBridge: MiniAppMessageBridge,
+    miniAppNavigator: MiniAppNavigator?
+): CompletableFuture<MiniAppDisplay> = coroutineScope.future {
+    MiniApp.instance().create(appId, miniAppMessageBridge, miniAppNavigator)
+}
+```
+
+If your minimum SDK level is lower than 24, then you can use a normal Java callback interface:
```
khairul-alam-licon, line 659: I am not sure, if it's good to mention "Android API level" instead of "SDK level".

miniapp/USERGUIDE.md
```diff
@@ -612,6 +612,80 @@ dependency {
...
+    dependencies {
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:1.3.50"
+    }
+}
+
+// In your project "myApp/build.gradle" file
+apply plugin: 'kotlin-android'
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:1.3.50"
```
khairul-alam-licon, line 636: Looks like if the hostapps do `implementation project` of miniapp SDK, then they can get this library because the same dependency has been declared in miniapp SDK. But, still we can just mention it as it's required to integrate Kotlin.

miniapp/USERGUIDE.md
```diff
@@ -612,6 +612,80 @@ dependency {
...
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-jdk8:1.3.0"
```
khairul-alam-licon, line 637: Maybe we can put a comment above of this line that we need it to adopt [CoroutineScope.future](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-jdk8/kotlinx.coroutines.future/kotlinx.coroutines.-coroutine-scope/future.html)?

khairul-alam-licon (approved):
Looks good, Thank you for this guideline 👍 
I have left few minor comments.

minh-rakuten (approved):

miniapp/USERGUIDE.md
corycaywood: We have this set as an `implementation` dependency in our SDK rather than an `api` dependency, so the Host App wouldn't get this dependency on their own compile classpath. 

Anyways, I think that it's a good practice for Apps to directly have dependencies declared which they use in their own code rather than relying on a transitive dependency. For example, the transitive dependency could be removed and then their own code would be broken (although that's really unlikely to happen in the case of the kotlin-stdlb dependency).

PR #231, opened 2020-12-22
fix: proguard check MiniAppWebView

minh-rakuten:
# Description
Fix miniapp view init due to R8 obfuscation.

## Links
MINI-2445

# Checklist
- [] I wrote/updated tests for new/changed code

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
@@ -51,7 +51,7 @@ internal open class MiniAppWebView(
     }
 
     init {
-        if (this::class.java.simpleName == "MiniAppWebView")
+        if (this::class.java == MiniAppWebView::class.java)
             commonInit()
     }
 
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
@@ -51,7 +51,7 @@ internal open class MiniAppWebView(
     }
 
     init {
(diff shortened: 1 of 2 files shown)


khairul-alam-licon (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -51,7 +51,7 @@ internal open class MiniAppWebView(
     }
 
     init {
-        if (this::class.java.simpleName == "MiniAppWebView")
+        if (this::class.java == MiniAppWebView::class.java)
```
corycaywood, line 54: Shouldn't this also work with `this::class == MiniAppWebView::class`? But I'm not sure if there's any advantage of using `::class` or `::class.java` 🤔  

If it's working with `::class.java` then probably it's okay.

corycaywood (approved):
Looks good 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
minh-rakuten: I have updated to `::class`.

khairul-alam-licon:
```
Output of `this::class` or `this::class.java`:

without obfuscation: `com.rakuten.tech.mobile.miniapp.display.MiniAppWebView`
with obfuscation: `com.rakuten.tech.mobile.miniapp.display.d`
```
Previously, it causes the issue because, output of `this::class.java.simpleName` was "d" with obfuscation.

However, let's merge this as a patch.  My suggestion is doing something else while comparing these values, so maybe we can improve this later.

khairul-alam-licon (approved):

PR #232, opened 2020-12-22
feat: pass parameters to a mini app

khairul-alam-licon:
# Description
- Allow the Host App to pass a String parameters to a mini app when calling `MiniApp.create`. The parameters should then be appended to the mini app's URL as query params. For example: `https://mscheme.1234/miniapp/index.html?param1=value1&param2=value2`.
- Update the Demo App settings page with a new setting for the mini app parameters which will be passed to the mini app. 

## Links
- MINI-2284

# Checklist

### DIFF
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -576,6 +576,26 @@ downloadedMiniApps.forEach {
 }
 ```
 
+### Passing parameters to a miniapp
+
+For a mini app, you can pass query parameters as String using `MiniApp.create` to be appended with miniapp's url.
+For example: `https://mscheme.1234/miniapp/index.html?param1=value1&param2=value2`
+
+```kotlin
+class MiniAppActivity : Activity(), CoroutineScope {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+    //...
+        launch {
+            val miniAppDisplay = withContext(Dispatchers.IO) {
+                MiniApp.instance().create("mini_app_id", "param1=value1&param2=value2", miniAppMessageBridge)
+            }
+    //...
+        }
+    }
+}
+```
+
 ## Troubleshooting & FAQs
 
 <details><summary markdown="span"><b>Exception: "Network requests must not be performed on the main thread."</b>
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -601,7 +601,7 @@ class MiniAppActivity : Activity(), CoroutineScope {
     //...
         launch {
             val miniAppDisplay = withContext(Dispatchers.IO) {
-                MiniApp.instance().create("mini_app_id", "param1=value1&param2=value2", miniAppMessageBridge)
+                MiniApp.instance().create("mini_app_id", miniAppMessageBridge, "param1=value1&param2=value2")
             }
     //...
         }
--- a/miniapp/USERGUIDE.md
+++ b/miniapp/USERGUIDE.md
@@ -601,7 +601,12 @@ class MiniAppActivity : Activity(), CoroutineScope {
     //...
         launch {
             val miniAppDisplay = withContext(Dispatchers.IO) {
-                MiniApp.instance().create("mini_app_id", miniAppMessageBridge, "param1=value1&param2=value2")
+                MiniApp.instance().create(
+                    appId = "mini_app_id",
+                    miniAppMessageBridge = miniAppMessageBridge,
+                    miniAppNavigator = miniAppNavigator,
+                    queryParams = "param1=value1&param2=value2"
+                )
             }
     //...
         }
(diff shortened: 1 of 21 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -45,6 +46,7 @@ abstract class MiniApp internal constructor() {
     @Throws(MiniAppNotFoundException::class, MiniAppHasNoPublishedVersionException::class, MiniAppSdkException::class)
     abstract suspend fun create(
         appId: String,
+        queryParams: String,
```
minh-rakuten, line 49: This `queryParams` can have default value such as empty string so there will be no impact for current function implementation.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppScheme.kt
```diff
@@ -35,6 +36,18 @@ internal class MiniAppScheme private constructor(miniAppId: String) {
...
 
+    fun appendParametersToUrl(url: String, queryParams: String): String {
+        return if (queryParams.isEmpty()) url
+        else "$url${resolveParameters(queryParams)}"
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun resolveParameters(queryParams: String): String {
+        val qusMark = "?"
+        return if (queryParams.contains(qusMark) && queryParams.startsWith(qusMark)) queryParams
```
minh-rakuten, line 47: I think that our spec defines the param as this format: `param=value&param2=value` so we don't have to check the `?`.

minh-rakuten (commented):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppScheme.kt
khairul-alam-licon: Checking qus mark is a safe call here. At-least we need to filter the qus mark here. Additionally, I was thinking if we also need to verify something else if the params are valid or not.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
khairul-alam-licon: Yeah, I have updated this part.

miniapp/USERGUIDE.md
```diff
@@ -589,6 +589,26 @@ downloadedMiniApps.forEach {
...
+For example: `https://mscheme.1234/miniapp/index.html?param1=value1&param2=value2`
+
+```kotlin
+class MiniAppActivity : Activity(), CoroutineScope {
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+    //...
+        launch {
+            val miniAppDisplay = withContext(Dispatchers.IO) {
+                MiniApp.instance().create("mini_app_id", miniAppMessageBridge, "param1=value1&param2=value2")
```
corycaywood, line 604: Is this example correct? I think that this won't work because the third parameter is `MiniAppNavigator`.

Maybe it's better to use named parameters in the example:
```suggestion
                MiniApp.instance().create(
                    appId = "mini_app_id", 
                    miniAppMessageBridge = miniAppMessageBridge,
                    queryParams = "param1=value1&param2=value2"
                )
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppScheme.kt
corycaywood: I don't think we should do any additional check whether the params are valid - we should just pass them directly to the mini app and then it's their responsibility to parse and verify. The reason is because it's really difficult for us to predict how the mini app will use the params, so we would probably end up breaking something for some mini apps.

Also, I think that passing additional `?` in the params is technicadlly valid. For example, this is valid: `param?=value?`. Something like `??param=value` is also valid, but I can't think of any situation where they would purposely want that 🤔  So this is probably a good idea to ensure just one beginning `?` here.

corycaywood (approved):
Looks good 👍  Just one issue with the documentation

miniapp/USERGUIDE.md
khairul-alam-licon: Yeah, I'll update this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppScheme.kt
khairul-alam-licon: Yeah, I agree to let miniapp itself to handle the parsing and verification. And, in future, we can add more verification inside SDK based on client's requests.

minh-rakuten (approved):
LGTM.

PR #233, opened 2020-12-24
feat: miniapp analytics

minh-rakuten:
# Description
- Integrating the analytics into miniapp sdk.
- Tracking hostapp launching, miniapp open and closing.
I have tested the local build with real analytics sdk and see data tracking on Kibana.

Note: To test with dependency implementation, it is required to config library in gradle and account setting.

## Links
MINI-2293

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt
```diff
@@ -75,20 +78,36 @@ class MiniappSdkInitializer : ContentProvider() {
...
+        rasAppId = manifestConfig.rasAppId(),
+        subscriptionKey = manifestConfig.subscriptionKey(),
+        hostAppUserAgentInfo = manifestConfig.hostAppUserAgentInfo(),
+        isPreviewMode = manifestConfig.isPreviewMode(),
+        isTestMode = manifestConfig.isTestMode()
+    )
+
+    private fun executeMiniAppAnalytics(backwardCompatibleHostId: String) {
+        MiniAppAnalytics.init(rasProjectId = backwardCompatibleHostId)
+        MiniAppAnalytics.instance?.sendAnalytics(
```
khairul-alam-licon, line 104: It's a good practice to write note of analytics intention for future maintenance.

e.g. `// send analytics tracking when Host App is launched which contains our SDK.`

In other places:
`// send analytics tracking when Host App displays a mini app.`
`// send analytics tracking when Host App closes a mini app.`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -58,11 +63,21 @@ internal class RealMiniAppDisplay(
     // not rendered successfully in the mini app e.g. select tags
     override suspend fun getMiniAppView(activityContext: Context): View? =
         if (isContextValid(activityContext)) {
+            getMiniAppAnalytics()?.sendAnalytics(
```
khairul-alam-licon, line 66: I think, the good place to send this one inside `provideMiniAppWebView` after ensuring that the `MiniAppWebView` is displaying?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
```diff
@@ -58,11 +63,21 @@ internal class RealMiniAppDisplay(
...
+                eType = Etype.CLICK,
+                actype = Actype.OPEN,
+                miniAppInfo = miniAppInfo
+            )
             provideMiniAppWebView(activityContext)
         } else throw sdkExceptionForNoActivityContext()
 
     @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
     override fun destroyView() {
+        getMiniAppAnalytics()?.sendAnalytics(
```
khairul-alam-licon, line 76: 👍

miniapp/src/main/AndroidManifest.xml
```diff
@@ -15,10 +15,15 @@
...
-          android:exported="false"
-          android:initOrder="100"
-          tools:replace="android:initOrder" />
+            android:name="com.rakuten.tech.mobile.sdkutils.SdkUtilsInitProvider"
+            android:authorities="${applicationId}.SdkUtilsInitProvider"
+            android:exported="false"
+            android:initOrder="100"
+            tools:replace="android:initOrder" />
+
+        <meta-data  android:name="analytics.accountId"
```
khairul-alam-licon, line 24: Hmm.. is it possible to define it as `com.rakuten.tech.mobile.miniapp.analytics.accountId` ? 🤔

khairul-alam-licon (commented):
Looks good so far, I have left few comments meanwhile.

miniapp/src/main/AndroidManifest.xml
minh-rakuten: I think that dummy value is fine 😄.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/RealMiniAppDisplay.kt
minh-rakuten: It's tracking the behavior that developers open mini app view, I guess we don't track whether `MiniAppWebView` init properly.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt
minh-rakuten: good idea.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/Event.kt
```diff
@@ -0,0 +1,7 @@
+package com.rakuten.tech.mobile.miniapp.analytics
```
corycaywood, line 1: I think we should instead create a separate analytics module for `com.rakuten.tech.mobile.analytics` (like mentioned earlier).

We can't release our SDK as it is like this because analytic events will not be sent even if the Host App includes analytics (we don't yet know when public analytics will be available). But using a separate `compileOnly` module will let Host Apps use analytics immediately, and it will still be easy to add the real analytics to the demo app later.

Also, using an "undefined" package of `com.rakuten.tech.mobile.analytics` seems more generic and could let a developer define their own implementation of analytics instead of needing to rely on the dependency we provide.

corycaywood (commented):
Looking good 👍  Just some issues about the IDs and also re-thinking the analytics dependency.

minh-rakuten:
@corycaywood @khairul-alam-licon I have updated to
- Set account config from `BuildConfigField` and put them in params to override the destination to sent.
- Create the dummy analytics dependency used in `compileOnly`.

miniapp/proguard-rules.pro
```diff
@@ -28,3 +28,4 @@
 ##---------------End: proguard configuration for Gson  ----------
 
 -keep class com.google.android.gms.ads.MobileAds { *; }
+-keep class com.rakuten.tech.mobile.analytics.Event { *; }
```
corycaywood, line 31: Do you also need to keep `com.rakuten.tech.mobile.analytics.RatTracker`? I'm not really sure, but it seems like Proguard would fail if the Host App doesn't have the Analytics SDK since it can't find `RatTracker`. Or maybe we need to keep `com.rakuten.tech.mobile.miniapp.analytics.MiniAppAnalytics`?

analytics/build.gradle
```diff
@@ -0,0 +1,44 @@
...
+    kotlinOptions {
+        jvmTarget = '1.6'
+    }
+}
+
+dependencies {
+
+    implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation 'androidx.core:core-ktx:1.3.2'
+    implementation 'androidx.appcompat:appcompat:1.2.0'
```
corycaywood, line 39: You shouldn't need dependencies for appcompat, material, or the testing libraries for this module.

corycaywood (commented):
Looking good, but see my comments about using Kotlin for the `analytics` module.

Also, if you use java instead, you could the simplify the gradle config to just this:
```
plugins {
    id 'java-library'
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -0,0 +1,44 @@
...
+        // Send to this acc/aid
+        params["acc"] = BuildConfig.ANALYTICS_ACC
+        params["aid"] = BuildConfig.ANALYTICS_AID
+
+        params["actype"] = actype.value
+        if (miniAppInfo != null) {
+            params["cp"] = JSONObject()
+                .put("mini_app_project_id", rasProjectId)
+                .put("mini_app_id", miniAppInfo.id)
+                .put("mini_app_version_id", miniAppInfo.version.versionId)
```
corycaywood, line 39: Sorry, I forgot there's one more parameter that needs to be attached to events: `mini_app_sdk_version` for the version of the SDK.

analytics/src/main/java/com/rakuten/tech/mobile/analytics/Tracker.kt
```diff
@@ -0,0 +1,12 @@
+package com.rakuten.tech.mobile.analytics
+
+/** Event tracking dummy class. Please notice the proguard class path. */
+class Event(name: String, parameters: Map<String, Any>) {
+    fun track() {}
+}
+
+class RatTracker {
+    companion object {
+        fun event(name: String, parameters: Map<String, Any>) = Event(name, parameters)
```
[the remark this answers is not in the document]
minh-rakuten: When I replaced the mock one with real analytics sdk, it's working 🎉  so I don't think we need to spend further effort here.

miniapp/proguard-rules.pro
minh-rakuten: I tested with both mocking and real analytics sdk, no need proguard rule for `RatTracker`.

analytics/.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
khairul-alam-licon, line 1: There is an `.gitignore` file in the root directory, so maybe we can remove this.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
minh-rakuten: Updated.

analytics/build.gradle
```diff
@@ -0,0 +1,41 @@
...
+
+    buildTypes {
+        release {
+            minifyEnabled false
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+        }
+    }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_6
+        targetCompatibility JavaVersion.VERSION_1_6
```
khairul-alam-licon, line 28: We can use `VERSION_1_8`?

Maybe we can cleanup some properties like this from this file. I found an optimized version:

```
apply from: '../config/android/library.gradle'
apply plugin: 'kotlin-android'
apply plugin: 'kotlin-android-extensions'

android {
    defaultConfig {
        testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
        consumerProguardFiles 'proguard-rules.pro'
    }

    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
    kotlinOptions {
        jvmTarget = JavaVersion.VERSION_1_8
    }

    dependencies {
        implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlin_version"
    }
}
```

Thus, we don't need to add `consumer-rules.pro` file.

settings.gradle
```diff
@@ -1 +1,2 @@
+include ':analytics'
```
khairul-alam-licon, line 1: We can inline all modules? e.g. `include ':testapp', ':miniapp', ':analytics'`

miniapp/build.gradle
```diff
@@ -75,6 +81,7 @@ dependencies {
     testImplementation "org.robolectric:robolectric:$robolectric"
     testImplementation "com.squareup.okhttp3:mockwebserver:$mockwebserver"
     testImplementation "com.google.android.gms:play-services-ads:$google_ads"
+    testImplementation project(':analytics')
```
khairul-alam-licon, line 84: Not sure if we need this 🤔

khairul-alam-licon (commented):
Looking nice! 
I left few comments based on new `analytics` module change.

miniapp/build.gradle
minh-rakuten: yeah we need this for unit test.

analytics/build.gradle
minh-rakuten: I would keep `VERSION_1_6` because it's just dummy package so I want to keep it as minimum possible.

analytics/src/main/java/com/rakuten/tech/mobile/analytics/Tracker.kt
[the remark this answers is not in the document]
corycaywood: Hmm, strange I got this error when I tried:
```
java.lang.NoSuchFieldError: No field Companion of type Lcom/rakuten/tech/mobile/analytics/RatTracker$Companion
```

Which makes sense because the kotlin code will be compiled to call `RatTracker.Companion.event` which doesn't exist in the real analytics package.

Did you run `./gradlew clean` before building? Or maybe I'm doing something different 🤔 

Seems it should also work if you change it to an `object` class instead:

```
object RatTracker {
  @JvmStatic
  fun event()
}
```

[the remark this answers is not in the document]
minh-rakuten: updated

corycaywood (approved):
Looks good! 👍

analytics/build.gradle
khairul-alam-licon: I was concerning to keep consistency with `miniapp` module because we don't need to manage two different Java versions.

khairul-alam-licon (approved):
Looks good 👍

PR #234, opened 2020-12-25
feat: add crash reporting integration in demo app

khairul-alam-licon:
# Description
Add [appcenter crash reporting library](https://docs.microsoft.com/en-us/appcenter/sdk/crashes/android) to the demo app to diagnosis crash reports in appcenter. 

## Links
- MINI-2009

# Checklist

### DIFF
--- a/testapp/build.gradle
+++ b/testapp/build.gradle
@@ -40,6 +40,7 @@ android {
             projectId               : property("HOST_PROJECT_ID") ?: "test-host-project-id",
             subscriptionKey         : property("HOST_APP_SUBSCRIPTION_KEY") ?: "test-subs-key",
             adMobAppId              : property("ADMOB_APP_ID") ?: "",
+            appCenterSecret         : property("APPCENTER_SECRET") ?: "",
             appName                 : defaultAppName + " DEBUG"
     ]
     def stagingManifestPlaceholders = debugManifestPlaceholders.clone()
@@ -51,6 +52,7 @@ android {
             projectId               : property("HOST_PROJECT_PROD_ID") ?: "test-host-project-id",
             subscriptionKey         : property("HOST_APP_PROD_SUBSCRIPTION_KEY") ?: "test-subs-key",
             adMobAppId              : property("PROD_ADMOB_APP_ID") ?: "",
+            appCenterSecret         : property("PROD_APPCENTER_SECRET") ?: "",
             appName                 : defaultAppName
     ]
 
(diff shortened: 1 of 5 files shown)


minh-rakuten (approved):
Can't wait to see this works!

testapp/build.gradle
```diff
@@ -51,6 +52,7 @@ android {
             projectId               : property("HOST_PROJECT_PROD_ID") ?: "test-host-project-id",
             subscriptionKey         : property("HOST_APP_PROD_SUBSCRIPTION_KEY") ?: "test-subs-key",
             adMobAppId              : property("PROD_ADMOB_APP_ID") ?: "",
+            appCenterSecret         : property("PROD_APPCENTER_SECRET") ?: "",
```
minh-rakuten, line 55: I think we only need `APPCENTER_SECRET` because there is only single AppCenter to manage both staging and production apk.

corycaywood (commented):
Nice, it's looking good 👍 

But there's an issue that we won't be able to see code lines when the APK has been minified with R8. You have to upload the mappings to App Center: https://docs.microsoft.com/en-us/appcenter/diagnostics/android-deobfuscation#app-center-cli

You could add this to the App Center publish step on CircleCI ([see here](https://github.com/rakutentech/android-miniapp/blob/master/.circleci/config.yml#L40)):
```
post-steps:
  - run:
      command: appcenter crashes uploading-mappings // etc..
```

I haven't done this before, so I'm not sure how well it works 🤔  I guess we could just manually upload the mappings for now if it's too difficult to do automatically. What do you think?

khairul-alam-licon:
@corycaywood Yeah, Thank you for pointing out the case when minified with R8. Automatically uploading the `mapping.txt` files would be nice but at this moment, it looks like a bit tricky to identify few configuration values e.g. `--mapping --version-name --version-code`. In that case, can we forward with manual uploading for now?

corycaywood:
> @corycaywood Yeah, Thank you for pointing out the case when minified with R8. Automatically uploading the `mapping.txt` files would be nice but at this moment, it looks like a bit tricky to identify few configuration values e.g. `--mapping --version-name --version-code`. In that case, can we forward with manual uploading for now?

Yeah, I think it's okay for now to upload the mappings manually. Anyways, we only need the mappings in the case that we need to investigate a crash, so probably we can just manually upload them when we need to see the details of a crash.

corycaywood (approved):
LGTM 👍

khairul-alam-licon:
Thank you for the reviews!

PR #239, opened 2021-01-06
rename device permission methods and types	

khairul-alam-licon:
# Description
Rename `MiniAppMessageBridge.requestPermission` to `requestDevicePermission`, and rename `MiniAppPermissionType` to `MiniAppDevicePermissionType` (can use a typealias for this).

Should deprecate `requestPermission` and `MiniAppPermissionsType` but keep the same functionality.

## Links
- MINI-2118

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
```diff
@@ -3,6 +3,7 @@ package com.rakuten.tech.mobile.miniapp.permission
 import androidx.annotation.Keep
 
 /** Type of miniapp permission. **/
+@Deprecated("Use MiniAppDevicePermissionType instead")
 enum class MiniAppPermissionType(val type: String) {
```
corycaywood, line 7: Couldn't you use a `typealias` here? And then you don't need to have two duplicate implementations:
```
typealias MiniAppPermissionType = MiniAppDevicePermissionType
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -124,6 +138,7 @@ abstract class MiniAppMessageBridge {
         when (callbackObj.action) {
             ActionType.GET_UNIQUE_ID.action -> onGetUniqueId(callbackObj)
             ActionType.REQUEST_PERMISSION.action -> onRequestPermission(callbackObj)
+            ActionType.REQUEST_DEVICE_PERMISSION.action -> onRequestDevicePermission(callbackObj)
```
corycaywood, line 141: I don't think we need to change the name of the JavaScript method. In the JS SDK we've encapsulated this call within `requestLocationPermssion`, so the mini app developers never call `requestPermission` directly anyways. I'm afraid that changing the JS SDK name could cause difficulties with backwards compatibility in the future. So this could just be:
```
ActionType.REQUEST_PERMISSION.action -> onRequestDevicePermission(callbackObj)
```
What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -159,14 +174,32 @@ abstract class MiniAppMessageBridge {
 
     private fun onRequestPermission(callbackObj: CallbackObj) {
```
corycaywood, line 175: As mentioned above, I don't think we need to change the JavaScript method name, so probably you can remove this function.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -159,14 +174,32 @@ abstract class MiniAppMessageBridge {
...
+    }
+
+    private fun onRequestDevicePermission(callbackObj: CallbackObj) {
+        try {
+            val permissionParam = Gson().fromJson<DevicePermission>(
+                callbackObj.param.toString(),
+                object : TypeToken<DevicePermission>() {}.type
+            )
+
+            requestDevicePermission(
```
corycaywood, line 200: I think you'll need to do something here similar to what you did when changing the `getUserName` functions. First try calling `requestDevicePermission` and then if it's not implemented, call the deprecated `requestPermission` instead.

corycaywood (commented):
Looking good, but see my comments below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
khairul-alam-licon: Yeah, I'll update this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Yeah, I have updated this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: I have updated the PR, now I am using this when `requestDevicePermission` has not been implemented.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Thanks, I have updated this part.

minh-rakuten (approved):
👍 👍

corycaywood (approved):
Looks good 👍

PR #244, opened 2021-01-15
feat: send a message to a single contact

khairul-alam-licon:
# Description
- Added `ChatBridgeDispatcher.sendMessageToContact` interface in SDK
- Implement `sendMessageToContact` interface in the demo app
- Display a dialog where the user can select a single Contact ID
- After the message is "sent", a dialog should appear that tells the user that the message was sent and to which Contact ID (Doesn't need to actually send a message because we don't have an interface for this in the demo app)

## Links
- MINI-1651

# Checklist
- [x] I have read the [contributing guidelines](../.github/CONTRIBUTING.md).
- [x] I have completed the [SDK Development Learning Path](../.github/CONTRIBUTING.md#SDK-Development-Learning-Path) (if submitting a feature PR)
- [x] I wrote/updated tests for new/changed code
- [x] I removed all sensitive data before every commit, including API endpoints and keys
- [x] I ran `./gradlew check` without errors

PR #245, opened 2021-01-18
refactor: minimize user info bridge

minh-rakuten:
# Description
Refactor the `UserInfoBridgeDispatcher` and wrap it with another container class.
Refactor based on detekt check -> remove some suppress warnings.

Why:
- The `UserInfoBridgeDispatcher` is too big -> make it smaller and simpler.
- `UserInfoBridgeDispatcher` now just only has contract code, single function principle for contract class.
- This is a cleaner way to manage the `UserInfoBridgeDispatcher` initialization.

I also think this would be a good example for [ChatMessageBridgeDispatcher PR](https://github.com/rakutentech/android-miniapp/pull/244).

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
@@ -23,7 +23,7 @@ import com.rakuten.tech.mobile.miniapp.permission.MiniAppDevicePermissionResult
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppDevicePermissionType
 import com.rakuten.tech.mobile.miniapp.permission.ui.MiniAppCustomPermissionWindow
 
-@Suppress("TooGenericExceptionCaught", "ComplexMethod", "LargeClass", "TooManyFunctions", "LongMethod")
+@Suppress("TooGenericExceptionCaught", "TooManyFunctions", "LongMethod", "LargeClass", "ComplexMethod")
 /** Bridge interface for communicating with mini app. **/
 abstract class MiniAppMessageBridge {
     private lateinit var bridgeExecutor: MiniAppBridgeExecutor
@@ -33,7 +33,7 @@ abstract class MiniAppMessageBridge {
     private lateinit var customPermissionWindow: MiniAppCustomPermissionWindow
     private lateinit var miniAppId: String
     private lateinit var activity: Activity
-    private lateinit var userInfoBridgeDispatcher: UserInfoBridgeDispatcher
+    private val userInfoBridgeWrapper = UserInfoBridgeWrapper()
     private val adBridgeDispatcher = AdBridgeDispatcher()
 
     private lateinit var screenBridgeDispatcher: ScreenBridgeDispatcher
@@ -61,10 +61,7 @@ abstract class MiniAppMessageBridge {
 
         this.screenBridgeDispatcher = ScreenBridgeDispatcher(activity, bridgeExecutor, allowScreenOrientation)
         adBridgeDispatcher.setBridgeExecutor(bridgeExecutor)
-
-        if (this::userInfoBridgeDispatcher.isInitialized)
-            this.userInfoBridgeDispatcher.init(bridgeExecutor, customPermissionCache, miniAppId)
-        else this.userInfoBridgeDispatcher = object : UserInfoBridgeDispatcher() {}
+        userInfoBridgeWrapper.setMiniAppComponents(bridgeExecutor, customPermissionCache, miniAppId)
 
         miniAppViewInitialized = true
     }
@@ -143,11 +140,11 @@ abstract class MiniAppMessageBridge {
             ActionType.SHARE_INFO.action -> onShareContent(callbackObj.id, jsonStr)
             ActionType.LOAD_AD.action -> adBridgeDispatcher.onLoadAd(callbackObj.id, jsonStr)
             ActionType.SHOW_AD.action -> adBridgeDispatcher.onShowAd(callbackObj.id, jsonStr)
-            ActionType.GET_USER_NAME.action -> userInfoBridgeDispatcher.onGetUserName(callbackObj.id)
-            ActionType.GET_PROFILE_PHOTO.action -> userInfoBridgeDispatcher.onGetProfilePhoto(callbackObj.id)
-            ActionType.GET_ACCESS_TOKEN.action -> userInfoBridgeDispatcher.onGetAccessToken(callbackObj.id)
+            ActionType.GET_USER_NAME.action -> userInfoBridgeWrapper.onGetUserName(callbackObj.id)
+            ActionType.GET_PROFILE_PHOTO.action -> userInfoBridgeWrapper.onGetProfilePhoto(callbackObj.id)
+            ActionType.GET_ACCESS_TOKEN.action -> userInfoBridgeWrapper.onGetAccessToken(callbackObj.id)
             ActionType.SET_SCREEN_ORIENTATION.action -> screenBridgeDispatcher.onScreenRequest(callbackObj)
-            ActionType.GET_CONTACTS.action -> userInfoBridgeDispatcher.onGetContacts(callbackObj.id)
+            ActionType.GET_CONTACTS.action -> userInfoBridgeWrapper.onGetContacts(callbackObj.id)
         }
     }
 
@@ -158,11 +155,8 @@ abstract class MiniAppMessageBridge {
      * Set implemented userInfoBridgeDispatcher.
      * Can use the default provided class from sdk [UserInfoBridgeDispatcher].
      **/
-    fun setUserInfoBridgeDispatcher(bridgeDispatcher: UserInfoBridgeDispatcher) {
-        userInfoBridgeDispatcher = bridgeDispatcher
-        if (miniAppViewInitialized)
-            userInfoBridgeDispatcher.init(bridgeExecutor, customPermissionCache, miniAppId)
-    }
+    fun setUserInfoBridgeDispatcher(bridgeDispatcher: UserInfoBridgeDispatcher) =
+        userInfoBridgeWrapper.setUserInfoBridgeDispatcher(bridgeDispatcher)
 
     private fun onGetUniqueId(callbackObj: CallbackObj) {
         try {
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
@@ -277,7 +277,7 @@ abstract class MiniAppMessageBridge {
 }
 
 internal object ErrorBridgeMessage {
-    const val NO_IMPL = "method has not been implemented by the Host App."
+    const val NO_IMPL = "has not been implemented by the Host App."
     const val ERR_NO_SUPPORT_HOSTAPP = "No support from hostapp"
     const val ERR_UNIQUE_ID = "Cannot get unique id:"
     const val ERR_REQ_DEVICE_PERMISSION = "Cannot request device permission:"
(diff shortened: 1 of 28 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridgeDispatcher.kt
```diff
@@ -1,26 +1,13 @@
...
-    "TooGenericExceptionCaught", "LongMethod", "UnnecessaryAbstractClass",
-    "LargeClass", "TooManyFunctions"
-)
-abstract class UserInfoBridgeDispatcher {
-
-    private lateinit var bridgeExecutor: MiniAppBridgeExecutor
-    private lateinit var customPermissionCache: MiniAppCustomPermissionCache
-    private lateinit var miniAppId: String
-
+interface UserInfoBridgeDispatcher {
```
khairul-alam-licon, line 10: I think, this a breaking change for the host app because now they need to call it as `object : UserInfoBridgeDispatcher`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -158,11 +155,8 @@ abstract class MiniAppMessageBridge {
      * Set implemented userInfoBridgeDispatcher.
      * Can use the default provided class from sdk [UserInfoBridgeDispatcher].
      **/
-    fun setUserInfoBridgeDispatcher(bridgeDispatcher: UserInfoBridgeDispatcher) {
-        userInfoBridgeDispatcher = bridgeDispatcher
-        if (miniAppViewInitialized)
-            userInfoBridgeDispatcher.init(bridgeExecutor, customPermissionCache, miniAppId)
-    }
+    fun setUserInfoBridgeDispatcher(bridgeDispatcher: UserInfoBridgeDispatcher) =
```
khairul-alam-licon, line 158: From now, if HostApp forget to invoke this function, then I think it will cause issue. 

e.g. in sample app, I was noticing unlimited loading when not calling this function.
Screen Shot 2021-01-22 at 15 19 07

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -143,11 +140,11 @@ abstract class MiniAppMessageBridge {
             ActionType.SHARE_INFO.action -> onShareContent(callbackObj.id, jsonStr)
             ActionType.LOAD_AD.action -> adBridgeDispatcher.onLoadAd(callbackObj.id, jsonStr)
             ActionType.SHOW_AD.action -> adBridgeDispatcher.onShowAd(callbackObj.id, jsonStr)
-            ActionType.GET_USER_NAME.action -> userInfoBridgeDispatcher.onGetUserName(callbackObj.id)
```
khairul-alam-licon, line 146: I think, if the purpose is to reduce the size of the class, then we can skip changing anything in `MiniAppMessageBridge` which will be safer for the HostApp also.

Rather we can improve your idea e.g.

```
internal class UserInfoBridgeWrapper(
    private val bridgeExecutor: MiniAppBridgeExecutor,
    private val customPermissionCache: MiniAppCustomPermissionCache,
    private val miniAppId: String,
    private val userInfoBridgeDispatcher: UserInfoBridgeDispatcher
) {

    fun onGetUserName(callbackId: String) {
        // move the implementation from `UserInfoBridgeDispatcher` to here like you already did.
    }
}

// then you can call it inside `UserInfoBridgeDispatcher`.

abstract class UserInfoBridgeDispatcher {

      internal fun init(
        bridgeExecutor: MiniAppBridgeExecutor,
        miniAppCustomPermissionCache: MiniAppCustomPermissionCache,
        miniAppId: String
       ) {
        ...
        this.userInfoWrapper = UserInfoBridgeWrapper(bridgeExecutor, customPermissionCache, miniAppId, this)
       }
   
    internal fun onGetUserName(callbackId: String) = userInfoWrapper.onGetUserName(callbackId)
}
```

What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
```diff
@@ -0,0 +1,154 @@
...
+@Suppress("TooGenericExceptionCaught")
+internal class UserInfoBridgeWrapper {
+    private lateinit var bridgeExecutor: MiniAppBridgeExecutor
+    private lateinit var customPermissionCache: MiniAppCustomPermissionCache
+    private lateinit var miniAppId: String
+    private var isMiniAppComponentReady = false
+    private lateinit var userInfoBridgeDispatcher: UserInfoBridgeDispatcher
+    private var isReady = false
+
+    internal fun setMiniAppComponents(
```
khairul-alam-licon, line 22: We don't need to mention this function as `internal` additionally as the entire class is already `internal` 🤔

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -160,7 +160,7 @@ class MiniAppDisplayActivity : BaseActivity() {
         miniAppMessageBridge.setAdMobDisplayer(AdMobDisplayer(this@MiniAppDisplayActivity))
         miniAppMessageBridge.allowScreenOrientation(true)
 
-        val userInfoBridgeDispatcher = object : UserInfoBridgeDispatcher() {
```
khairul-alam-licon, line 163: We need to notify the users for this change using changelog.

khairul-alam-licon (commented):
Thank you for your thinking, I haven't checked how much we could minimize by this refactoring but I can guess that `UserInfoBridgeWrapper` will also face the similar problem in future. But of course, we need to consider some kind of patterns for the minimization. 

Besides, I have left few comments...

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: This one has a cross reference problem. Anyway, not only `UserInfoBridgeDispatcher`, the `MiniAppMessageBridge` is smaller and not need to maintain the complexity of initialization as before.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: I noticed that in the master branch implementation, if user does not invoke this function, there is no response or show loading.
The error should be thrown so I will throw the error then.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
minh-rakuten: Correct but no harm to be cautious 😄

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
minh-rakuten: I will bring it back as abstract class.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
khairul-alam-licon: Yeah, no harm but we could write clean code ;)

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridgeDispatcher.kt
minh-rakuten: So unfortunate that we used abstract class before but it's okay, I will bring it back as abstract class to not create a change in host app.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
minh-rakuten: I have updated this part. You will see:

In master branch, there is no response in js sample mini app (user details screen did not handle ui here but Auth screen does so you can check it) and no error sent to miniapp.
In this refactor, there is also no response but the error has been sent to miniapp, developer can see it in chrome log.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Yeah, +1 for removing complexity of initialization from `MiniAppMessageBridge`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Thank you for this update, I think this is a good improvement.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
```diff
@@ -0,0 +1,147 @@
...
+        this.customPermissionCache = miniAppCustomPermissionCache
+        this.miniAppId = miniAppId
+        isMiniAppComponentReady = true
+    }
+
+    fun setUserInfoBridgeDispatcher(userInfoBridgeDispatcher: UserInfoBridgeDispatcher) {
+        this.userInfoBridgeDispatcher = userInfoBridgeDispatcher
+    }
+
+    private fun <T> whenReady(callbackId: String, callback: () -> T) {
```
khairul-alam-licon, line 36: 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
```diff
@@ -0,0 +1,147 @@
+package com.rakuten.tech.mobile.miniapp.js
```
khairul-alam-licon, line 1: This file can be moved to `com.rakuten.tech.mobile.miniapp.js.userinfo`.

khairul-alam-licon (approved):
Looks good 👍 
Removing initialization complexity from `MiniAppMessageBridge` is a good address.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
minh-rakuten: Let me move it.

corycaywood (approved):
Thanks for doing this, looks good 👍 

Just left a minor suggestion.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/UserInfoBridgeWrapper.kt
```diff
@@ -0,0 +1,147 @@
...
+import com.rakuten.tech.mobile.miniapp.js.ErrorBridgeMessage.NO_IMPL
+import com.rakuten.tech.mobile.miniapp.js.userinfo.Contact
+import com.rakuten.tech.mobile.miniapp.js.userinfo.TokenData
+import com.rakuten.tech.mobile.miniapp.js.userinfo.UserInfoBridgeDispatcher
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionCache
+import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
+import java.util.ArrayList
+
+@Suppress("TooGenericExceptionCaught")
+internal class UserInfoBridgeWrapper {
```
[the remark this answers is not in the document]
minh-rakuten: yes, sounds better.

PR #257, opened 2021-02-16
fix: ad map & post message error

minh-rakuten:
# Description
- Remove ad unit id in map when ad loads failed -> user can able to call load ad again.
- Escape character for bridge post error.

## Links
MINI-2706
MINI-2707

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMobDisplayer.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMobDisplayer.kt
@@ -55,6 +55,7 @@ class AdMobDisplayer(private val context: Activity) : MiniAppAdDisplayer, Corout
                     }
 
                     override fun onAdFailedToLoad(adError: LoadAdError) {
+                        interstitialAdMap.remove(adUnitId)
                         onFailed.invoke(adError.message)
                     }
                 }
@@ -92,7 +93,7 @@ class AdMobDisplayer(private val context: Activity) : MiniAppAdDisplayer, Corout
                 val ad = RewardedAd(context, adUnitId)
                 rewardedAdMap[adUnitId] = ad
 
-                val adLoadCallback = createRewardedAdLoadCallback(onLoaded, onFailed)
+                val adLoadCallback = createRewardedAdLoadCallback(adUnitId, onLoaded, onFailed)
 
                 ad.loadAd(AdRequest.Builder().build(), adLoadCallback)
             }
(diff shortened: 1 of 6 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/ads/AdMobDisplayer.kt
```diff
@@ -55,6 +55,7 @@ class AdMobDisplayer(private val context: Activity) : MiniAppAdDisplayer, Corout
                     }
 
                     override fun onAdFailedToLoad(adError: LoadAdError) {
+                        interstitialAdMap.remove(adUnitId)
```
corycaywood, line 58: Should we add a unit test for this?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebView.kt
```diff
@@ -134,7 +134,7 @@ internal open class MiniAppWebView(
     override fun runErrorCallback(callbackId: String, errorMessage: String) {
         post {
             evaluateJavascript(
-                "MiniAppBridge.execErrorCallback(\"$callbackId\", \"$errorMessage\")"
+                "MiniAppBridge.execErrorCallback(`$callbackId`, `${errorMessage.replace("'", "\\'")}`)"
```
corycaywood, line 137: Same here - maybe we should have a unit test for escaped characters?

Also, I just noticed that `runSuccessCallback` above is escaping backticks - I don't remember the particular reason that was needed, but maybe we should have a shared escaping function for both `runSuccessCallback` and `runErrorCallback` to ensure that quotation marks and backticks are escaped for both? What do you think?

corycaywood: Oh, I just noticed: the response in `runSuccessCallback` is being sent with backticks surrounding it, so that's why it needs to escape backticks. So I think here maybe you need to do the same thing here - escape backticks instead of single quotes

minh-rakuten: Updated with backtick escape and unit test.

corycaywood (approved):
Looks good! 👍

khairul-alam-licon (approved):
Looks good 👍

khairul-alam-licon:
Thank you for adding the changelog, I think, it's ready to merge.

PR #259, opened 2021-02-17
sample: params validation

minh-rakuten:
# Description
Add the url query validation in setting of sample app.

## Links
MINI-2666

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
@@ -10,6 +10,7 @@ import android.view.MenuItem
 import androidx.appcompat.widget.Toolbar
 import androidx.databinding.DataBindingUtil
 import com.rakuten.tech.mobile.miniapp.MiniApp
+import com.rakuten.tech.mobile.miniapp.MiniAppSdkException
 import com.rakuten.tech.mobile.miniapp.testapp.R
 import com.rakuten.tech.mobile.miniapp.testapp.databinding.SettingsMenuActivityBinding
 import com.rakuten.tech.mobile.testapp.AppScreen.MINI_APP_INPUT_ACTIVITY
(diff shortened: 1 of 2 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
```diff
@@ -166,12 +166,14 @@ class SettingsMenuActivity : BaseActivity() {
         launch {
             try {
                 MiniApp.instance(AppSettings.instance.miniAppSettings).listMiniApp()
+                URL("https://www.example.com$urlParameters").toURI()
```
corycaywood, line 169: If I remember correctly, we automatically add the `?` to the params string. So I think this should be:
```suggestion
                URL("https://www.example.com?$urlParameters").toURI()
```

corycaywood (commented):
Looking good - just one thing I think needs to be fixed.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
```diff
@@ -166,12 +166,14 @@ class SettingsMenuActivity : BaseActivity() {
...
+
                 settings.isSettingSaved = true
                 runOnUiThread {
                     settingsProgressDialog.cancel()
                     navigateToPreviousScreen()
                 }
-            } catch (error: MiniAppSdkException) {
+            } catch (error: Exception) {
```
khairul-alam-licon, line 176: We also can utilize it to show meaningful message to the demo app user e.g.

```
if (error.message.toString().contains("Illegal character in fragment"))
   showAlertDialog(this@SettingsMenuActivity, "Invalid url parameter: " + urlParameters)
else
   showAlertDialog(this@SettingsMenuActivity, error.message.toString())
```

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
minh-rakuten: updated.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
minh-rakuten: updated.

khairul-alam-licon: Ah, I see! 
Ok, according to the new change, is it possible to use `MalformedURLException` here instead of using a generic Exception? Or, we can use a generic `errTitle` in the alert dialog instead of `"URL parameter"`. 
Because not every `Exception` will be under `"URL parameter"`.

khairul-alam-licon (approved):
Looks good, I have let a minor suggestion below.

corycaywood (approved):
LGTM 👍

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/SettingsMenuActivity.kt
minh-rakuten: Not only `MalformedURLException`, there is also `URLException` which could be thrown.

Only parameter check can throw `Exception` in this scope so I used `Exception` here.

PR #264, opened 2021-02-26
draft feat: required and optional custom permission 

khairul-alam-licon:
# Description
Describe the changes in this pull request.
Explain your rationale of technical decisions you made (unless discussed before).

TODO: unit test

## Links
- MINI-2503

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -113,11 +153,56 @@ internal class RealMiniApp(
...
     }
 
+    private fun isRequiredPermissionDenied(appId: String): Boolean {
+        getCachedRequiredPermissions(appId).find {
+            it.second != MiniAppCustomPermissionResult.ALLOWED
+        }?.let { return true }
+        return false
+    }
+
+    private fun getCachedRequiredPermissions(
```
corycaywood, line 163: Could you move `getCachedRequiredPermissions` and `getCachedOptionalPermissions` into `MiniAppManifestCache`? It seems like it would be better encapsulated in that class.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -113,11 +153,56 @@ internal class RealMiniApp(
...
+
+        val manifest = miniAppManifestCache.readMiniAppManifest(appId)
+        // TODO: check if it throws error for empty version id
+        manifest.requiredPermissions.forEach { (first) ->
+            miniAppCustomPermissionCache.readPermissions(appId).pairValues.find {
+                it.first == first
+            }?.let { requiredPermissions.add(it) }
+        }
+
+        return requiredPermissions
```
corycaywood, line 177: Couldn't you use `mapNotNull` here and for `getCachedOptionalPermissions`?
```suggestion
        val manifest = miniAppManifestCache.readMiniAppManifest(appId)
        val cachedPermissions =  miniAppCustomPermissionCache.readPermissions(appId).pairValues
        return manifest.requiredPermissions.mapNotNull {(first) ->
            cachedPermissions.find { it.first == first }
        }
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+ * A caching class to read and store the [MiniAppManifest] per MiniApp
+ * using [SharedPreferences].
+ */
+// TODO: unit testcases
+internal class MiniAppManifestCache(context: Context) {
+    private val prefs: SharedPreferences = context.getSharedPreferences(
+        "com.rakuten.tech.mobile.miniapp.manifest.cache", Context.MODE_PRIVATE
+    )
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
corycaywood, line 22: Why does this need the `otherwise = VisibleForTesting.NONE`? I think it should use `PRIVATE` by default, so you don't need to set `otherwise`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+    fun doesDataExist(miniAppId: String) = prefs.contains(miniAppId)
+
+    /**
+     * Reads the grant results from SharedPreferences.
+     * @param [miniAppId] the key provided to find the stored manifest per MiniApp.
+     * @return [MiniAppManifest] an object to contain the manifest per MiniApp.
+     * if data has been stored in cache, otherwise empty value.
+     */
+    @SuppressLint("RestrictedApi")
```
corycaywood, line 31: See my comment above: I think you don't need this if you remove the `otherwise`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+        val empty = MiniAppManifest(emptyList(), emptyList(), emptyMap())
+        return if (doesDataExist(miniAppId)) {
+            try {
+                val cachedManifest: MiniAppManifest = Gson().fromJson(
+                    prefs.getString(miniAppId, ""),
+                    object : TypeToken<MiniAppManifest>() {}.type
+                )
+                cachedManifest
+            } catch (e: Exception) {
+                // if there is any exception, return the empty value
```
corycaywood, line 42: Maybe you should also do a `Log.e` for the exception?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+ * using [SharedPreferences].
+ */
+// TODO: unit testcases
+internal class MiniAppManifestCache(context: Context) {
+    private val prefs: SharedPreferences = context.getSharedPreferences(
+        "com.rakuten.tech.mobile.miniapp.manifest.cache", Context.MODE_PRIVATE
+    )
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun doesDataExist(miniAppId: String) = prefs.contains(miniAppId)
```
corycaywood, line 23: Actually, does this even need to be a separate method? You should be able to write unit tests by directly modifying shared preferences.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+     */
+    @SuppressLint("RestrictedApi")
+    fun readMiniAppManifest(miniAppId: String): MiniAppManifest {
+        val empty = MiniAppManifest(emptyList(), emptyList(), emptyMap())
+        return if (doesDataExist(miniAppId)) {
+            try {
+                val cachedManifest: MiniAppManifest = Gson().fromJson(
+                    prefs.getString(miniAppId, ""),
+                    object : TypeToken<MiniAppManifest>() {}.type
+                )
```
corycaywood, line 39: Maybe the retrieval from `prefs` should be moved outside of this `if` statement? It would make the flow a little more clear. For example (including the change to a nullable return type):
```suggestion
        val manifestJsonStr = prefs.getString(miniAppId, null) 
            ?: return empty
        return try {
            val cachedManifest: MiniAppManifest = Gson().fromJson(
                manifestJsonStr,
                object : TypeToken<MiniAppManifest>() {}.type
            )
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+                cachedManifest
+            } catch (e: Exception) {
+                // if there is any exception, return the empty value
+                empty
+            }
+        } else {
+            // if value hasn't been found in SharedPreferences, save the empty value
+            storeMiniAppManifest(miniAppId, empty)
```
corycaywood, line 47: This seems like it could be dangerous - the function is named `readMiniAppManifest`, but in some cases it can also `store` and not just `read` so someone could call this function not realizing it has this behavior. Is there a case why you would need to store the empty manifest?

I think it would be better to remove this and just return `null` in the case that the manifest doesn't exist.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * Reads the grant results from SharedPreferences.
+     * @param [miniAppId] the key provided to find the stored manifest per MiniApp.
+     * @return [MiniAppManifest] an object to contain the manifest per MiniApp.
+     * if data has been stored in cache, otherwise empty value.
+     */
+    @SuppressLint("RestrictedApi")
+    fun readMiniAppManifest(miniAppId: String): MiniAppManifest {
+        val empty = MiniAppManifest(emptyList(), emptyList(), emptyMap())
```
corycaywood, line 33: As I mentioned above, I think the return type should be nullable instead of returning an empty map. This is because a `null` return value signals to the Host App that the Mini App has not been downloaded.

testapp/src/main/res/layout/item_list_manifest_permission.xml
```diff
@@ -24,16 +23,30 @@
...
+          android:layout_width="match_parent"
           android:layout_height="wrap_content"
-          android:textColor="@android:color/black"
-          android:textSize="@dimen/text_large_16"
-          app:layout_constraintStart_toStartOf="parent"
-          app:layout_constraintTop_toTopOf="parent"
-          tools:text="Manifest Permission" />
+          android:orientation="horizontal">
+
+          <TextView
```
corycaywood, line 31: We also need to display the "Optional" permissions on the first-time launch screen. Optional permissions should be shown with a toggle switch similar to the permissions management screen.

corycaywood (commented):
It's looking good! I've left some comments below.

testapp/src/main/res/layout/item_list_manifest_permission.xml
khairul-alam-licon: Yeah, I am displaying both "required" & "optional" permissions according to the following ux, basically, I am currently using a single adapter to manage both. should I improve here something e.g. mentioning "optional" keyword also? Sorry, previously I haven't asked enough idea for this part...

Screen Shot 2021-03-02 at 19 59 17

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
khairul-alam-licon: Yeah, I have updated this part, currently the responsibility of `readMiniAppManifest` is only returning the manifest if available, otherwise `null`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
khairul-alam-licon: Yeah, I have removed this function and will work on unit test accordingly. Thank you for all the improvement feedbacks. 

_P/S: I have realized, I can address some of the feedbacks in custom permission cache too using a separate ticket later._

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: I have updated this part with moving it to `MiniAppManifestCache`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
khairul-alam-licon: I have updated this part with moving it to `MiniAppManifestCache`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -179,7 +190,8 @@ abstract class MiniApp internal constructor() {
                     initVerifier = { CachedMiniAppVerifier(context) }
                 ),
                 miniAppInfoFetcher = MiniAppInfoFetcher(apiClient),
-                miniAppCustomPermissionCache = MiniAppCustomPermissionCache(context)
+                miniAppCustomPermissionCache = customPermissionCache,
+                miniAppManifestCache = MiniAppManifestCache(context, customPermissionCache)
```
minh-rakuten, line 194: You don't need to init them here, you can init them in with `lazy` in `RealMiniApp`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -7,29 +7,67 @@ import com.rakuten.tech.mobile.miniapp.display.Displayer
...
-@Suppress("TooManyFunctions")
+@Suppress("TooManyFunctions", "LongMethod", "ExpressionBodySyntax")
 internal class RealMiniApp(
     private val apiClientRepository: ApiClientRepository,
     private val miniAppDownloader: MiniAppDownloader,
     private val displayer: Displayer,
     private val miniAppInfoFetcher: MiniAppInfoFetcher,
-    private val miniAppCustomPermissionCache: MiniAppCustomPermissionCache
+    private val miniAppCustomPermissionCache: MiniAppCustomPermissionCache,
+    private val miniAppManifestCache: MiniAppManifestCache
```
minh-rakuten, line 22: Let's define them in as lazy property var with initialization.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
```diff
@@ -0,0 +1,101 @@
...
+        return false
+    }
+
+    @VisibleForTesting
+    fun getCachedRequiredPermissions(
+        miniAppId: String
+    ): List<Pair<MiniAppCustomPermissionType, MiniAppCustomPermissionResult>> {
+        val manifest = readMiniAppManifest(miniAppId)
+        val cachedPermissions = miniAppCustomPermissionCache.readPermissions(miniAppId).pairValues
+        return manifest?.requiredPermissions?.mapNotNull { (first) ->
```
minh-rakuten, line 86: if `manifest` is null then does `getCachedRequiredPermissions` get crash?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -7,29 +7,67 @@ import com.rakuten.tech.mobile.miniapp.display.Displayer
...
-        miniAppCustomPermissionCache.readPermissions(miniAppId)
+    override fun getCustomPermissions(
+        miniAppId: String
+    ): MiniAppCustomPermission {
+        // return only the permissions listed in the Mini App's manifest.
+        val manifestPermissions = miniAppManifestCache.getCachedAllPermissions(miniAppId)
+        return MiniAppCustomPermission(miniAppId, manifestPermissions)
+    }
+
+    override fun setCustomPermissions(miniAppCustomPermission: MiniAppCustomPermission) {
```
minh-rakuten, line 43: I think it's better to remove the `temporaryManifest` and change the signature into `setCustomPermission(miniAppManifest, miniAppCustomPermission)`

minh-rakuten (commented):
Looks good.

@corycaywood @khairul-alam-licon That `temporaryManifest` in `RealMiniApp` needs to be considered.

khairul-alam-licon:
> @corycaywood @khairul-alam-licon That temporaryManifest in RealMiniApp needs to be considered.

Yeah, I want to cache the manifest only when `create` a miniapp. So, I used this temporary variable in the miniapp instance. Otherwise, I was caching the manifest immediately while fetching it from the api.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: Yeah, we could. But, it will be extra work for the hostapp as SDK's miniapp instance already have the manifest information, so HostApp doesn't need to send it additionally to store the permission.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppManifestCache.kt
khairul-alam-licon: Yes right, so I have updated `getCachedRequiredPermissions` & `getCachedOptionalPermissions` to handle this null.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
minh-rakuten: I'm thinking about user does not know which miniapp they are set custom permission. When they use it wrongly, they will think it's sdk fault.

I'll let you & @corycaywood make a decision on it 😄

khairul-alam-licon: Ok, I found a discussion in iOS side: https://github.com/rakutentech/ios-miniapp/pull/227#discussion_r587246181
I'll follow up accordingly. I'll vote for caching it right after fetching from api, so we don't need this `temporaryManifest` variable anymore.

corycaywood: Actually the discussion about caching the manifest was relevant to checking the manifest permissions when calling `MiniApp.create`, not when calling `MiniApp.setCustomPermissions`. This is a little bit tricky, so I'm not sure how to handle it...

The `temporaryManifest` variable here is just caching the last downloaded manifest without saving the mini app id or version, right? So how do you know that you're checking the correct manifest? Also, `setCustomPermissions` can be used on the permissions management screen, so I think this current implementation could cause problems in that use case.

Maybe you don't need to check the manifest permissions when calling `setCustomPermission`? i.e. we allow the host app to store any permissions, even if not in the manifest. But then we would need to change some of our implementation so that we always check the manifest when reading the cached permissions. What do you think?

khairul-alam-licon: Okay, I have updated this part.

Yeah, it makes sense not to check manifest permissions when calling `setCustomPermission`. But, we still need to check manifest when calling `getCustomPermissions`... 🤔 Also, I have removed `temporaryManifest` and saving the manifest when fetching the latest manifest from api. 

@corycaywood Do you think, it's looked okay, then I'll fix the broken tests.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -45,6 +55,9 @@ internal class RealMiniApp(
         queryParams: String
     ): MiniAppDisplay = when {
         appId.isBlank() -> throw sdkExceptionForInvalidArguments()
+        miniAppManifestCache.isRequiredPermissionDenied(
```
corycaywood, line 58: On iOS, it's first downloading the mini app and then checking if permissions are denied. It's a subtle distinction, but basically the downloading of the mini app shouldn't be dependent on whether the permissions are granted. It's the launching of the mini app that is dependent on the permissions.

So basically, I think you should move the call to `isRequestedPermissionsDenied` to after `miniAppDownloader.getMiniApp`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -97,8 +113,16 @@ internal class RealMiniApp(
         }
     }
 
-    override suspend fun getMiniAppManifest(appId: String, versionId: String): MiniAppManifest =
-        miniAppDownloader.fetchMiniAppManifest(appId, versionId)
+    override suspend fun getMiniAppManifest(appId: String, versionId: String): MiniAppManifest {
+        val latestManifest = miniAppDownloader.fetchMiniAppManifest(appId, versionId)
+        // store the latest manifest value in cache after fetched from api
+        miniAppManifestCache.storeApiManifest(appId, latestManifest)
```
corycaywood, line 119: This is only storing the App ID and not the version, so we still don't have a guarantee that the manifest is for the correct version. Could you also cache the version?

Also, I'm not sure if `MiniAppManifestCache` is the right place to cache this. It's unrelated to the functionality of our manifest cache - it's only about saving us from hitting the API multiple times in a row for the same thing. So maybe it should be moved into `MiniAppDownloader`? And then we can cache only one manifest: the last one that we retrieved from the API (along with the app id and version id).

That also saves us from having to add extra calls to `miniAppManifestCache` - we can just call `miniAppDownloader.fetchMiniAppManifest`, and if it has the cached version then it returns the cached manifest, otherwise it downloads it.

What do you think?

corycaywood (commented):
I've left a couple more comments. I'm not sure if `MiniAppManifestCache` is the right place to store the API response. I was thinking about something like this:

1. Cache the last retrieved manifest and associated mini app ID and version ID closer to the API request (such as in `MiniAppDownloader`).
2. When `MiniAppDownloader.fetchManifest` is called, it returns the cached version if already downloaded (or downloads it if not cached)
3. When `MiniApp.create` is called, first download the mini app, then get the manifest using  `MiniAppDownloader.fetchManifest`
4. Store the manifest for downloaded version to `MiniAppManifestCache`
5. Finally, check `MiniAppManifestCache.isRequiredPermissionDenied`, then create and return the mini app.

What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
corycaywood: Sorry, I had another comment that I forgot to click the submit button on 😣 

Maybe we could store all of the passed permissions when `setCustomPermissions` is called.

An then when the Mini App is launched (`MiniApp.create`), we can do the following:
- Check if manifest is downloaded (or cached in memory) for the specified mini app id and version
  - If not, then download the manifest
- Check the cached permissions and update them to remove any permissions which aren't contained in the manifest

What do you think?

khairul-alam-licon: Updated.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: Yes, I used version id too to store the manifest from api, so its saving us from hitting the api unnecessarily.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/ManifestApiCache.kt
```diff
@@ -0,0 +1,42 @@
...
+        }
+    }
+
+    fun storeManifest(
+        miniAppId: String,
+        versionId: String,
+        miniAppManifest: MiniAppManifest
+    ) {
+        val jsonToStore: String = Gson().toJson(miniAppManifest)
+        prefs.edit().putString(primaryKey(miniAppId, versionId), jsonToStore).apply()
```
corycaywood, line 38: This will cache every manifest response separately, so this cache will continually grow with outdated entries. 

Really we only need to cache a single entry: the most recently retrieved manifest. But that makes it more difficult to use with the miniAppId and versionId. So maybe you could just clear this cache every time that you store a new manifest? For example:

```suggestion
        prefs.edit().clear().putString(primaryKey(miniAppId, versionId), jsonToStore).apply()
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -97,9 +129,20 @@ internal class RealMiniApp(
         }
     }
 
+    private suspend fun setManifestFromApi(appId: String, versionId: String) {
+        val apiManifest = getMiniAppManifest(appId, versionId)
+        val cachedManifest = downloadedManifestCache.readDownloadedManifest(appId)
+
+        if (cachedManifest != apiManifest)
```
corycaywood: This seems a little strange to compare the manifests directly. What we actually want to compare is the `versionId` for the downloaded manifest and then only download the new manifest if the cached manifest is for a different `versionId`. This will avoid doing an extra network request in the case that the mini app is already downloaded.

Could you update the `DownloadedManifestCache` to also store the `versionId`? I guess you would need to wrap the manifest in another object, similar to what iOS did here: https://github.com/rakutentech/ios-miniapp/pull/227/files#diff-e2aeedd67de0916c8e00eba89a3fcd7d60f5d49c2a466eb995eed0624f90d5f9R59

Or maybe there's another way to achieve this?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -80,22 +80,44 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
-
         val miniAppId = miniAppCustomPermission.miniAppId
         val allPermissions = prepareAllPermissionsToStore(miniAppId, supplied)
         applyStoringPermissions(MiniAppCustomPermission(miniAppId, allPermissions))
     }
 
+    fun updateManifestPermissions(
+        miniAppId: String,
+        manifestPermissions: List<Pair<MiniAppCustomPermissionType, MiniAppCustomPermissionResult>>
+    ) {
```
corycaywood: The function name makes it sound like this is updating the cached manifest - but it's actually updating the cached permissions. How about `updateFromManifestPermissions` instead?

Or actually - this class doesn't even need to know that these are manifest permissions. It just needs to know that it should remove permissions which weren't passed. So how about this?
```suggestion
    fun removePermissionsNotMatching(
        miniAppId: String,
        permissions: List<Pair<MiniAppCustomPermissionType, MiniAppCustomPermissionResult>>
    ) {
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -80,22 +80,44 @@ internal class MiniAppCustomPermissionCache(context: Context) {
...
+                applyStoringPermissions(filteredValue)
+            } else {
+                cachedPermissions.removeAll { (first) ->
+                    first.type in changedPermissions.groupBy { it.first.type }
+                }
+
+                val filteredValue = MiniAppCustomPermission(miniAppId, cachedPermissions)
+                applyStoringPermissions(filteredValue)
+            }
+        }
```
corycaywood, line 118: I'm not really understanding the logic in this function, but maybe I'm missing something. Is this also adding permissions from the manifest to the cached permissions?

Basically we only want to make sure that the cached permissions don't contain any permissions which aren't in the manifest (we don't need to worry about adding any permissions from the manifest to the cached permissions). So couldn't you just do something like this?

```suggestion
        val cachedPermissions = readPermissions(miniAppId).pairValues.toMutableList()
        val newPermissions = cachedPermissions.mapNotNull { cached -> 
            manifestPermissions.find { it.first == cached.first } 
        }
        val filteredValue =
            MiniAppCustomPermission(miniAppId, newPermissions)
        applyStoringPermissions(filteredValue)
```

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppPermissionAdapter.kt
```diff
@@ -28,53 +26,46 @@ class PreloadMiniAppPermissionAdapter :
...
 
     override fun onBindViewHolder(holder: ViewHolder, position: Int) {
-        holder.permissionName.text = toReadableName(manifestPermissionNames[position])
+        holder.permissionName.text = toReadableName(manifestPermissions[position].type)
+        holder.permissionSwitch.visibility =
+            if (manifestPermissions[position].isRequired) View.GONE else View.VISIBLE
+        holder.permissionStatus.visibility =
+            if (manifestPermissions[position].isRequired) View.VISIBLE else View.GONE
+
         holder.permissionSwitch.isChecked =
```
corycaywood, line 35: The "optional" permissions should default to `true` if the permission hasn't been requested before - i.e. the first time that the mini app is downloaded.

corycaywood (commented):
It's looking really close. I left a few more comments below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: I have added a separate class called `CachedManifest` for this.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/ManifestApiCache.kt
```diff
@@ -0,0 +1,42 @@
...
+import com.google.gson.Gson
+import com.google.gson.reflect.TypeToken
+import com.rakuten.tech.mobile.miniapp.MiniAppManifest
+import kotlin.Exception
+
+/**
+ * A caching class to read and store the [MiniAppManifest] per MiniApp's id & version id using [SharedPreferences].
+ */
+@Suppress("TooGenericExceptionCaught")
+internal class ManifestApiCache(
```
corycaywood, line 15: I'm not really sure if the `storage` folder is the right place for this 🤔  This is where information about the downloaded Mini App is stored, so I feel like it's going to be confusing seeing both `DownloadedManifestCache` and `ManifestApiCache` here.

Maybe you could move this class to the `api` folder and name it something like `ManifestResponseCache`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -47,6 +61,9 @@ internal class RealMiniApp(
         appId.isBlank() -> throw sdkExceptionForInvalidArguments()
         else -> {
             val (basePath, miniAppInfo) = miniAppDownloader.getMiniApp(appId)
+            val versionId = fetchInfo(appId).version.versionId
```
corycaywood, line 64: I think you don't need to use `fetchInfo` here. You already have `miniAppInfo` from above, so you can just use the `version.versionId` from that.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -47,6 +61,9 @@ internal class RealMiniApp(
...
+            checkWithManifest(appId, versionId)
```
corycaywood, line 65: There's still some duplicate code in the `create` functions. I think you didn't move all of it because of the `versionId` above? See my comment above above `versionId`, and then I think you can share almost all of the code in a single function.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/DownloadedManifestCache.kt
```diff
@@ -0,0 +1,108 @@
...
+            null
+        }
+    }
+
+    fun storeDownloadedManifest(
+        miniAppId: String,
+        cachedManifest: CachedManifest
+    ) {
+        val jsonToStore: String = Gson().toJson(cachedManifest)
+        prefs.edit().putString(primaryKey(miniAppId, cachedManifest.versionId), jsonToStore).apply()
```
corycaywood, line 48: We already have the `versionId` in the `CachedManifest` object, so why do you need set the version as part of the key?

I think we should store the manifest using only the `miniAppId` as the key. We only want to store one manifest for each mini app id, and it should be the manifest for the currently downloaded version. Also, this is causing problems because it means you need to pass the `versionId` every time you retrieve the manifest, but the `versionId` isn't relevant when retrieving the manifest.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -113,11 +138,36 @@ internal class RealMiniApp(
...
+    }
+
+    private suspend fun updateManifestFromApi(appId: String, versionId: String) {
+        val cachedManifest = downloadedManifestCache.readDownloadedManifest(appId, versionId)
+        if (cachedManifest?.versionId != versionId) {
+            val apiManifest = getMiniAppManifest(appId, versionId)
+            downloadedManifestCache.storeDownloadedManifest(appId, CachedManifest(versionId, apiManifest))
+            val customPermissions = miniAppCustomPermissionCache.readPermissions(appId)
+            val manifestPermissions = downloadedManifestCache.getAllPermissions(customPermissions, versionId)
+            miniAppCustomPermissionCache.removePermissionsNotMatching(appId, manifestPermissions)
```
corycaywood, line 158: It seems this block is unrelated to `updateManifestFromApi`? This is about removing cached permissions - and actually I think we want to do this even if `cachedManifest?.versionId != versionId` is NOT true. This is because the Host App could have set permissions for an already cached mini app.

So maybe you should move this block to `checkWithManifest` just below the call to `updateManifestFromApi`?

corycaywood (commented):
It's looking really close. I left a few more comments below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/DownloadedManifestCache.kt
khairul-alam-licon: Right, using `versionId` in this case causes issues. I have updated this part.

corycaywood (approved):
LGTM 👍

miniapp/USERGUIDE.md
```diff
@@ -92,15 +92,16 @@ The SDK is configured via `meta-data` tags in your `AndroidManifest.xml`. The fo
 ### #3 Create and display a Mini App
 **API Docs:** [MiniApp.create](api/com.rakuten.tech.mobile.miniapp/-mini-app/create.html), [MiniAppDisplay](api/com.rakuten.tech.mobile.miniapp/-mini-app-display/), [MiniAppMessageBridge](api/com.rakuten.tech.mobile.miniapp.js/-mini-app-message-bridge)
 
-`MiniApp.create` is used to create a `View` for displaying a specific mini app. You must provide the mini app ID which you wish to create (you can get the mini app ID by [Fetching Mini App Info](#fetching-mini-app-info) first). Calling `MiniApp.create` will do the following:
+`MiniApp.create` is used to create a `View` for displaying a specific mini app. Before calling `MiniApp.create`, the Host App should first get the manifest using `MiniApp.getMiniAppManifest`, show permission prompt to user, then set the result with `MiniApp.setCustomPermissions`.
```
minh-rakuten, line 95: We should also mention that if Host App downloads the miniapp without granted required permission, our SDK will throw exception to notify Host App.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -113,11 +124,30 @@ internal class RealMiniApp(
...
+            val apiManifest = getMiniAppManifest(appId, versionId)
+            downloadedManifestCache.storeDownloadedManifest(appId, CachedManifest(versionId, apiManifest))
+        }
+
+        val customPermissions = miniAppCustomPermissionCache.readPermissions(appId)
+        val manifestPermissions = downloadedManifestCache.getAllPermissions(customPermissions)
+        miniAppCustomPermissionCache.removePermissionsNotMatching(appId, manifestPermissions)
+
+        if (downloadedManifestCache.isRequiredPermissionDenied(customPermissions))
+            throw MiniAppSdkException(ERR_REQUIRED_PERMISSION_DENIED)
```
minh-rakuten, line 139: We can create the indicated permission for this exception `MiniAppPermissionException` so Host App can catch it and show the permission dialog again.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppWindow.kt
```diff
@@ -49,8 +49,8 @@ class PreloadMiniAppWindow(private val context: Context, private val preloadMini
 
         if (!isAccepted()) {
             launchScreen()
-        } else if (!doesDataExist()) {
-            storeAcceptance(DEFAULT_ACCEPTANCE) // should be false only after accept.
+        } else if (!prefs.contains(miniAppId)) {
```
minh-rakuten, line 52: What is the difference of those statements`if (!isAccepted())` and `else if (!prefs.contains(miniAppId))`?

minh-rakuten (approved):
Looks good. Please see my below suggestion for 2505.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppWindow.kt
khairul-alam-licon: Right, yeah, I'll change this part in 2505 ticket.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
khairul-alam-licon: Yeah, it can be an improvement task.

PR #271, opened 2021-03-10
feat: compare new manifest to display first time screen

khairul-alam-licon:
# Description
- Compare downloadedManifest and the latestManifest. If any changes found in requiredPermissions, then display first-time screen to the user to agree the latest manifest.
- Added a SDK bugfix for MINI-2897

## Links
- MINI-2505, MINI-2897, MINI-2905

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -2,13 +2,14 @@ package com.rakuten.tech.mobile.miniapp
 
 import android.util.Log
 import androidx.annotation.VisibleForTesting
+import com.rakuten.tech.mobile.miniapp.api.MetadataPermissionObj
 import com.rakuten.tech.mobile.miniapp.api.ApiClient
+import com.rakuten.tech.mobile.miniapp.api.ManifestApiCache
 import com.rakuten.tech.mobile.miniapp.api.ManifestEntity
 import com.rakuten.tech.mobile.miniapp.api.MetadataEntity
 import com.rakuten.tech.mobile.miniapp.api.UpdatableApiClient
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
 import com.rakuten.tech.mobile.miniapp.storage.CachedMiniAppVerifier
-import com.rakuten.tech.mobile.miniapp.api.ManifestApiCache
 import com.rakuten.tech.mobile.miniapp.storage.MiniAppStatus
 import com.rakuten.tech.mobile.miniapp.storage.MiniAppStorage
 import kotlinx.coroutines.CoroutineDispatcher
@@ -147,25 +148,21 @@ internal class MiniAppDownloader(
 
     @VisibleForTesting
     fun prepareMiniAppManifest(metadataEntity: MetadataEntity): MiniAppManifest {
-        val requiredPermissions = metadataEntity.metadata?.requiredPermissions?.map {
-            Pair(MiniAppCustomPermissionType.getValue(it.name ?: ""), it.reason ?: "")
-        } ?: emptyList()
-
-        val optionalPermissions = metadataEntity.metadata?.optionalPermissions?.map {
-            Pair(MiniAppCustomPermissionType.getValue(it.name ?: ""), it.reason ?: "")
-        } ?: emptyList()
-
-        val requiredFilteredValues = requiredPermissions.toMutableList()
-        requiredFilteredValues.removeAll { (first) ->
-            first.type == MiniAppCustomPermissionType.UNKNOWN.type
-        }
-        val optionalFilteredValues = optionalPermissions.toMutableList()
-        optionalFilteredValues.removeAll { (first) ->
-            first.type == MiniAppCustomPermissionType.UNKNOWN.type
-        }
-
+        val requiredPermissions = getList(metadataEntity.metadata?.requiredPermissions ?: emptyList())
+        val optionalPermissions = getList(metadataEntity.metadata?.optionalPermissions ?: emptyList())
         val customMetadata = metadataEntity.metadata?.customMetaData ?: emptyMap()
-        return MiniAppManifest(requiredFilteredValues, optionalFilteredValues, customMetadata)
+
+        return MiniAppManifest(requiredPermissions, optionalPermissions, customMetadata)
+    }
+
+    private fun getList(permissions: List<MetadataPermissionObj>): List<Pair<MiniAppCustomPermissionType, String>> {
+        val pairs = ArrayList<Pair<MiniAppCustomPermissionType, String>>()
+        permissions.forEach {
+            val permissionType = MiniAppCustomPermissionType.getValue(it.name ?: "")
+            if (permissionType.type != MiniAppCustomPermissionType.UNKNOWN.type)
+               pairs.add(Pair(permissionType, it.reason ?: ""))
+        }
+        return pairs
     }
 
     @SuppressWarnings("LongMethod")
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -148,19 +148,19 @@ internal class MiniAppDownloader(
 
     @VisibleForTesting
     fun prepareMiniAppManifest(metadataEntity: MetadataEntity): MiniAppManifest {
-        val requiredPermissions = getList(metadataEntity.metadata?.requiredPermissions ?: emptyList())
-        val optionalPermissions = getList(metadataEntity.metadata?.optionalPermissions ?: emptyList())
+        val requiredPermissions = listOfPermissions(metadataEntity.metadata?.requiredPermissions ?: emptyList())
+        val optionalPermissions = listOfPermissions(metadataEntity.metadata?.optionalPermissions ?: emptyList())
         val customMetadata = metadataEntity.metadata?.customMetaData ?: emptyMap()
 
         return MiniAppManifest(requiredPermissions, optionalPermissions, customMetadata)
     }
 
-    private fun getList(permissions: List<MetadataPermissionObj>): List<Pair<MiniAppCustomPermissionType, String>> {
+    private fun listOfPermissions(permissions: List<MetadataPermissionObj>): List<Pair<MiniAppCustomPermissionType, String>> {
         val pairs = ArrayList<Pair<MiniAppCustomPermissionType, String>>()
         permissions.forEach {
             val permissionType = MiniAppCustomPermissionType.getValue(it.name ?: "")
             if (permissionType.type != MiniAppCustomPermissionType.UNKNOWN.type)
-               pairs.add(Pair(permissionType, it.reason ?: ""))
+                pairs.add(Pair(permissionType, it.reason ?: ""))
         }
         return pairs
     }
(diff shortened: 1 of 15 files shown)


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppWindow.kt
```diff
@@ -155,15 +129,48 @@ class PreloadMiniAppWindow(
         }
     }
 
-    private fun isAccepted(): Boolean {
-        try {
-            if (prefs.contains(miniAppId)) return prefs.getBoolean(miniAppId, DEFAULT_ACCEPTANCE)
-        } catch (e: Exception) {
-            return false
+    private fun isAcceptedRequiredPermissions(manifest: MiniAppManifest): Boolean {
```
corycaywood, line 132: The SDK throws an `Exception` when the required permissions aren't granted, so couldn't you just catch that `Exception` and display the first-time launch screen? (I think this is what iOS is doing) I'm not sure you need to manually check the required permissions here.

By the way, is there a specific `Exception` type for this case where required permissions aren't granted? If not, I think we should add one (similar to iOS) like `RequiredPermissionsNotGrantedException`.

corycaywood (commented):
Looks good - just one comment below.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppWindow.kt
khairul-alam-licon: > The SDK throws an Exception when the required permissions aren't granted, so couldn't you just catch that Exception and display the first-time launch screen? (I think this is what iOS is doing) I'm not sure you need to manually check the required permissions here.

Yeah, I was about to start like this, but I realized there is something refactoring I need to figure out, so I have check it manually without catching the exception.

> By the way, is there a specific Exception type for this case where required permissions aren't granted? If not, I think we should add one (similar to iOS) like RequiredPermissionsNotGrantedException.

Updated.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
```diff
@@ -148,15 +148,24 @@ internal class MiniAppDownloader(
...
 
+        val requiredFilteredValues = requiredPermissions.toMutableList()
+        requiredFilteredValues.removeAll { (first) ->
+            first.type == MiniAppCustomPermissionType.UNKNOWN.type
+        }
+        val optionalFilteredValues = optionalPermissions.toMutableList()
+        optionalFilteredValues.removeAll { (first) ->
+            first.type == MiniAppCustomPermissionType.UNKNOWN.type
+        }
+
```
minh-rakuten, line 166: I think you can do a check `it.name` null or not before adding item to a list -> we don't need to do filter the values.

testapp/src/main/res/layout/window_preload_miniapp.xml
```diff
@@ -58,33 +58,37 @@
...
     </LinearLayout>
 
     <LinearLayout
       android:id="@+id/lnAction"
       android:layout_width="match_parent"
       android:layout_height="wrap_content"
       android:layout_alignParentBottom="true"
-      android:padding="@dimen/activity_vertical_margin"
+      android:paddingTop="@dimen/small_8"
+      android:paddingBottom="@dimen/small_8"
```
minh-rakuten, line 69: I think 16dp would make the UI less cramped.

minh-rakuten (commented):
I notice that the dialog is shown even there is no new permission from new miniapp version.

corycaywood (approved):
Looks good! 👍

khairul-alam-licon:
> I notice that the dialog is shown even there is no new permission from new miniapp version.

Yeah, basically the first time screen can be displayed when "reqPermissions" has changed (permissions were added or removed or "reason" changed), same for the "optPermissions"  and metadata information. Although, we will may have more UI improvement in the future e.g. adding a "New" level, then the confusion can be solved.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppWindow.kt
```diff
@@ -103,32 +96,13 @@ class PreloadMiniAppWindow(
...
-                                    false,
-                                    it.second
-                                )
-                                manifestPermissions.add(permission)
-                            }
-
-                            permissionAdapter.addManifestPermissionList(manifestPermissions)
+                        Observer { apiManifest ->
+                            val downloadedManifest = viewModel.miniApp.getDownloadedManifest(miniAppId)
+                            if ((downloadedManifest == apiManifest) && isAcceptedRequiredPermissions(apiManifest))
```
minh-rakuten, line 102: `downloadedManifest` can never be equal to `apiManifest` because this is object operator comparision. You have to declare the function to check whether they are same in `MiniAppManifest`.

minh-rakuten:
> Yeah, basically the first time screen can be displayed when "reqPermissions" has changed (permissions were added or removed or "reason" changed), same for the "optPermissions" and metadata information. Although, we will may have more UI improvement in the future e.g. adding a "New" level, then the confusion can be solved.

Actually, I mean this [one](https://github.com/rakutentech/android-miniapp/pull/271#discussion_r592138696). We need to check the difference of `MiniAppManifest`.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/preload/PreloadMiniAppWindow.kt
khairul-alam-licon: I think `downloadedManifest == apiManifest` is enough to meet our requirement for now, we don't have to check by individual properties. Can you check with a single version e.g. launch the miniapp for a single version id > then "Accept" > then re-launch the same version's miniapp again -> it will not display the screen again because of the equality.

khairul-alam-licon: And, I think, this is the benefit from Kotlin: https://kotlinlang.org/docs/operator-overloading.html#equality-and-inequality-operators which can be overridden to provide custom equality check implementation.

khairul-alam-licon: There is might be a confusion because some permission names are defined as "optional" in one version, but as "required" in another version. 

e.g. 

Profile Photo is optional in poc version 1.8.3
Profile Photo is required in poc version 1.8.0

So, the screen should be displayed when `1.8.3` & `1.8.0` are launching one after another.

minh-rakuten: Ah thank you for your explaination. We can close this one 👍.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
khairul-alam-licon: I think, we can do the null check using `?:` and I am filtering the unknown values to eliminate as those are unnecessary.

minh-rakuten: I have created the [PR](https://github.com/khairul-alam-licon/android-miniapp/pull/5) for this. Hope it help.

khairul-alam-licon: Thank you for the PR to understand the point of views:
Here are my feedback [on the PR](https://github.com/khairul-alam-licon/android-miniapp/pull/5/files)

**SDK: MiniAppDownloader.prepareMiniAppManifest**

- `map` & `forEach` usually iterates over all the items, the significant difference is `map` returns a list containing the results. On the other hand, `forEach` requires to performs the operation on each item by walking inside the loop. I was using `map` twice & your PR is also using `forEach` twice by calling `getList`, no? It means, the operation are still the same. Moreover, `map` is chainable e.g. we can attach other operations like `flatMap`, `filter`, `groupBy` ... , so it's good to use if possible.
- Because of using `getList` now we have two functions, but we can make a single function using `map` like what I am already doing. 

**Sample app: PreloadMiniAppWindow**
- Thank you for pointing that we don't need `SharedPreferences` to store acceptance. I want to adopt this change.

**Sample app: PreloadMiniAppViewModel**
- I don't know what does it mean by `_requiredShowManifest`, we don't have any concept like that, rather we have "required permission" concept. 
- The placements of some of the functions into ViewModel is okay, maybe I want to discuss about the architecture pattern in a separate session, but I want to mention that in future, we will have more UI improvement, then we will need to refactor e.g. in some usecases, we'll may need to trigger something to launch the screen from the viewmodel according to the changes from your PR. 

**Another discussion for clarity**, [not related to the PR](https://github.com/khairul-alam-licon/android-miniapp/pull/5/files).
- This feature will not work without valid `versionId`, it means we don't need to initiate this screen from `MiniAppInputActivity` anymore, rather we can now move it to `MiniAppDisplayActivity` because the flow has been changed inside `MiniApp.create`. After moving it to `MiniAppDisplayActivity`, we can finally utilize `RequiredPermissionsNotGrantedException` and then we will no need to use `isAcceptedRequiredPermissions` function what I introduced. 

After considering those things, I feel that we can make a separate PR later to address all of these key issues. I want to avoid this time since we are having a tight schedule. 

@minh-rakuten @corycaywood What do you think?

minh-rakuten: 1/ SDK: MiniAppDownloader.prepareMiniAppManifest

Does the refactor do the same as this current implementation? Which one gives better time and space complexity? Two functions here but it's shorter and gives better performance.

2/ PreloadMiniAppWindow & PreloadMiniAppViewModel
I'm trying to correct the function of viewmodel and view here. `_requiredShowManifest` is the purpose of this screen. You mentioned that we show this dialog again when there are changes in manifest, not only required permissions.

khairul-alam-licon: I think that `_miniAppManifest` is enough to express the purpose instead of using `_requiredShowManifest`. But as long as it doesn't break the functionality its ok

khairul-alam-licon: I think that `_miniAppManifest` is enough to express the purpose instead of using `_requiredShowManifest`. But as long as it doesn't break the functionality its okay.

khairul-alam-licon: @minh-rakuten Thanks for the discussions and affords, I have merged your changes, I have validated it, it works well 👍

testapp/src/main/res/values/strings.xml
```diff
@@ -25,7 +25,7 @@
     <string name="userdata_warning_not_saved_contact">No contact has been saved yet</string>
 
     <string name="miniapp_preload_tutorial">Would you like to download this miniapp?</string>
-    <string name="miniapp_preload_perm_status_required">(required)</string>
+    <string name="miniapp_preload_perm_status_required">(Required)</string>
```
khairul-alam-licon, line 28: I have changed here as its expected in MINI-2905.

minh-rakuten (approved):
👍 👍

PR #272, opened 2021-03-12
allow mini app in the bridge to request only permissions that is defined in manifest

khairul-alam-licon:
# Description
- Ignore/filter to not request custom permissions if it is not mentioned in manifest in `CustomPermissionBridgeDispatcher`.
- MINI-2934

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppDownloader.kt
@@ -23,7 +23,7 @@ import java.net.ConnectException
 import java.net.HttpURLConnection
 import java.net.URL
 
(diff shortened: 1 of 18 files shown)


corycaywood (commented):
It's looking good for the permissions passed to the Host App. But there's one thing missing: we need to pass back to the Mini App the same permissions list that was passed to the SDK.

So if they pass a permission which isn't in the manifest, then we should remove that permission when passing to the Host App (as you've already done), but then we also need to append that permission as "DENIED" to the list which we pass back to the Mini App.

corycaywood (commented):
LGTM 👍  Just need to fix the `config` submodule. Looks like it was reverted to the wrong commit.

corycaywood (approved):

minh-rakuten (approved):
👍

PR #276, opened 2021-03-17
feat: retrieve audience & scope from manifest

minh-rakuten:
# Description
- Check audience/scope in the manifest.
- Return the AccessTokenPermission to miniapp.
- Refactor the usage of lifecycle observer in sample app to avoid memory leak.

## Links
MINI-2680

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
@@ -1,16 +1,20 @@
 package com.rakuten.tech.mobile.miniapp
 
 import androidx.annotation.Keep
+import com.rakuten.tech.mobile.miniapp.permission.AccessTokenPermission
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
 
 /**
  * A data class to represent data in the mini app's manifest.
+ * @property requiredPermissions List of required permissions requested by Mini App.
+ * @property optionalPermissions List of required permissions requested by Mini App.
+ * @property accessTokenPermissions List of audiences and scopes requested by Mini App.
+ * @property customMetaData Custom metadata set by Mini App.
  */
 @Keep
 data class MiniAppManifest(
-    // List of permissions requested by Mini App in their manifest
     val requiredPermissions: List<Pair<MiniAppCustomPermissionType, String>>,
     val optionalPermissions: List<Pair<MiniAppCustomPermissionType, String>>,
-    // custom metadata set by Mini App in their manifest
+    val accessTokenPermissions: List<AccessTokenPermission>,
     val customMetaData: Map<String, String>
 )
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
@@ -7,7 +7,7 @@ import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
 /**
  * A data class to represent data in the mini app's manifest.
  * @property requiredPermissions List of required permissions requested by Mini App.
- * @property optionalPermissions List of required permissions requested by Mini App.
+ * @property optionalPermissions List of optional permissions requested by Mini App.
  * @property accessTokenPermissions List of audiences and scopes requested by Mini App.
  * @property customMetaData Custom metadata set by Mini App.
  */
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
@@ -1,7 +1,7 @@
 package com.rakuten.tech.mobile.miniapp
 
 import androidx.annotation.Keep
-import com.rakuten.tech.mobile.miniapp.permission.AccessTokenPermission
+import com.rakuten.tech.mobile.miniapp.permission.AccessTokenScope
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
 
 /**
@@ -15,6 +15,6 @@ import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
 data class MiniAppManifest(
     val requiredPermissions: List<Pair<MiniAppCustomPermissionType, String>>,
     val optionalPermissions: List<Pair<MiniAppCustomPermissionType, String>>,
-    val accessTokenPermissions: List<AccessTokenPermission>,
+    val accessTokenPermissions: List<AccessTokenScope>,
     val customMetaData: Map<String, String>
 )
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/TokenData.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/TokenData.kt
@@ -8,5 +8,5 @@ import com.rakuten.tech.mobile.miniapp.permission.AccessTokenScope
 data class TokenData(
     val token: String,
     val validUntil: Long,
-    var accessTokenScope: AccessTokenScope = AccessTokenScope()
+    var scopes: AccessTokenScope = AccessTokenScope()
 )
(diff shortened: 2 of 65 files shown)


corycaywood (commented):
Looking good 👍  Just a couple comments to align with iOS.

Also we should probably update the JS Bridge module before merging this (But not sure if you or @Climbatize is working on that?)

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/AccessTokenPermission.kt
```diff
@@ -8,8 +8,9 @@ import androidx.annotation.Keep
  *  @property scopes List of areas that token can access.
  */
 @Keep
-data class AccessTokenPermission(val audience: String, val scopes: List<String>) {
-
-//    fun contains(accessTokenPermission: AccessTokenPermission): Boolean =
-//        this.audience == accessTokenPermission.audience && this.scopes.containsAll(accessTokenPermission.scopes)
+data class AccessTokenPermission(
```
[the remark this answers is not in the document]
minh-rakuten: name convention updated.

corycaywood (approved):
LGTM 👍 Just need to wait for the JS Bridge updates.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -145,6 +164,7 @@ internal class UserInfoBridge {
         const val ERR_GET_ACCESS_TOKEN = "Cannot get access token:"
         const val ERR_ACCESS_TOKEN_NO_PERMISSION =
             "Permission has not been accepted yet for getting access token."
+        const val ERR_ACCESS_TOKEN_NOT_MATCH_MANIFEST = "Not match the audience and scope in manifest."
```
khairul-alam-licon, line 167: Maybe we can write: "The audience and scopes have not been matched in the manifest."

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -92,28 +96,43 @@ internal class UserInfoBridge {
...
+        } else
+            bridgeExecutor.postError(callbackObj.id, "$ERR_GET_ACCESS_TOKEN $ERR_ACCESS_TOKEN_NOT_MATCH_MANIFEST")
+    }
+
+    private fun doesMatchManifest(tokenScope: AccessTokenScope): Boolean {
+        if (tokenScope.scopes.isNotEmpty()) {
+            val atpList = downloadedManifestCache.getAccessTokenPermissions(miniAppId)
+            atpList.forEach {
+                if (it.audience == tokenScope.audience && it.scopes.containsAll(tokenScope.scopes))
+                    return true
```
khairul-alam-licon, line 126: Using `forEach { }` will let you to walk through every iteration, I have a feelings that we'll be lucky if we find the match at the first iteration and unlucky when it will find the match at the last. 

Instead we may do something using `find { }` e.g.

```
val cachedAccessTokens = downloadedManifestCache.getAccessTokenPermissions(miniAppId)
cachedAccessTokens.find {
   it.audience == tokenScope.audience && it.scopes.containsAll(tokenScope.scopes)
}?.let { return true }
```

But, anyway this is not the ideal time for tracking the performance issue, so you can consider this comment as optional.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -92,28 +96,43 @@ internal class UserInfoBridge {
...
+                if (e.message == "The `UserInfoBridgeDispatcher.getAccessToken` $NO_IMPL")
+                    userInfoBridgeDispatcher.getAccessToken(miniAppId, successCallback, errorCallback)
+                else
+                    throw e
+            }
+        } else
+            bridgeExecutor.postError(callbackObj.id, "$ERR_GET_ACCESS_TOKEN $ERR_ACCESS_TOKEN_NOT_MATCH_MANIFEST")
+    }
+
+    private fun doesMatchManifest(tokenScope: AccessTokenScope): Boolean {
```
khairul-alam-licon, line 121: Do we have any plan to match the entire manifest here in future? if no, then maybe we can rename it to more specific e.g. `doesAccessTokenMatch` or something like that..

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -92,28 +96,43 @@ internal class UserInfoBridge {
...
-                userInfoBridgeDispatcher.getAccessToken(miniAppId, successCallback, errorCallback)
-            else
-                throw e
+            try {
+                userInfoBridgeDispatcher.getAccessToken(miniAppId, tokenPermission, successCallback, errorCallback)
+            } catch (e: MiniAppSdkException) {
+                if (e.message == "The `UserInfoBridgeDispatcher.getAccessToken` $NO_IMPL")
+                    userInfoBridgeDispatcher.getAccessToken(miniAppId, successCallback, errorCallback)
+                else
+                    throw e
```
khairul-alam-licon, line 115: I am not sure, if we can get any other `MiniAppSdkException` message except `$NO_IMPL` one inside this `try {...} catch` block.  So, maybe we don't need to `throw e`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/TokenData.kt
```diff
@@ -1,10 +1,12 @@
...
 import androidx.annotation.Keep
+import com.rakuten.tech.mobile.miniapp.permission.AccessTokenScope
 
 /** Access token object for miniapp. */
 @Keep
 data class TokenData(
     val token: String,
-    val validUntil: Long
+    val validUntil: Long,
+    var accessTokenScope: AccessTokenScope = AccessTokenScope()
```
khairul-alam-licon, line 11: Is there any reason to initiate here a new `AccessTokenScope()` object?

khairul-alam-licon (commented):
Looks good, I have few minor comments...

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifest.kt
```diff
@@ -1,16 +1,20 @@
 package com.rakuten.tech.mobile.miniapp
 
 import androidx.annotation.Keep
+import com.rakuten.tech.mobile.miniapp.permission.AccessTokenScope
 import com.rakuten.tech.mobile.miniapp.permission.MiniAppCustomPermissionType
 
 /**
  * A data class to represent data in the mini app's manifest.
+ * @property requiredPermissions List of required permissions requested by Mini App.
+ * @property optionalPermissions List of required permissions requested by Mini App.
```
khairul-alam-licon, line 10: We can change to `List of optional permissions`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
minh-rakuten: it's not true, when you're using `find { }` then there is the possibility that it must go through every element in the list to get an item that you want. Furthermore, when you use find then you return an item value, it's a redundant step.

I can change into a normal loop if it makes you feel less iterated action.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
minh-rakuten: there is still chance that there is another exception, that's why I also have a test case for this.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/TokenData.kt
minh-rakuten: to not break previous sdk version on `TokenData`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
khairul-alam-licon: Ok, if that so, I think, you don't need to change here anything for now.

khairul-alam-licon (approved):
Looks good 👍

PR #277, opened 2021-03-17
feat: implement file picker capability

khairul-alam-licon:
# Description
- Allow Mini Apps to access files from the device using a [file input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file). The JS SDK Sample App already has this functionality on the "File Upload" screen, so you can use the sample app to verify.
Android has `onShowFileChooser` which can be used for this purpose.

## Links
- MINI-2170

# Checklist


miniapp/USERGUIDE.md
```diff
@@ -622,6 +622,44 @@ mini app scheme and should close external webview.
...
+- Then, HostApp activity can use `filePathHostCallback` to receive the data per `filePickingReqCode` as following:
+
+```kotlin
+override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+    super.onActivityResult(requestCode, resultCode, data)
+
+    if (requestCode == filePickingReqCode && resultCode == Activity.RESULT_OK) {
+        data?.let { intent ->
+            val result: Uri? = intent.data
+            filePathHostCallback?.onReceiveValue(arrayOf(result!!))
```
khairul-alam-licon, line 657: When HostApp will call `filePathHostCallback?.onReceiveValue(null)` in order to do kind of "resetting" the callback, it will be crashed if there is no exception handling in HostApp. I am not sure about this standard / idea how to mitigate this risk.

miniapp/USERGUIDE.md
```diff
@@ -622,6 +622,44 @@ mini app scheme and should close external webview.
...
+var filePathHostCallback: ValueCallback<Array<Uri>>? = null
+val filePickingReqCode = REQUEST_CODE
+
+miniAppFilePicker = object : MiniAppFilePicker {
+    override fun requestFile(
+        filePathCallback: ValueCallback<Array<Uri>>?,
+        callback: (requestCode: Int) -> Unit
+    ) {
+        filePathHostCallback = filePathCallback
+        callback.invoke(filePickingReqCode)
```
minh-rakuten, line 643: from what I understand is Hostapp can ignore this callback when they want to implement their own file request so you can document about it too.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -163,4 +164,21 @@ internal class MiniAppWebChromeClient(
...
+        filePathCallback: ValueCallback<Array<Uri>>?,
+        fileChooserParams: FileChooserParams?
+    ): Boolean {
+        try {
+            miniAppFilePicker?.requestFile(filePathCallback) { requestCode ->
+                val intent = fileChooserParams?.createIntent()
+                (context as Activity).startActivityForResult(intent, requestCode)
+            }
+        } catch (e: Exception) {
+            throw (MiniAppFilePickingException(e.message.toString()))
```
minh-rakuten, line 180: Is there any chance that we could get this exception?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -230,6 +249,13 @@ class MiniAppDisplayActivity : BaseActivity() {
         super.onActivityResult(requestCode, resultCode, data)
         if (requestCode == externalWebViewReqCode && resultCode == Activity.RESULT_OK) {
             data?.let { intent -> sampleWebViewExternalResultHandler.emitResult(intent) }
+        } else if (requestCode == filePickingReqCode && resultCode == Activity.RESULT_OK) {
+            data?.let { intent ->
+                if (filePathHostCallback == null) return
+
+                val result: Uri? = intent.data
+                filePathHostCallback?.onReceiveValue(arrayOf(result!!))
```
minh-rakuten, line 257: I think you're using `filePathHostCallback?` so you can remove `if (filePathHostCallback == null) return`
-> I'm not sure so let's test it 🙇‍♂️

minh-rakuten (commented):
Looks really good to me.

@khairul-alam-licon @corycaywood Can we do anything with default implementation for file picker?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
khairul-alam-licon: One chance is `ActivityNotFoundException`.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
khairul-alam-licon: I think, you are right..

miniapp/USERGUIDE.md
khairul-alam-licon: Yeah, I have wrote about it in the user guide.

miniapp/USERGUIDE.md
```diff
@@ -622,6 +622,44 @@ mini app scheme and should close external webview.
...
+
+miniAppFilePicker = object : MiniAppFilePicker {
+    override fun requestFile(
+        filePathCallback: ValueCallback<Array<Uri>>?,
+        callback: (requestCode: Int) -> Unit
+    ) {
+        filePathHostCallback = filePathCallback
+        callback.invoke(filePickingReqCode) // invoke the request code what is defined by HostApp.
+    }
+}
```
corycaywood, line 645: I think the implementation could be simplified for the Host App: it doesn't seem like it's necessary to save the `filePathCallback` within the Host App. This could be contained within `MiniAppFilePicker`. For example, something like this:

```
val filePickingReqCode = REQUEST_CODE 
val miniAppFilePicker = MiniAppFilePicker(requestCode = filePickingReqCode)

override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
    super.onActivityResult(requestCode, resultCode, data)
    if (requestCode == filePickingReqCode && resultCode == Activity.RESULT_OK) {
        data?.let { intent ->
            val result: Uri? = intent.data
            miniAppFilePicker.onReceivedFile(arrayOf(result) ?: emptyArray())
        }
    }
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFilePicker.kt
```diff
@@ -0,0 +1,20 @@
...
+interface MiniAppFilePicker {
+
+    /**
+     * Request to pick a file from the device within the miniapp webview per [requestCode].
+     * @param filePathCallback a file path callback to pass to HostApp has been found
+     * by [MiniAppWebChromeClient.onShowFileChooser].
+     * @param callback to pass the request code of file picking using an intent inside sdk,
+     * which will also be used to retrieve the data using onActivityResult in the host app.
+     */
+    fun requestFile(
```
corycaywood, line 16: Related to my comment above about simplifying the integration: I don't think we need this method.

However, it would be nice if the Host App had the ability to override the `onShowFileChooser` functionality completely, so maybe we could add that here? Basically refactor your implementation from `MiniAppWebChromeClient` into this class (and then the Host App can choose to override that implementation if they want to).

```
open class MiniAppFilePicker(var requestCode: Int) {

    internal var callback: ValueCallback<Array<Uri>>? = null

    fun onShowFileChooser(
        filePathCallback: ValueCallback<Array<Uri>>?,
        fileChooserParams: WebChromeClient.FileChooserParams?,
        context: Context
    ) : Boolean {
        try {
            this.callback = filePathCallback
            val intent = fileChooserParams?.createIntent()
            (context as Activity).startActivityForResult(intent, requestCode)
        } catch (e: Exception) {
            throw (MiniAppFilePickingException(e.message.toString()))
        }
        return true
    }

    fun onReceivedFiles(files: Array<Uri>) {
        callback?.onReceiveValue(files)
    }
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
corycaywood, line 180: What happens with the Mini App/Host App when this exception is thrown? There's no place where the exception is caught, so isn't this going to crash the App?

Maybe we should just return an empty array to the mini app if there was an exception?

corycaywood (commented):
It's a good start! I left a couple comments about how we could simplify this integration for the Host App.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFilePicker.kt
khairul-alam-licon: Yeah, your idea is working well, specially `onReceivedFiles` is convenient for the HostApp, so I have updated it.

miniapp/USERGUIDE.md
khairul-alam-licon: Thanks, I have updated this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
khairul-alam-licon: `onShowFileChooser` is expecting a `Boolean`, so I am not sure if we can return an empty array, instead maybe we can do `Log.e` the error message? 🤔

corycaywood: @khairul-alam-licon Yes, I think logging the error and then return `false` is the right solution. It's better than crashing the App 😆

khairul-alam-licon: Haha, true 😄
I have updated this part accordingly...

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -163,4 +163,13 @@ internal class MiniAppWebChromeClient(
...
             onHideCustomView()
     }
+
+    override fun onShowFileChooser(
+        webView: WebView?,
+        filePathCallback: ValueCallback<Array<Uri>>?,
+        fileChooserParams: FileChooserParams?
+    ): Boolean {
+        miniAppFilePicker?.onShowFileChooser(filePathCallback, fileChooserParams, context)
+        return true
```
corycaywood, line 173: Shouldn't this return the boolean from `MiniAppFilePicker`? 
```suggestion
        return miniAppFilePicker?.onShowFileChooser(filePathCallback, fileChooserParams, context) ?: false
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFilePicker.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * HostApp can override [WebChromeClient.onShowFileChooser] to choose a file
+     * which is requested using HTML forms with 'file' input type.
+     * @param filePathCallback The callback to receive the Uris.
+     * @param fileChooserParams The file chooser parameters can be used to create an intent to choose file.
+     * @param context Activity's context from the HostApp.
+     */
+    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    fun onShowFileChooser(
```
corycaywood, line 27: I just noticed: it seems a little strange that this class is named `MiniAppFilePicker` but this method is named `onShowFileChooser`. So maybe we should either change the class name to `MiniAppFileChooser` OR change this method name to `onShowFilePicker`. What do you think?

khairul-alam-licon: Right, so I have renamed it to `MiniAppFileChooser`.

miniapp/USERGUIDE.md
```diff
@@ -622,6 +622,43 @@ mini app scheme and should close external webview.
...
+```
+
+Also, HostApp can pass the necessary parameters using `MiniAppFileChooser.onShowFileChooser` to choose a file as following:
+
+```kotlin
+val filePathCallback: ValueCallback<Array<Uri>>? = VALUE_HERE
+val fileChooserParams: WebChromeClient.FileChooserParams? = VALUE_HERE
+
+miniAppFileChooser.onShowFileChooser(filePathCallback, fileChooserParams, context)
+```
```
corycaywood, line 660: Is this something the Host App would really want to do? 

I think a better example here would be how they can override the `onShowFileChooser` functionality.

corycaywood (approved):
LGTM 👍  I'm just not sure about the `onShowFileChooser` example in the User Guide. I think we can either remove it completely or replace it with an example of how to override `onShowFileChooser`.

miniapp/USERGUIDE.md
khairul-alam-licon: Yeah, I am thinking for an example how they can override it, currently, I have removed the guideline about it.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -0,0 +1,51 @@
...
+     */
+    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    fun onShowFileChooser(
+        filePathCallback: ValueCallback<Array<Uri>>?,
+        fileChooserParams: WebChromeClient.FileChooserParams?,
+        context: Context
+    ): Boolean {
+        try {
+            this.callback = filePathCallback
+            val intent = fileChooserParams?.createIntent()
```
minh-rakuten, line 34: Unit test for `fileChooserParams` null case.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -0,0 +1,51 @@
...
+        return true
+    }
+
+    /**
+     * Receive the files from the HostApp.
+     * @param files The array of Uri to be invoked by a filePathCallback after successfully retrieved
+     * by [Activity.onActivityResult] in the HostApp.
+     */
+    fun onReceivedFiles(files: Array<Uri>) {
+        callback?.onReceiveValue(files)
```
minh-rakuten, line 49: Unit test for `callback` null case.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/display/MiniAppWebChromeClient.kt
```diff
@@ -163,4 +163,13 @@ internal class MiniAppWebChromeClient(
...
             onHideCustomView()
     }
+
+    override fun onShowFileChooser(
+        webView: WebView?,
+        filePathCallback: ValueCallback<Array<Uri>>?,
+        fileChooserParams: FileChooserParams?
+    ): Boolean {
+        return miniAppFileChooser?.onShowFileChooser(filePathCallback, fileChooserParams, context)
+            ?: false
```
minh-rakuten, line 173: We need to have test cases here too.

minh-rakuten (commented):
Only test case job left 👍

minh-rakuten (approved):
It's possible to merge but let's wait until the access token get merge first.

miniapp/USERGUIDE.md
```diff
@@ -623,6 +623,50 @@ mini app scheme and should close external webview.
 
 Using `#ExternalResultHandler.emitResult(String)` to transmit the url string to mini app view.
 
+### File choosing
+**API Docs:** [MiniAppFileChooser](api/com.rakuten.tech.mobile.miniapp.file/-mini-app-file-chooser/)
+
+The mini app is able to choose the file which is requested using HTML forms with 'file' input type whenever users press a "Select file" button.
+HostApp can use `MiniAppFileChooser` to override `onShowFileChooser` to customize file choosing mode and other options.
```
corycaywood, line 630: I think these examples are in the wrong order: Most of the time the Host App will want to use `MiniAppFileChooserDefault`. So we should put that example first and make sure it's clear that it's the recommended usage.

PR #285, opened 2021-04-01
network request retry on 500 response code

khairul-alam-licon:
# Description
Implement request retries in the SDK for the `500` response code. The SDK should retry 5 times, with a backoff multiplier of 2. For example:

Initial Request --> failed --> wait 0.5 seconds --> First retry --> failed --> wait 1 second --> Second retry --> failed --> wait 2 seconds --> etc...

## Links
- MINI-1639

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
@@ -121,11 +121,7 @@ internal class RetrofitRequestExecutor(
             if (retryCount++ < TOTAL_RETRIES) {
                 retryCount++
                 Log.d(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
-
-                // setting retrying policy about waiting time
-                val backOff = 2.0
-                val waitTime = 0.5 * backOff.pow(retryCount.toDouble())
-                delay(1000 * waitTime.toLong())
+                delay(getWaitingTime(retryCount))
                 // recall the request
                 executeRequest(call.clone(), retryCount)
             }
@@ -145,6 +141,14 @@ internal class RetrofitRequestExecutor(
             is MiniAppSdkException -> throw error
             else -> throw MiniAppSdkException(error) // when response is not Type T or malformed JSON is received
         }
+    }
+
+    @VisibleForTesting
+    internal fun getWaitingTime(retryCount: Int): Long {
+        // calculating waiting time to retry request when 500 response code
+        val backOff = 2.0
+        val waitTime = 1000 * 0.5 * backOff.pow(retryCount.toDouble())
+        return waitTime.toLong()
     }
 
     @Throws(MiniAppSdkException::class, MiniAppNotFoundException::class)
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
@@ -1,22 +1,15 @@
 package com.rakuten.tech.mobile.miniapp.api
 
+import android.util.Log
 import androidx.annotation.VisibleForTesting
 import com.google.gson.annotations.SerializedName
-import com.rakuten.tech.mobile.miniapp.MiniAppInfo
-import com.rakuten.tech.mobile.miniapp.MiniAppHasNoPublishedVersionException
-import com.rakuten.tech.mobile.miniapp.MiniAppSdkException
-import com.rakuten.tech.mobile.miniapp.MiniAppNetException
-import com.rakuten.tech.mobile.miniapp.MiniAppNotFoundException
-import com.rakuten.tech.mobile.miniapp.sdkExceptionForInternalServerError
+import com.rakuten.tech.mobile.miniapp.*
 import okhttp3.ResponseBody
-import retrofit2.Call
-import retrofit2.Converter
-import retrofit2.HttpException
-import retrofit2.Response
-import retrofit2.Retrofit
+import retrofit2.*
 import retrofit2.http.Url
 import java.net.SocketTimeoutException
 import java.net.UnknownHostException
+import kotlin.math.pow
 
 internal class ApiClient @VisibleForTesting constructor(
     retrofit: Retrofit,
@@ -50,7 +43,8 @@ internal class ApiClient @VisibleForTesting constructor(
     suspend fun list(): List<MiniAppInfo> {
         val request = appInfoApi.list(
             hostId = hostId,
-            testPath = testPath)
+            testPath = testPath
+        )
         return requestExecutor.executeRequest(request)
     }
 
@@ -59,7 +53,8 @@ internal class ApiClient @VisibleForTesting constructor(
         val request = appInfoApi.fetchInfo(
             hostId = hostId,
             miniAppId = appId,
-            testPath = testPath)
+            testPath = testPath
+        )
         val info = requestExecutor.executeRequest(request)
 
         if (info.isNotEmpty()) {
@@ -101,12 +96,32 @@ internal class RetrofitRequestExecutor(
     private val retrofit: Retrofit
 ) {
 
-    private inline fun <reified T : ErrorResponse> createErrorConvertor(retrofit: Retrofit) =
+    private var retryCount = 0
+
+    private inline fun <reified T : ErrorResponse> createErrorConverter(retrofit: Retrofit) =
         retrofit.responseBodyConverter<T>(T::class.java, arrayOfNulls<Annotation>(0))
 
     @Suppress("TooGenericExceptionCaught", "ThrowsCount")
     suspend fun <T> executeRequest(call: Call<T>): T = try {
         val response = call.execute()
+
+        // retry network request when there is 500 error code from the server
+        if (response.code() >= 500) {
+            if (retryCount++ < TOTAL_RETRIES) {
+                Log.v(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
+
+                // setting retrying policy about waiting time
+                val backOff = 2.0
+                val waitTime = 0.5 * backOff.pow(retryCount.toDouble())
+                try {
+                    Thread.sleep(waitTime.toLong())
+                } catch (interruptedException: InterruptedException) {
+                    Log.e(MINIAPP_NETWORK_RETRY, "intercept: ", interruptedException)
+                }
+                // recall the request
+                executeRequest(call.clone())
+            }
+        }
 
         when {
             response.isSuccessful -> {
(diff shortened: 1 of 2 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -101,13 +105,40 @@ internal class RetrofitRequestExecutor(
...
+        // retry network request when there is 500 error code from the server
+        if (response.code() >= 500) {
+            if (retryCount++ < TOTAL_RETRIES) {
+                Log.v(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
+
+                // setting retrying policy about waiting time
+                val backOff = 2.0
+                val waitTime = 0.5 * backOff.pow(retryCount.toDouble())
+                try {
+                    Thread.sleep(waitTime.toLong())
```
khairul-alam-licon, line 133: I am not sure, if this is the good solution for now, but raised this PR for a discussion. 
Based on the final conclusion, I'll try to improve the test coverage.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -101,13 +105,40 @@ internal class RetrofitRequestExecutor(
...
+        "NestedBlockDepth",
+        "ComplexMethod"
+    )
     suspend fun <T> executeRequest(call: Call<T>): T = try {
         val response = call.execute()
 
+        // retry network request when there is 500 error code from the server
+        if (response.code() >= 500) {
+            if (retryCount++ < TOTAL_RETRIES) {
+                Log.v(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
```
Climbatize, line 127: will you keep a verbose log?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
Climbatize: Maybe you can use a `delay` before the `executeRequest` call? I'm not a Kotlin expert though

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
khairul-alam-licon: Yeah, maybe I can use `Log.d` here? 🤔

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
khairul-alam-licon: Yeah, `delay` is a good idea!

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -101,13 +106,36 @@ internal class RetrofitRequestExecutor(
...
+        // retry network request when there is 500 error code from the server
+        if (response.code() >= 500) {
+            if (retryCount++ < TOTAL_RETRIES) {
+                Log.d(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
+
+                // setting retrying policy about waiting time
+                val backOff = 2.0
+                val waitTime = 0.5 * backOff.pow(retryCount.toDouble())
+                try {
+                    delay(waitTime.toLong())
```
corycaywood, line 130: `delay` takes the time in milliseconds, so you need to use `* 1000` here.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -101,13 +106,36 @@ internal class RetrofitRequestExecutor(
     private val retrofit: Retrofit
 ) {
 
-    private inline fun <reified T : ErrorResponse> createErrorConvertor(retrofit: Retrofit) =
+    private var retryCount = 0
```
corycaywood, line 109: This can't be a class property because it will mean that the property is shared by ALL network requests, but we want it to be specific to EACH network reqeust. Also, currently any subsequent network requests will already have `retryCount = 5`, so the network request won't even be attempted.

I think the retry count should be passed to the `executeRequest` method and give it a default value of 0. See the iOS implementation: https://github.com/rakutentech/ios-miniapp/commit/42d373a9c69ef2ab4890ea58f03a57ea9a4c8b27

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -101,13 +106,36 @@ internal class RetrofitRequestExecutor(
...
-    @Suppress("TooGenericExceptionCaught", "ThrowsCount")
+    @Suppress(
+        "TooGenericExceptionCaught", "ThrowsCount", "MagicNumber",
+        "LongMethod", "NestedBlockDepth", "ComplexMethod"
+    )
     suspend fun <T> executeRequest(call: Call<T>): T = try {
         val response = call.execute()
 
+        // retry network request when there is 500 error code from the server
+        if (response.code() >= 500) {
```
corycaywood, line 122: I think this should be `response.code() > 500` in order to match the iOS implementation.

corycaywood (changes requested):
Looking good, but there are a couple of issues that need to be addressed below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
khairul-alam-licon: Oh, I thought, I am already matching it with iOS according to [statusCode >= 500](https://github.com/rakutentech/ios-miniapp/blob/b3bedb0d71d4c6bb2540b9f7968bf7c153dab6fe/MiniApp/Classes/core/MiniAppClient.swift#L99).

corycaywood: @khairul-alam-licon Oops, please ignore this comment. Sorry about that.

I didn't look closely at this line and I was thinking it was the `retryCount` line.

khairul-alam-licon: Oh, no problem 👍 
Yeah, I am looking forward to address other feedbacks accordingly.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
khairul-alam-licon: Yeah, I have updated this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -101,13 +106,36 @@ internal class RetrofitRequestExecutor(
...
+            if (retryCount++ < TOTAL_RETRIES) {
+                Log.d(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
+
+                // setting retrying policy about waiting time
+                val backOff = 2.0
+                val waitTime = 0.5 * backOff.pow(retryCount.toDouble())
+                try {
+                    delay(waitTime.toLong())
+                } catch (interruptedException: InterruptedException) {
```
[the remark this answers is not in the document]
khairul-alam-licon: Yeah, I have updated this part.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -100,14 +105,32 @@ internal class ApiClient @VisibleForTesting constructor(
...
+        "TooGenericExceptionCaught", "ThrowsCount", "MagicNumber",
+        "LongMethod", "NestedBlockDepth", "ComplexMethod"
+    )
+    suspend fun <T> executeRequest(call: Call<T>, _retryCount: Int = 0): T = try {
         val response = call.execute()
 
+        // retry network request when there is 500 error code from the server
+        var retryCount = _retryCount
+        if (response.code() >= 500) {
+            if (retryCount++ < TOTAL_RETRIES) {
```
corycaywood, line 121: Can you add a unit test for the retry behavior? If you're using a `TestCoroutineScope` in the unit tests, then you should be able to use `advanceTimeBy` in order to test this: 
https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-test/kotlinx.coroutines.test/-delay-controller/advance-time-by.html

corycaywood (approved):
Looks good, but please try to add unit tests for this functionality.

Climbatize (approved):
LGTM

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
minh-rakuten, line 121: You can put only single line of `if` here and remove `NestedBlockDepth`, and perhaps even`ComplexMethod`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
```diff
@@ -100,14 +105,32 @@ internal class ApiClient @VisibleForTesting constructor(
...
+                retryCount++
+                Log.d(MINIAPP_NETWORK_RETRY, "Retrying in $retryCount out of $TOTAL_RETRIES times.")
```
minh-rakuten, line 123: In my opinion, I think we should remove this log for the sdk.

minh-rakuten (commented):
There is one last important thing to change is we must avoid the recursive `executeRequest` here because there is an await of response.

We can do it by extract the code from line 116 ~ 132 into different function `executeWithRetry`. Something like: ```val response = executeWithRetry(call: Call<T>, _retryCount: Int = 0)```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
khairul-alam-licon: Initially, I was about to put a verbose log, later I was following [iOS way](https://github.com/rakutentech/ios-miniapp/blob/b3bedb0d71d4c6bb2540b9f7968bf7c153dab6fe/MiniApp/Classes/core/MiniAppClient.swift#L104). But, this can be optional.

khairul-alam-licon:
> There is one last important thing to change is we must avoid the recursive executeRequest here because there is an await of response.
> We can do it by extract the code from line 116 ~ 132 into different function executeWithRetry. Something like: val response = executeWithRetry(call: Call<T>, _retryCount: Int = 0)

I have updated, please review when you will be available.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/ApiClient.kt
Climbatize: yeah, iOS way is really good, I like what their developer did

minh-rakuten (approved):
👍 👍

khairul-alam-licon:
Thanks for the reviews.

PR #302, opened 2021-05-06
Add custom permission to send messages to contact id

khairul-alam-licon:
# Description
Sending messages to a specific Contact ID via the JS SDK requires that a permission should be allowed. This PR aims to add this spec.

## Links
- MINI-3184

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
@@ -20,6 +20,7 @@ enum class MiniAppCustomPermissionType(val type: String) {
     PROFILE_PHOTO("rakuten.miniapp.user.PROFILE_PHOTO"),
     CONTACT_LIST("rakuten.miniapp.user.CONTACT_LIST"),
     ACCESS_TOKEN("rakuten.miniapp.user.ACCESS_TOKEN"),
+    SEND_MESSAGE("rakuten.miniapp.user.SEND_MESSAGE"),
     LOCATION("rakuten.miniapp.device.LOCATION"),
     UNKNOWN("UNKNOWN");
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - **Deprecated:** Old `getUniqueId`interface.
 - **Feature:** Added `getUniqueId` new interface for invoking data using `onSuccess` and `onError`.
 - **Feature:** Support `mailto` uri address. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).
+- **Feature:** Added `rakuten.miniapp.user.SEND_MESSAGE` custom permission and applied to `ChatBridgeDispatcher.sendMessageToContactId`.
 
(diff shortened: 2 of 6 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
```diff
@@ -20,6 +20,7 @@ enum class MiniAppCustomPermissionType(val type: String) {
     PROFILE_PHOTO("rakuten.miniapp.user.PROFILE_PHOTO"),
     CONTACT_LIST("rakuten.miniapp.user.CONTACT_LIST"),
     ACCESS_TOKEN("rakuten.miniapp.user.ACCESS_TOKEN"),
+    SEND_MESSAGE("rakuten.miniapp.user.SEND_MESSAGE"),
```
minh-rakuten, line 23: I set the value in JS SDK as `rakuten.miniapp.action.SEND_MESSAGE` so can we change this?

khairul-alam-licon: Ah I see, I have followed iOS which has been merged already: 
https://github.com/rakutentech/ios-miniapp/blob/5ed44f98b7328eff4ba4294ca24c9f3b99e3b9a5/MiniApp/Classes/core/JavascriptBridge/MiniAppJavaScriptEnumerations.swift#L47 

I think, sending message is related to user permission, so we can use `user` for it 🤔

minh-rakuten: Sending message is an action which is not related to object permission concept so I think `action` here is more reasonable than `user`.
@rleojoseph Can we change into `rakuten.miniapp.action.SEND_MESSAGE` for ios?

minh-rakuten (commented):
LGTM but we need to add the permission name in `PermissionUtils` in sample app.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
rleojoseph: @minh-rakuten I guess it makes sense. May be something like `rakuten.miniapp.user.action.SEND_MESSAGE` ? @corycaywood what do you think of this ?

minh-rakuten (approved):
👍 👍 
Let's wait for this [PR](https://github.com/rakutentech/js-miniapp/pull/125) and then we can update bridge submodule.

khairul-alam-licon:
> Let's wait for this PR and then we can update bridge submodule.

Yeah, also I'll update the permission string based on the decision.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppPermission.kt
khairul-alam-licon: Yeah, I have updated using `rakuten.miniapp.user.action.SEND_MESSAGE`, Thanks for the discussion.

PR #304, opened 2021-05-11
encrypt the stored MiniApp custom permissions

khairul-alam-licon:
# Description
SDK should encrypt the Mini App permissions before saving them to the cache.  Previously, custom permissions has been stored in a normal `SharedPreferences`, now we can migrate it to `EncryptedSharedPreferences`. This PR aims to fix this issue.

## Links
- MINI-1661

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
@@ -2,9 +2,13 @@ package com.rakuten.tech.mobile.miniapp.permission
 
 import android.content.Context
 import android.content.SharedPreferences
+import android.util.Log
 import androidx.annotation.VisibleForTesting
+import androidx.security.crypto.EncryptedSharedPreferences
+import androidx.security.crypto.MasterKeys
 import com.google.gson.Gson
 import com.google.gson.reflect.TypeToken
+import com.rakuten.tech.mobile.miniapp.MiniAppVerificationException
 import java.lang.Exception
 
 /**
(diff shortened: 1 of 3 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -134,3 +166,16 @@ internal class MiniAppCustomPermissionCache(context: Context) {
         return isPermissionGranted
     }
 }
+
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
+internal fun initEncryptedSharedPreference(context: Context) = try {
+    EncryptedSharedPreferences.create(
+        "com.rakuten.tech.mobile.miniapp.custom.permissions.cache.hash",
```
corycaywood, line 173: This isn't storing hashes, so maybe this should be `com.rakuten.tech.mobile.miniapp.custom.permissions.cache.encrypted`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
```diff
@@ -3,20 +3,52 @@ package com.rakuten.tech.mobile.miniapp.permission
...
+            nonEncryptedPreferences.edit().clear().apply()
+        }
+    }
+
+    private fun SharedPreferences.copyTo(dest: SharedPreferences) = with(dest.edit()) {
+        for ((key, value) in all) {
+            val v = value ?: continue
+            when (v) {
+                is String -> putString(key, v)
+                else -> error("Unknown type: $v")
```
corycaywood, line 46: I think that the Host App isn't able to catch this anywhere, so this Exception would crash the App, right? In what cases could this occur?

It seems like maybe we should just log the error instead of throwing an exception. What do you think?

corycaywood (approved):
Looks good, I'm just concerned about the exception being thrown.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
khairul-alam-licon: Yeah, I have updated the parameter.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/permission/MiniAppCustomPermissionCache.kt
khairul-alam-licon: I have updated this area.

minh-rakuten (approved):
👍

PR #305, opened 2021-05-12
[MINI-2947] verify hash of manifest before creating Mini App

khairul-alam-licon:
# Description
Store a hash of the downloaded manifest.json for each Mini App, the hash should be verified before creating the Mini App. If the hash verification fails, then the manifest.json should be re-downloaded from the API before launching the Mini App.

## Links
MINI-2947

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifestVerifier.kt
```diff
@@ -0,0 +1,72 @@
...
+                prefs.edit().putString(appId, hash).apply()
+            }
+        }
+
+    @SuppressWarnings("TooGenericExceptionCaught", "LongMethod", "NestedBlockDepth")
+    private fun calculateHash(
+        cachedManifest: CachedManifest
+    ): String = try {
+        val messageDigest = MessageDigest.getInstance("SHA-256")
+        val hashedBytes = messageDigest.digest(cachedManifest.toString().toByteArray(Charsets.UTF_8))
```
khairul-alam-licon, line 49: Currently, manifest data still has been saved in `SharedPreferenes` instead of file, so I have used `CachedManifest`. However, we have plan to migrate it using MINI-2945.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifestVerifier.kt
```diff
@@ -0,0 +1,72 @@
+package com.rakuten.tech.mobile.miniapp
```
corycaywood, line 1: It seems like the root package isn't the right place for this - usually most classes in the root are public classes. Should this go under `storage` instead?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifestVerifier.kt
```diff
@@ -0,0 +1,72 @@
...
+import androidx.security.crypto.EncryptedSharedPreferences
+import androidx.security.crypto.MasterKeys
+import com.rakuten.tech.mobile.miniapp.storage.CachedManifest
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.async
+import kotlinx.coroutines.withContext
+import java.security.MessageDigest
+
+internal class MiniAppManifestVerifier
```
corycaywood, line 17: This class is very similar to `CachedMiniAppVerifier`. I think you could share the functionality between them either by creating a base class or something like `StoredHashVerifier` which contains common functionality.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifestVerifier.kt
```diff
@@ -0,0 +1,72 @@
...
+    private val prefs: SharedPreferences,
+    private val coroutineDispatcher: CoroutineDispatcher
+) {
+    constructor(context: Context) : this(
+        prefs = initEncryptedSharedPreference(context),
+        coroutineDispatcher = Dispatchers.IO
+    )
+
+    /** Verifies that the cached data for the Mini App manifest have not been modified. */
+    fun verify(appId: String, cachedManifest: CachedManifest): Boolean {
```
corycaywood, line 28: `verify` and `storeHashAsync` don't need to accept `CachedManifest` - you could instead just accept a `ByteArray` here in order to make this more generic. It would make this functionality re-usable in the future if needed. See my comment above about sharing functionality.

corycaywood (commented):
Looking good 👍  But as mentioned below, I think there's some refactoring that could be done to reduce duplication between this and `CachedMiniAppVerifier`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifestVerifier.kt
khairul-alam-licon: Yeah, I was concerned the same, I'll update this part accordingly.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppManifestVerifier.kt
khairul-alam-licon: Yeah, I agree. In this time, we haven't migrated manifest info storage from `SharedPreferences` to `File` by MINI-2945.
Otherwise, we could have a common `calculateHash` too and it would be easier to pass `ByteArray`.

I have updated this part by adding `StoreHashVerifier`, Please have a look if it's okay. Thank you for your feedback.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/verifier/StoreHashVerifier.kt
```diff
@@ -0,0 +1,42 @@
...
+        withContext(coroutineDispatcher) {
+            async {
+                prefs.edit().putString(appId, hash).apply()
+            }
+        }
+}
+
+@Suppress("TooGenericExceptionCaught", "SwallowedException")
+internal fun initEncryptedSharedPreference(context: Context, fileName: String) = try {
+    EncryptedSharedPreferences.create(
```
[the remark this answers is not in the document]
khairul-alam-licon: Yeah, I am looking into that, Thanks for the information.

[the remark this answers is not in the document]
khairul-alam-licon: Yeah, I have updated accordingly. Thanks for the info.

munir-rakuten (approved):
looks good to me

PR #306, opened 2021-05-13
refactor: check canonical path

minh-rakuten:
# Description
Check canonical path as google security suggestion.

In my perspective, we pass the path for a zip to be extracted so I don't think there is any real issue here. The only possibility is someone can do reverse engineering and modify android runtime execution.

## Links
MINI-3204

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
@@ -1,6 +1,5 @@
 package com.rakuten.tech.mobile.miniapp.storage
 
-import com.rakuten.tech.mobile.miniapp.BuildConfig
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
@@ -20,11 +19,12 @@ internal class FileWriter(private val coroutineDispatcher: CoroutineDispatcher =
 
 internal fun ZipInputStream.decompress(zipPath: String) = use { input ->
     val outputFilePath = File(zipPath).apply { parentFile?.mkdirs() }
+    val basePath = outputFilePath.canonicalPath.substringBeforeLast(File.separator)
     var entry: ZipEntry?
 
     while (input.nextEntry.also { entry = it } != null) {
         val newFile = File(outputFilePath.parentFile, entry!!.name)
-        if (newFile.canonicalPath.contains(BuildConfig.LIBRARY_PACKAGE_NAME))
+        if (newFile.canonicalPath.startsWith(basePath))
             createData(newFile, entry!!, input)
     }
 }
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
@@ -1,5 +1,6 @@
 package com.rakuten.tech.mobile.miniapp.storage
 
+import com.rakuten.tech.mobile.miniapp.BuildConfig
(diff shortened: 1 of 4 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
```diff
@@ -23,14 +24,19 @@ internal fun ZipInputStream.decompress(zipPath: String) = use { input ->
 
     while (input.nextEntry.also { entry = it } != null) {
         val newFile = File(outputFilePath.parentFile, entry!!.name)
-        newFile.parentFile?.mkdirs()
-        if (entry!!.isDirectory)
-            newFile.mkdir()
-        else
-            writeFile(newFile, input)
+        if (newFile.canonicalPath.contains(BuildConfig.LIBRARY_PACKAGE_NAME))
```
khairul-alam-licon, line 27: Yeah, I'm not sure if it's enough to check only `BuildConfig.LIBRARY_PACKAGE_NAME`, maybe it's more safer to give more data e.g. `outputFilePath.parentFile?.canonicalPath.toString()`

khairul-alam-licon: Also, how about we could throw an exception while there is any vulnerability found, I had an idea e.g.

```
private fun checkZipVulnerability(newFile: File, outputDir: String) {
    if (!newFile.canonicalPath.startsWith(outputDir))
        throw MiniAppSdkException("A vulnerability has been found while zip path traversal with" + newFile.canonicalPath)
}

// Then we can use it as following:

    ...
    ...

    while (input.nextEntry.also { entry = it } != null) {
        ...
        try {
            checkZipVulnerability(newFile, outputFilePath.parentFile?.canonicalPath.toString())
        } catch (e: MiniAppSdkException) {
            e.printStackTrace()
            return@use
        }
        createData(newFile, entry!!, input)
    }
...
```

minh-rakuten: what is `outputDir`?

khairul-alam-licon: I have passed `outputFilePath.parentFile?.canonicalPath.toString()` as `outputDir`.
Basically, it's the `DIR` which was suggested in https://support.google.com/faqs/answer/9294009
Screen Shot 2021-05-17 at 15 06 34
We can use canonical paths to verify this issue.

minh-rakuten: ah the `newFile` is generated from the `outputDir` path so that check does not help. My check also doesn't help much, the purpose of this pr is just to remove the meaningless warning.

khairul-alam-licon (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/FileWriter.kt
corycaywood: @minh-rakuten But the value of `BuildConfig.LIBRARY_PACKAGE_NAME` is `com.rakuten.tech.mobile.miniapp`. So I think this only works in our Demo App because the package name of our demo app is `com.rakuten.tech.mobile.miniapp.testapp`.

If this ran in another App with the package name `com.example` then the output directory would be  `/data/data/com.example/files/miniapp`, so it looks like it would fail to extract the Zip.

I guess you could use `zipPath` and parse out the filename so you can have just the base directory?

minh-rakuten: updated to check start with canonical base path.

Climbatize (approved):

PR #309, opened 2021-05-18
[MINI-3131] Feature/update get access token error type

munir-rakuten:
# Description
Added Specific error types to use from host app to show in mini apps when failed to fetch access token.
Demo App Can send 

MiniAppAccessTokenError.scopesNotSupportedError
MiniAppAccessTokenError.audienceNotSupportedError
MiniAppAccessTokenError.authorizationFailureError(message = "authorization failure error msg")
MiniAppAccessTokenError.custom(message = "custom error msg")

## Links
MINI-3131

# Checklist
- [x] I wrote unit test for new code


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppErrors.kt
```diff
@@ -0,0 +1,9 @@
+package com.rakuten.tech.mobile.miniapp.errors
+
+/** Type of errors for access token fetch **/
+@Suppress("MatchingDeclarationName")
+enum class AccessTokenErrorType(val errorKey: String) {
+    AudienceNotSupportedError("AudienceNotSupportedError"),
+    ScopesNotSupportedError("ScopesNotSupportedError"),
+    AuthorizationFailureError("AuthorizationFailureError");
```
minh-rakuten, line 8: The spec of `Authorization failure` declared that it is a custom error message sent from hostapp.
If we go only with `AccessTokenErrorType.errorKey` then the hostapp user cannot pass the error message so

we can create a `MiniAppAccessTokenError` class which contains `error: AccessTokenErrorType = OtherError` and `message: String = ""`.

When user does not pass the `message` which is empty as default, we will let the errorCallback send `AccessTokenErrorType.errorKey`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -100,8 +101,27 @@ internal class UserInfoBridge {
                 accessToken.scopes = tokenPermission
                 bridgeExecutor.postValue(callbackObj.id, Gson().toJson(accessToken))
             }
-            val errorCallback = { message: String ->
-                bridgeExecutor.postError(callbackObj.id, "$ERR_GET_ACCESS_TOKEN $message")
+
+            // send a specifically formatted error key to show specific error in mini app
+            val errorCallback = { errorKey: String ->
```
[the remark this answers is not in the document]
munir-rakuten: @minh-rakuten  yes i wanted to do like this..bt it will be a major change and can break the previous implementation...

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -100,15 +102,29 @@ internal class UserInfoBridge {
...
+                bridgeExecutor.postError(callbackObj.id, returnAccessTokenError(errorType))
             }
 
             userInfoBridgeDispatcher.getAccessToken(miniAppId, tokenPermission, successCallback, errorCallback)
         } else
             bridgeExecutor.postError(callbackObj.id, "$ERR_GET_ACCESS_TOKEN $ERR_ACCESS_TOKEN_NOT_MATCH_MANIFEST")
     }
 
+    private fun returnAccessTokenError(errorType: MiniAppAccessTokenError): String {
+        return when (errorType.error) {
```
khairul-alam-licon, line 117: I thought, just returning `AccessTokenErrorType.getValue(errorType.error)` would be enough 🤔 
Or, I am not sure about the context of `AuthorizationFailureError` here.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -195,7 +196,7 @@ class MiniAppDisplayActivity : BaseActivity() {
                     miniAppId: String,
                     accessTokenScope: AccessTokenScope,
                     onSuccess: (tokenData: TokenData) -> Unit,
-                    onError: (message: String) -> Unit
```
khairul-alam-licon, line 198: Then, we need to update the user-guide too:
Screen Shot 2021-05-25 at 0 03 10

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - **Feature:** Added `getUniqueId` new interface for invoking data using `onSuccess` and `onError`.
 - **Feature:** Support `mailto` uri address. See [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a).
 - **Feature:** Added `rakuten.miniapp.user.action.SEND_MESSAGE` custom permission and applied to `ChatBridgeDispatcher.sendMessageToContactId`.
+- **Change:** Updated `getAccessToken` result with new Error type i.e `AccessTokenErrorType` to support more errors
```
khairul-alam-licon, line 9: I think, if we still keep the old one, then we can replace the word "Updated" here by "Added" with mentioning that the old one is being deprecated.

khairul-alam-licon (commented):
Looking really close! 
I have left few concerns.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
munir-rakuten: yes we need to update it @khairul-alam-licon

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
munir-rakuten: for the AuthorizationFailureError and Error type the message will be set by the host app so it needs the case @khairul-alam-licon

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -100,15 +102,29 @@ internal class UserInfoBridge {
                 accessToken.scopes = tokenPermission
                 bridgeExecutor.postValue(callbackObj.id, Gson().toJson(accessToken))
             }
-            val errorCallback = { message: String ->
-                bridgeExecutor.postError(callbackObj.id, "$ERR_GET_ACCESS_TOKEN $message")
```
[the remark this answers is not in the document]
munir-rakuten: thanks, yes it can be done like that, @corycaywood  may suggest, should we support it?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridgeDispatcher.kt
```diff
@@ -46,13 +47,35 @@ interface UserInfoBridgeDispatcher {
...
         onSuccess: (tokenData: TokenData) -> Unit,
         onError: (message: String) -> Unit
     ) = getAccessToken(miniAppId, onSuccess, onError)
 
+    /**
+     * Get access token from host app.
+     * @param accessTokenScope contains audience and scope for permission validation.
+     * @param onError contains custom error message send from host app
+     **/
+    @JvmName("getAccessToken")
```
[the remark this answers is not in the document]
munir-rakuten: yes, no need, bt i left it just for clearity... @khairul-alam-licon

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ *  Contains the components to use custom errors from host app.
+ *  @property type The type of error.
+ *  @property message error message send to min app.
+ */
+@Keep
+data class MiniAppError(
+    val type: String = "",
+    val message: String = ""
```
corycaywood, line 13: I think both of these should be nullable.
@Climbatize Can the JS SDK handle the case where one or both is null?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
```diff
@@ -0,0 +1,16 @@
...
+
+import androidx.annotation.Keep
+
+/**
+ *  Contains the components to use custom errors from host app.
+ *  @property type The type of error.
+ *  @property message error message send to min app.
+ */
+@Keep
+data class MiniAppError(
```
corycaywood, line 11: Maybe this type should be something like `MiniAppBridgeError` so it's more clear where it's used. What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
```diff
@@ -0,0 +1,16 @@
...
+    val type: String = "",
+    val message: String = ""
+) {
+    constructor() : this(type = "", message = "")
```
corycaywood, line 15: I think this constructor isn't needed for anything, right?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppErrorTypes.kt
```diff
@@ -0,0 +1,13 @@
+package com.rakuten.tech.mobile.miniapp.errors
+
+/** Enumeration that is used to return  Access Token error. **/
+enum class AccessTokenErrorType(val type: String) {
+    // Requested Audience is not supported
+    AudienceNotSupportedError("AudienceNotSupportedError"),
```
corycaywood, line 6: I'm not sure if these need to be defined as an enum. Shouldn't it be enough to define them as just strings since the `MiniAppError` accepts a string as the `type`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
corycaywood, line 11: It seems like `MiniAppError` should be internal as this is exposing the JSON format that we're using. Also, we want them to only provide the supported types of errors to `getAccessToken`, so I think we need at least three classes:

- `MiniAppBridgeErrorModel` - internal data model (same as `MiniAppError` here currently)
- `MiniAppBridgeError` - abstract public class that's used as base class for errors to pass to mini app
- `MiniAppAccessTokenError` - public class that extends `MiniAppBridgeError` and is used as the error type for `getAccessToken`. It can also have pre-defined errors in the `companion object` to construct the specific errors. (i.e. `MiniAppAccessTokenError.audienceNotSupportedError`)

Do you think that makes sense?

corycaywood (commented):
Looking good, but I think we need a few tweaks for the public interface.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
munir-rakuten: @corycaywood  yes, we can use `MiniAppBridgeError` instead of `MiniAppError`, May be JS SDK use `MiniAppError`, so i thought it may look consistent in both platform, i will update it no problem

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
munir-rakuten: yes, i will remove this constructor.. @corycaywood

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppError.kt
Climbatize: @corycaywood I managed this case so that's even fine to send `{}`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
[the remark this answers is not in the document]
Climbatize: @khairul-alam-licon @munir-rakuten I think we don't have to support the previous implementation there, as the callback is generated by the SDK and not by the host app, and is managed by the JS bridge

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
Climbatize: I think you still need to update `USERGUIDE.md` @munir-rakuten :)

Climbatize (commented):
LGTM. I'll approve the request when the JS-Miniapp PR https://github.com/rakutentech/js-miniapp/pull/130 will be validated and the submodule upgraded there, and the user guide updated 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/userinfo/UserInfoBridge.kt
```diff
@@ -100,8 +102,12 @@ internal class UserInfoBridge {
...
+
+            // send a specifically formatted error key to show specific error in mini app
+            val errorCallback = { callback: MiniAppAccessTokenError ->
+                // convert error callback to common MiniAppBridgeErrorModel
+                val errorBridgeModel = MiniAppBridgeErrorModel(callback.type, callback.message)
+                bridgeExecutor.postError(callbackObj.id, Gson().toJson(errorBridgeModel))
             }
```
Climbatize, line 111: this has to be done on line 115 where you just send a regular `String` as an error and not a `MiniAppBridgeErrorModel`

I think you will have to modify `doesAccessTokenMatch` method to return the right error type and not just a Boolean

Climbatize (changes requested):

khairul-alam-licon (approved):
Looks good 👍 
Also, need to update with JS side before merging this PR.

Climbatize (approved):

PR #316, opened 2021-05-26
[MINI-3141] additional methods to add and remove single or multiple rat credentials from host app

munir-rakuten:
# Description
When configuring the SDK instance in an app, allow for 0 or more RAT configurations to be used.

## Links
MINI-3141

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -43,9 +43,34 @@ internal class MiniAppAnalytics(val rasProjectId: String) {
...
+        listOfExternalConfig.map { config ->
+                params["acc"] = config.acc
+                params["aid"] = config.aid
+                RatTracker.event(eType.value, params).track()
+        }
     } catch (e: Exception) {
         Log.e("MiniAppAnalytics", e.message.orEmpty())
     }
+
+    /** Is add single analytic configuration. **/
```
khairul-alam-licon, line 57: Maybe we can remove 'Is' from the beginning 🤔

khairul-alam-licon (commented):
Looks good, do we need to add something how to use those functions in USERGUIDE?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,17 +17,17 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
 
 /** Only init when analytics dependency is provided. */
 @Suppress("SwallowedException", "TooGenericExceptionCaught")
-internal class MiniAppAnalytics(val rasProjectId: String) {
+class MiniAppAnalytics(private val rasProjectId: String) {
```
corycaywood, line 20: I'm not sure that you need to make this class public - usually we use `MiniApp` as our main public interface. You could leave `MiniAppAnalytics` as internal, then add new public functions in  `MiniApp.addAnalyticsConfig` and `MiniApp.removeAnalyticsConfig` which then just call `MiniAppAnalytics.instance().addAnalyticsConfig`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -43,9 +43,34 @@ internal class MiniAppAnalytics(val rasProjectId: String) {
...
         Log.e("MiniAppAnalytics", e.message.orEmpty())
     }
+
+    /** Is add single analytic configuration. **/
+    fun addAnalyticsConfig(miniAppAnalyticsConfig: MiniAppAnalyticsConfig) {
+        listOfExternalConfig.add(miniAppAnalyticsConfig)
+    }
+
+    /** Is add multiple analytic configuration. **/
+    fun addAnalyticsConfig(miniAppAnalyticsConfigs: List<MiniAppAnalyticsConfig>) {
```
corycaywood, line 63: I think we don't need to have two versions of the function. Maybe you could use `vararg` instead? `fun addAnalyticsConfig(vararg miniAppAnalyticsConfig: MiniAppAnalyticsConfig)`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -43,9 +43,34 @@ internal class MiniAppAnalytics(val rasProjectId: String) {
                 .put("mini_app_version_id", miniAppInfo.version.versionId)
         }
         params["cp"] = cp
-
         RatTracker.event(eType.value, params).track()
+        // Send to all the external acc/aid added by host app
+        listOfExternalConfig.map { config ->
```
corycaywood, line 48: Shouldn't this be `forEach` instead of `map`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -43,9 +43,34 @@ internal class MiniAppAnalytics(val rasProjectId: String) {
...
+                params["acc"] = config.acc
+                params["aid"] = config.aid
+                RatTracker.event(eType.value, params).track()
+        }
     } catch (e: Exception) {
         Log.e("MiniAppAnalytics", e.message.orEmpty())
     }
+
+    /** Is add single analytic configuration. **/
+    fun addAnalyticsConfig(miniAppAnalyticsConfig: MiniAppAnalyticsConfig) {
```
corycaywood, line 58: Please also add unit tests for the new functionality.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalyticsConfig.kt
```diff
@@ -0,0 +1,17 @@
...
+ *  @property acc The RAT account id.
+ *  @property aid The RAT app id.
+ *  @property endPoint is not in use.
+ */
+@Keep
+data class MiniAppAnalyticsConfig(
+    val acc: Int,
+    val aid: Int
+) {
+    constructor(endPoint: String, acc: Int, aid: Int) : this(acc, aid)
```
corycaywood, line 16: I don't think this constructor is used for anything, right? Also, we have no need for the `endpoint` setting. So probably this can be removed.

corycaywood (commented):
The functionality is looking good, but see my comments below. 

Also, I think you should apply this in the Demo App. You can create a new build variable (similar to the current `ACCOUNT_ACC`: https://github.com/rakutentech/android-miniapp/blob/master/testapp/build.gradle#L39), and then we can setup a second account in the CI settings.

corycaywood (commented):
@munir-rakuten @khairul-alam-licon  Sorry, looking at this now, I've realized a problem with this approach. Actually it's an existing problem with using a single `instance` of `MiniAppAnalytics`.

The Host App might be using multiple different project ids in their App (i.e. for different types of mini apps in different sections of the App). So in that case, they probably want to track to a different Analytics ID for each type of mini app. So I think that using a single `instance` of `MiniAppAnalytics` won't work. We will need to refactor this so that a new instance of `MiniAppAnalytics` is created each time a new instance is created of `MiniApp`. 

So maybe we could do the following:
- Remove the methods for `addAnalyticsConfig` and `removeAnalyticsConfig`
- Add a new property to `MiniAppSdkConfig` for `List<MiniAppAnalyticsConfig>`
- Refactor the usage `MiniAppAnalytics` so that it doesn't have a single instance. It should instead accept `rasProjectId` and `List<MiniAppAnalyticsConfig>` in the constructor. Then a new instance of `MiniAppAnalytics` should be created and used whenever `RealMiniApp.updateConfiguration` is called.
- Then we need to somehow deal with Analytics event which needs to be sent at SDK initialization time (https://github.com/rakutentech/android-miniapp/blob/master/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniappSdkInitializer.kt#L85). The problem is that the `MiniAppAnalytics` instance won't be available in `MiniappSdkInitializer`. So maybe we can just a specific static method for tracking that event?

What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -135,6 +140,9 @@ internal class RealMiniApp(
             miniAppDownloader.updateApiClient(it)
             miniAppInfoFetcher.updateApiClient(it)
         }
+        if (newConfig.rasProjectId != null)
```
khairul-alam-licon, line 143: I think, this statement is always true, so we can write `newConfig.rasProjectId.isEmpty()` instead 🤔

testapp/src/main/java/com/rakuten/tech/mobile/testapp/SampleApplication.kt
```diff
@@ -4,6 +4,8 @@ import android.app.Application
 import com.google.android.gms.ads.MobileAds
 import com.microsoft.appcenter.AppCenter
 import com.microsoft.appcenter.crashes.Crashes
+import com.rakuten.tech.mobile.miniapp.MiniApp
+import com.rakuten.tech.mobile.miniapp.analytics.MiniAppAnalyticsConfig
```
khairul-alam-licon, line 8: We can remove those imports?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/AppSettings.kt
```diff
@@ -88,6 +90,11 @@ class AppSettings private constructor(context: Context) {
             cache.urlParameters = urlParameters
         }
 
+    var miniAppAnalyticsConfigs: List<MiniAppAnalyticsConfig>
+        get() = cache.miniAppAnalyticsConfigs ?: emptyList()
+        set(miniAppAnalyticsConfigs) {
+            cache.miniAppAnalyticsConfigs = miniAppAnalyticsConfigs
+        }
```
khairul-alam-licon, line 97: maybe we can add a new line after this line.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/AppSettings.kt
```diff
@@ -174,6 +188,13 @@ private class Settings(context: Context) {
         get() = prefs.getString(URL_PARAMETERS, null)
         set(urlParameters) = prefs.edit().putString(URL_PARAMETERS, urlParameters).apply()
 
+    var miniAppAnalyticsConfigs: List<MiniAppAnalyticsConfig>?
```
khairul-alam-licon, line 191: are we using this value somewhere? or, is there any plan in future e.g. using it in an UI.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalyticsConfig.kt
```diff
@@ -0,0 +1,15 @@
+package com.rakuten.tech.mobile.miniapp.analytics
+
+import androidx.annotation.Keep
+
+/**
+ *  Contains the components which need to add extra analytics credentials from host app.
+ *  @property acc The RAT account id.
+ *  @property aid The RAT app id.
+ *  @property endPoint is not in use.
```
khairul-alam-licon, line 9: I think, we can remove this line.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +17,86 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
-        // Send to this acc/aid
-        params["acc"] = BuildConfig.ANALYTICS_ACC
-        params["aid"] = BuildConfig.ANALYTICS_AID
+        /** common function to create params to send to tracker. */
+        @Suppress("LongMethod")
+        private fun createParams(
+            rasProjectId: String,
+            acc: Int,
+            aid: Int,
+            actype: Actype,
```
khairul-alam-licon, line 53: We can rename it to `acType` to preserve camelCase style.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +17,86 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
-    }
 
-    fun sendAnalytics(eType: Etype, actype: Actype, miniAppInfo: MiniAppInfo?) = try {
-        val params = mutableMapOf<String, Any>()
-        // Send to this acc/aid
-        params["acc"] = BuildConfig.ANALYTICS_ACC
-        params["aid"] = BuildConfig.ANALYTICS_AID
+        /** common function to create params to send to tracker. */
+        @Suppress("LongMethod")
+        private fun createParams(
```
khairul-alam-licon, line 49: Nice 🎉 Maybe we can add unit test if you want..

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +17,86 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+            acc = BuildConfig.ANALYTICS_ACC,
+            aid = BuildConfig.ANALYTICS_AID,
+            actype = actype,
+            miniAppInfo = miniAppInfo
+        )
+        whenHasAnalytics {
+            RatTracker.event(eType.value, params).track()
+        }
+        // Send to all the external acc/aid added by host app
+        for (config in configs) {
```
khairul-alam-licon, line 88: maybe we can use destructuring declaration e.g.

```
for ((acc, aid) in configs) { ... }
```

khairul-alam-licon (commented):
Looks good, besides, I have left few minor comments.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
```diff
@@ -15,7 +17,8 @@ data class MiniAppSdkConfig(
     val subscriptionKey: String,
     val hostAppVersionId: String = "",
     val hostAppUserAgentInfo: String,
-    val isPreviewMode: Boolean
+    val isPreviewMode: Boolean,
+    val miniAppAnalyticsConfigs: List<MiniAppAnalyticsConfig> = emptyList()
```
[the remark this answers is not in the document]
munir-rakuten: sure.. thanks for the insight.. @khairul-alam-licon

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
munir-rakuten: i think no need for any check here.. @khairul-alam-licon, we always have `rasProjectId` to create a miniApp so its ok not to check anything here

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/AppSettings.kt
munir-rakuten: @khairul-alam-licon yes we are not using it now, i think it will be useful for future to implement UI, now its just hard coded so i thought if we add ui later it will be useful to use this cache

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
munir-rakuten: i think its better to be like this because there is  a predefined parameter for `RatTracker` actype, so it can avoid confusion, and also this variable is not defined by me, it was already there @khairul-alam-licon

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt
corycaywood, line 22: `miniAppAnalyticsConfigs` looks like a typo. How about this instead?

```suggestion
    val miniAppAnalyticsConfigList: List<MiniAppAnalyticsConfig> = emptyList()
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,87 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+    val rasProjectId: String,
+    private val configs: List<MiniAppAnalyticsConfig>
+) {
 
     companion object {
-        var instance: MiniAppAnalytics? = null
-
-        fun init(rasProjectId: String) = whenHasAnalytics {
-            instance = MiniAppAnalytics(rasProjectId)
+        @Suppress("LongMethod")
```
corycaywood, line 27: This method doesn't look very long - does it really need this suppression?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,87 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
         }
-    }
 
-    fun sendAnalytics(eType: Etype, actype: Actype, miniAppInfo: MiniAppInfo?) = try {
-        val params = mutableMapOf<String, Any>()
-        // Send to this acc/aid
-        params["acc"] = BuildConfig.ANALYTICS_ACC
-        params["aid"] = BuildConfig.ANALYTICS_AID
+        /** common function to create params to send to tracker. */
+        @Suppress("LongMethod")
```
corycaywood, line 49: Same here - I don't think it needs the suppression?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,87 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
-        if (miniAppInfo != null) {
-            cp.put("mini_app_id", miniAppInfo.id)
-                .put("mini_app_version_id", miniAppInfo.version.versionId)
+            return params
         }
-        params["cp"] = cp
+    }
 
-        RatTracker.event(eType.value, params).track()
+    @Suppress("LongMethod")
```
corycaywood, line 76: Same here - I don't think it needs the suppression?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,87 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+        ) = try {
+            val params = createParams(
+                rasProjectId = rasProjectId,
+                acc = BuildConfig.ANALYTICS_ACC,
+                aid = BuildConfig.ANALYTICS_AID,
+                actype = actype,
+                miniAppInfo = miniAppInfo
+            )
+            whenHasAnalytics {
+                RatTracker.event(eType.value, params).track()
```
corycaywood, line 42: The same code is used below for tracking the event. Maybe you should extract it to a method to avoid mis-aligned behavior between the tracking methods? For example:

```
private fun trackEvent(eType: Etype, params: Map<String, Any>) = whenHasAnalytics {
    RatTracker.event(eType.value, params).track()
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,87 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+        /** common function to create params to send to tracker. */
+        @Suppress("LongMethod")
+        @VisibleForTesting
+        internal fun createParams(
+            rasProjectId: String,
+            acc: Int,
+            aid: Int,
+            actype: Actype,
+            miniAppInfo: MiniAppInfo?
+        ): MutableMap<String, Any> {
```
corycaywood, line 57: Does this need to be `Mutable`? Couldn't you just make it `Map<String, Any>`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,87 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+            val params = mutableMapOf<String, Any>()
```
corycaywood, line 58: Similar to my comment above - it's better to use Immutable objects when possible. So maybe you could do something like this:

```
val cp = JSONObject()
    .put("mini_app_project_id", rasProjectId)
    .put("mini_app_sdk_version", BuildConfig.VERSION_NAME)
if (miniAppInfo != null) {
    cp.put("mini_app_id", miniAppInfo.id)
        .put("mini_app_version_id", miniAppInfo.version.versionId)
}

return mapOf<String, Any>(
    "acc" to acc,
    "aid" to aid,
    "actype" to actype.value,
    "cp" to cp
)
```

corycaywood (approved):
Looks good 👍  I just left a few suggestions below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
munir-rakuten: oh! sorry it was long before, i refactor the code bt forget to remove them.. @corycaywood

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
khairul-alam-licon: Oh I see.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
munir-rakuten: @corycaywood  bt without it detekt didnt pass..what should i do?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/AppSettings.kt
khairul-alam-licon: I got it, we can add it in the same PR with UI implementation, so we can make things clean as possible.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,90 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+        // Send to this acc/aid
+        val params = createParams(
+            rasProjectId = rasProjectId,
+            acc = BuildConfig.ANALYTICS_ACC,
+            aid = BuildConfig.ANALYTICS_AID,
+            actype = actype,
+            miniAppInfo = miniAppInfo
+        )
+        whenHasAnalytics {
+            RatTracker.event(eType.value, params).track()
```
khairul-alam-licon, line 90: Can we use `trackEvent(eType, params)` instead?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
```diff
@@ -17,34 +18,90 @@ private inline fun <T> whenHasAnalytics(callback: () -> T) {
...
+            miniAppInfo: MiniAppInfo?
+        ) = try {
+            val params = createParams(
+                rasProjectId = rasProjectId,
+                acc = BuildConfig.ANALYTICS_ACC,
+                aid = BuildConfig.ANALYTICS_AID,
+                actype = actype,
+                miniAppInfo = miniAppInfo
+            )
+            whenHasAnalytics {
```
khairul-alam-licon, line 40: Do we need to additionally write this check when `trackEvent` method already has the same line?

khairul-alam-licon (approved):
Looks good, nice work 🎉

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/analytics/MiniAppAnalytics.kt
munir-rakuten: good catch!! ,thanks @khairul-alam-licon

PR #319, opened 2021-06-04
[MINI-2898] upload mappings to appcenter 

khairul-alam-licon:
# Description
This PR aims to upload `mapping.txt` to AppCenter's Diagnostics option using circleci. 

# Links
- MINI-2898 

# Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   android-sdk: rakutentech/android-sdk@0.2.1
-  app-center: rakutentech/app-center@0.1.3
+  app-center: rakutentech/app-center@dev:0.1.3
 
 commands:
   update-submodules:
@@ -115,31 +115,7 @@ workflows:
           file: mapping/staging/mapping.txt
           app: $APP_CENTER_APP_NAME
           token: $APP_CENTER_TOKEN
-          version-name: $APP_VERSION_NAME
-          version-code: 1
-          requires:
-            - android-sdk/build
-          filters:
-            branches:
-              only: master
-      - app-center/mapping:
-          name: deploy-test-app-release-mapping
-          file: mapping/release/mapping.txt
-          app: $APP_CENTER_APP_NAME
-          token: $APP_CENTER_TOKEN
-          version-name: $APP_VERSION_NAME
-          version-code: 1
-          requires:
-            - android-sdk/build
-          filters:
-            branches:
-              only: master
-      - app-center/mapping:
-          name: deploy-test-app-rc-mapping
-          file: mapping/rc/mapping.txt
-          app: $APP_CENTER_APP_NAME
-          token: $APP_CENTER_TOKEN
-          version-name: $APP_VERSION_NAME
+          version-name: $APP_CENTER_APP_NAME
           version-code: 1
           requires:
             - android-sdk/build
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   android-sdk: rakutentech/android-sdk@0.2.1
-  app-center: rakutentech/app-center@0.1.2
+  app-center: rakutentech/app-center@0.1.3
 
 commands:
   update-submodules:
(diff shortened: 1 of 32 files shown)


.circleci/config.yml
```diff
@@ -110,3 +110,41 @@ workflows:
...
+            tags:
+              only: /^v.*/
+            branches:
+              ignore: /.*/
+      - app-center/mapping:
+          name: deploy-test-app-rc-mapping
+          file: mapping/rc/mapping.txt
+          app: $APP_CENTER_APP_NAME
+          token: $APP_CENTER_TOKEN
+          version-name: $APP_CENTER_APP_NAME
```
khairul-alam-licon, line 144: I'll have a separate PR to fix it after ensuring the uploading will work.

Climbatize (approved):

.circleci/config.yml
```diff
@@ -110,3 +110,41 @@ workflows:
           filters:
             branches:
               only: /^feature\/.*/
+      - app-center/mapping:
+          name: deploy-test-app-stg-mapping
+          file: mapping/staging/mapping.txt
+          app: $APP_CENTER_APP_NAME
+          token: $APP_CENTER_TOKEN
+          version-name: $APP_CENTER_APP_NAME
+          version-code: 1
```
corycaywood, line 119: Is this correct? Doesn't this mean that the mappings will all be uploaded to the same version?

khairul-alam-licon: Yeah, currently we are using `1` as default value for all `version-code` in appcenter.

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 
 orbs:
   android-sdk: rakutentech/android-sdk@0.2.1
```
munir-rakuten, line 4: @khairul-alam-licon @Climbatize  does this orb needs to be updated to cache the mapping file in circle ci? or its already added?

.circleci/config.yml
corycaywood: But is that going to cause problems between versions? Can App Center distinguish the mappings based on just the version name?

khairul-alam-licon: Yeah as the [official document](https://docs.microsoft.com/en-us/appcenter/diagnostics/android-deobfuscation) says to match it with build's Gradle configuration, but we are not changing `versionCode` but the `versionName` https://github.com/rakutentech/android-miniapp/blob/master/testapp/build.gradle#L13 🤔

.circleci/config.yml
corycaywood: That's a good point. I think it should work because it's already persisting the whole `output` folder to the workspace: https://github.com/rakutentech/android-buildconfig/blob/master/circleci/android-sdk/config.yml#L79

But I just noticed it looks like the path being passed to the mapping job is incorrect.

.circleci/config.yml
```diff
@@ -110,3 +110,41 @@ workflows:
           filters:
             branches:
               only: /^feature\/.*/
+      - app-center/mapping:
+          name: deploy-test-app-stg-mapping
+          file: mapping/staging/mapping.txt
```
corycaywood, line 115: Looks like these paths are incorrect. I think it should be `apk/staging/mapping.txt`?

The build job is persisting the artifacts to a folder named `apk`:
https://github.com/rakutentech/android-buildconfig/blob/master/circleci/android-sdk/config.yml#L81

khairul-alam-licon: Oh really? But, when I am building it locally, I can see the following path:

Screen Shot 2021-06-08 at 12 54 13

corycaywood: @khairul-alam-licon See the above link. When you persist to the workspace on CircleCI, you provide a path which it is persisted to. In this case, it's persisted to a folder `apk`. This is why the `appcenter/publish` jobs in this file are pointing to `apk/staging/testapp-staging.apk`.

munir-rakuten: @corycaywood if the persist the outputs in apk folder, then why there is no middle directories like `apk/apk/staging/testapp-staging.apk` or `apk/mapping/staging/mapping.txt`? just curious..

corycaywood: @munir-rakuten See the documentation for persist to workspace:
https://circleci.com/docs/2.0/configuration-reference/#persisttoworkspace

It's not very clear on the explanation, but basically it takes the `root/path` that you pass and persists it to `working_directory/path`

corycaywood (approved):
👍

PR #322, opened 2021-06-09
[MINI-2898] upgrade appcenter orb version

khairul-alam-licon:
# Description
- app-center orb needs to be upgraded to latest version e.g. `rakutentech/app-center@0.1.3`
- update `version-name` with right value

# Checklist


Climbatize (approved):
👌

build.gradle
```diff
@@ -110,4 +110,5 @@ tasks.withType(Test) {
 tasks.withType(JavaCompile) {
     options.fork = true
     options.incremental = true
+    systemProperty 'APP_VERSION_NAME', project.version
```
corycaywood, line 113: I don't think this will work - this property will only be available in the JVM and not to the shell environment. Also, the app center task doesn't run any gradle commands anyways, it just uses NPM.

@Climbatize  @khairul-alam-licon If you need the version name, then you might need to get it from the App Center upload job. I think that the App Center CLI returns the version name after you've uploaded the APK. In my opinion, it might make more sense if the `app-center/publish` job were able to accept an optional parameter for the mapping file - then that single job would upload the APK, save the returned version name to an environment variable, and then upload the mapping for the version. What do you think?

Climbatize: To me it's clear that you are right @corycaywood, because we need a name such as `3.3.0-SNAPSHOT-STG-build-2106091845` or `2.9.0-SNAPSHOT-STG-build-1946` and ideally the version number should also be dynamic instead of  `1`

khairul-alam-licon: I think, fixing only the version name will be enough. We don't change version code in gradle configuration.

corycaywood (approved):
Looks good 👍

PR #324, opened 2021-06-10
fix: return null when file is not found from downloaded manifest cache

khairul-alam-licon:
# Description
`DownloadedManifestCache.readFromCachedFile` should return `null` while manifest cached file is not found, this PR aims to fix this issue to prevent a crash.

## Links
- MINI-3371

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/DownloadedManifestCache.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/DownloadedManifestCache.kt
@@ -127,13 +127,15 @@ internal class DownloadedManifestCache(context: Context) {
     }
 
     @VisibleForTesting
-    fun readFromCachedFile(miniAppId: String): CachedManifest? {
+    fun readFromCachedFile(miniAppId: String): CachedManifest? = try {
         val jsonToRead = File(getManifestPath(miniAppId), DEFAULT_FILE_NAME).bufferedReader()
             .use {
                 it.readText()
                     .dropLast(2) // for fixing the json format
             }
-        return toCachedManifest(jsonToRead)
+        toCachedManifest(jsonToRead)
+    } catch (e: Exception) {
+        null
     }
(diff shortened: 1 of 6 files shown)


munir-rakuten (approved):
Looks good to me 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/DownloadedManifestCache.kt
```diff
@@ -127,13 +127,15 @@ internal class DownloadedManifestCache(context: Context) {
     }
 
     @VisibleForTesting
-    fun readFromCachedFile(miniAppId: String): CachedManifest? {
+    fun readFromCachedFile(miniAppId: String): CachedManifest? = try {
         val jsonToRead = File(getManifestPath(miniAppId), DEFAULT_FILE_NAME).bufferedReader()
             .use {
                 it.readText()
                     .dropLast(2) // for fixing the json format
             }
```
Climbatize, line 135: I feel like the try catch to fix the file existence issue is kind of dirty. Should we create a `File` variable and test its existence, like

```suggestion
    fun readFromCachedFile(miniAppId: String): CachedManifest? {
        var file = File(getManifestPath(miniAppId), DEFAULT_FILE_NAME)
        if(!file.exists()) return null
        val jsonToRead = file.bufferedReader()
            .use {
                it.readText()
                    .dropLast(2) // for fixing the json format
            }
```
Maybe you can still keep the `try` for the json format errors

Climbatize (approved):
LGTM but I advise to consider my suggestion

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/DownloadedManifestCache.kt
khairul-alam-licon: Thanks for the suggestion, I have updated this part.

PR #332, opened 2021-06-29
Webview can accept specific file type

munir-rakuten:
# Description
Webview can accept specific image file type.

## Links
MINI-3346

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
@@ -80,7 +80,7 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
 
         val mtm = MimeTypeMap.getSingleton()
         for (mime in mimeTypes) {
-            if (mime != null && mime.trim { it <= ' ' }.startsWith(".")) {
+            if (mime.trim { it <= ' ' }.startsWith(".")) {
                 val extensionWithoutDot =
                     mime.trim { it <= ' ' }.substring(1, mime.trim { it <= ' ' }.length)
                 val derivedMime = mtm.getMimeTypeFromExtension(extensionWithoutDot)
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
@@ -52,7 +52,9 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
             }
             // Uses Intent.EXTRA_MIME_TYPES to pass multiple mime types.
             fileChooserParams?.acceptTypes?.let { acceptTypes ->
-                if (!acceptTypes.equals("") && acceptTypes.isNotEmpty()) {
+                if (acceptTypes.isNotEmpty()
+                    && !(acceptTypes.size == 1 && acceptTypes[0].equals(""))
+                ) {
                     // Accept all first.
                     intent?.type = "*/*"
                     // Convert to valid MimeType if with dot.
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
@@ -52,9 +52,7 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
             }
             // Uses Intent.EXTRA_MIME_TYPES to pass multiple mime types.
             fileChooserParams?.acceptTypes?.let { acceptTypes ->
-                if (acceptTypes.isNotEmpty()
-                    && !(acceptTypes.size == 1 && acceptTypes[0].equals(""))
-                ) {
+                if (acceptTypes.isNotEmpty() && !(acceptTypes.size == 1 && acceptTypes[0].equals(""))) {
                     // Accept all first.
                     intent?.type = "*/*"
                     // Convert to valid MimeType if with dot.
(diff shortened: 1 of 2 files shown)


khairul-alam-licon (approved):
Looks good 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -49,6 +49,26 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
             if (fileChooserParams?.mode == WebChromeClient.FileChooserParams.MODE_OPEN_MULTIPLE) {
                 intent?.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true)
             }
+            // Uses Intent.EXTRA_MIME_TYPES to pass multiple mime types.
+            fileChooserParams?.acceptTypes?.let { acceptTypes ->
+                if (!acceptTypes.equals("") && acceptTypes.size > 1) {
+                    // Accept all first.
+                    intent?.type = "*/*"
+                    val imageExtensionList = listOf(".jpg", ".jpeg", ".png")
```
corycaywood, line 57: This is only going to work for `image/` mime types - also this still doesn't cover all image types such as svg, gif, etc. It's possible someone will want to use this for audio, video, or any other type of files as well. So we would need to update this function every time that we need to support another mime-type.

Could you instead use [MimeTypeMap.getMimeTypeFromExtension](https://developer.android.com/reference/android/webkit/MimeTypeMap#getMimeTypeFromExtension(java.lang.String)) to convert the extension to a mime type?

munir-rakuten: @corycaywood updated the code for any type of files now..

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -49,6 +50,17 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
             if (fileChooserParams?.mode == WebChromeClient.FileChooserParams.MODE_OPEN_MULTIPLE) {
                 intent?.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true)
             }
+            // Uses Intent.EXTRA_MIME_TYPES to pass multiple mime types.
+            fileChooserParams?.acceptTypes?.let { acceptTypes ->
+                if (!acceptTypes.equals("") && acceptTypes.size > 1) {
```
corycaywood, line 55: `acceptTypes.size > 1` means that it won't work if only one type is passed, i.e. `accept=".jpg"`. Should it be like this instead?
```suggestion
                if (!acceptTypes.equals("") && acceptTypes.isNotEmpty()) {
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -58,6 +70,32 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
...
+     * Validation Utility for MimeTypes.
+     * @param mimeTypes The Mimetypes needs to convert to valid types.
+     */
+    @VisibleForTesting
+    internal fun extractValidMimeTypes(mimeTypes: Array<String>): List<String> {
+        val results: MutableList<String> = ArrayList()
+
+        val mtm = MimeTypeMap.getSingleton()
+        for (mime in mimeTypes) {
+            if (mime.trim { it <= ' ' }.startsWith(".")) {
```
corycaywood, line 83: Isn't a single space the default character for `trim`? So couldn't this be just `mime.trim()`?

Also, you're calling `mime.trim` several times so I think you should either set this to a variable or use something like `mime.trim().let { ... }`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -58,6 +70,32 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
...
+                val extensionWithoutDot =
+                    mime.trim { it <= ' ' }.substring(1, mime.trim { it <= ' ' }.length)
```
corycaywood, line 85: You could simplify this by using `removePrefix(".")`?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -58,6 +70,32 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
...
+                val derivedMime = mtm.getMimeTypeFromExtension(extensionWithoutDot)
+                if (derivedMime != null && !results.contains(derivedMime)) {
+                    // adds valid mime type derived from the file extension
+                    results.add(derivedMime)
+                }
+            } else if (mtm.getExtensionFromMimeType(mime) != null && !results.contains(mime)) {
+                // adds valid mime type checked again file extensions mappings
+                results.add(mime)
```
corycaywood, line 93: Shouldn't `mime` here be the trimmed one?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -58,6 +70,32 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
         return true
     }
 
+    /**
+     * Validation Utility for MimeTypes.
+     * @param mimeTypes The Mimetypes needs to convert to valid types.
+     */
+    @VisibleForTesting
+    internal fun extractValidMimeTypes(mimeTypes: Array<String>): List<String> {
+        val results: MutableList<String> = ArrayList()
```
corycaywood, line 79: I think you don't need to create a temporary list here. You could instead use `mapNotNull`, for example:

```
mimeTypes.mapNotNull { mime -> 
  if ( // something //) {
    mime
  } else {
    null
  }
}
```

corycaywood (commented):
It's looking good, but I left a few comments below.

Also, I think you're missing some unit test cases. For example the cases where a bad mime type is provided or duplicate types are provided.

corycaywood (approved):
Looks good! 👍

PR #331, opened 2021-06-29
[MINI-3240] feat: user points implementation 

khairul-alam-licon:
# Description
- Add an interface to retrieve the user's points information with custom permission e.g. `rakuten.miniapp.user.POINTS`
- In the Demo App settings screen, add an option where the points data can be set.
- Update with JS bridge. 

## Links
- MINI-3240

## Screenshot
Screen Shot 2021-06-29 at 14 22 07

# Checklist
- [x] I have read the [contributing guidelines](../.github/CONTRIBUTING.md).
- [x] I have completed the [SDK Development Learning Path](../.github/CONTRIBUTING.md#SDK-Development-Learning-Path) (if submitting a feature PR)
- [x] I wrote/updated tests for new/changed code
- [x] I removed all sensitive data before every commit, including API endpoints and keys
- [x] I ran `./gradlew check` without errors

PR #338, opened 2021-07-09
[MINI-3240] add user point error 

khairul-alam-licon:
# Description
Primary work of user's point interface has been added by https://github.com/rakutentech/android-miniapp/pull/331.
This PR aims to add `MiniAppPointsError` for sending it to the mini-apps.

## Links
- MINI-3240

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppPointsError.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppPointsError.kt
@@ -1,0 +1,16 @@
+package com.rakuten.tech.mobile.miniapp.errors
+
+/**
+ * A class to provide the custom errors specific for user's points.
+ */
+class MiniAppPointsError(val type: String? = null, val message: String? = null) :
+    MiniAppBridgeError(type, message) {
+
+    companion object {
+        /**
+         *  send custom error message from host app.
+         *  @property message error message send to mini app.
+         */
+        fun custom(message: String) = MiniAppPointsError(message = message)
+    }
+}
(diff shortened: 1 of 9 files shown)


corycaywood (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppPointsError.kt
```diff
@@ -0,0 +1,16 @@
...
+    MiniAppBridgeError(type, message) {
+
+    companion object {
+        /**
+         *  send custom error message from host app.
+         *  @property message error message send to mini app.
+         */
+        fun custom(message: String) = MiniAppPointsError(message = message)
+    }
+}
```
munir-rakuten, line 16: Do we need this class? As we don't have any specific type for this error and only message, so may be we can use `MiniAppBridgeError` for this as both looks identical..but if we include specific types in future we can implement like this, what do you think?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -222,11 +223,11 @@ class MiniAppDisplayActivity : BaseActivity() {
 
             override fun getPoints(
                 onSuccess: (points: Points) -> Unit,
-                onError: (message: String) -> Unit
+                onError: (pointsError: MiniAppPointsError) -> Unit
             ) {
                 val points = AppSettings.instance.points
                 if (points != null) onSuccess(points)
-                else onError("There is no points found in HostApp.")
+                else onError(MiniAppPointsError(null, "There is no points found in HostApp."))
```
munir-rakuten, line 230: We can use `MiniAppPointsError.custom(""There is no points found in HostApp.")` like this? or make the type nullable?

munir-rakuten (commented):
Looks good, I placed some suggestions though..

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppPointsError.kt
khairul-alam-licon: Yeah, we can use `MiniAppPointsError` to denote the feature specific error class, also we can use the `custom` function to pass the message only.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
khairul-alam-licon: Updated.

munir-rakuten (approved):

PR #340, opened 2021-07-12
fix: grep version name 

khairul-alam-licon:
# Description

The following line is not working. It's producing `No such file or directory` issue which results an empty `version-name` in the app-center.

```
version-name: $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/rc/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d \''";\')
```

This PR is fixing this.


.circleci/config.yml
```diff
@@ -41,6 +41,8 @@ workflows:
                 root: testapp/build/outputs
                 paths:
                   - mapping/
+            - run: |
+                echo $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d \''";\') > ./testapp/version-name.txt
```
Climbatize, line 45: I think your `tr` command is broken, on the end, here is one I tested
```suggestion
                echo $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d '";') > ./testapp/version-name.txt
```

.circleci/config.yml
```diff
@@ -122,7 +122,7 @@ workflows:
           file: mapping/staging/mapping.txt
           app: $APP_CENTER_APP_NAME
           token: $APP_CENTER_TOKEN
-          version-name: echo $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d \''";\')
+          version-name: echo $(testapp/version-name-staging.txt)
```
Climbatize, line 125: do you really need to create a file? If you included what you do line 45 into script, do you think it's possible?

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -179,14 +179,7 @@ class MiniAppDisplayActivity : BaseActivity() {
...
 
-            override fun getUserName(
-                onSuccess: (userName: String) -> Unit,
-                onError: (message: String) -> Unit
-            ) {
-                val name = AppSettings.instance.profileName
-                if (name.isNotEmpty()) onSuccess(name)
-                else onError("User name is not found.")
-            }
+
```
Climbatize, line 182: merge issue?

.circleci/config.yml
```diff
@@ -120,13 +120,13 @@ workflows:
...
           token: $APP_CENTER_TOKEN
-          version-name: $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d \''";\')
+          version-name: echo $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d '";')
           version-code: 1
           requires:
             - android-sdk/build
           filters:
             branches:
-              only: master
+              only: /.*/
```
Climbatize, line 129: do you nee a filter if you want all branches? :)

khairul-alam-licon: sorry, I temporary enabled it because of checking something in circleci in this branch. this is a redundant change.

.circleci/config.yml
khairul-alam-licon: @Climbatize Yeah, the original purpose to open this PR to investigate an issue e.g. `No such file or directory` while trying to grep the value by traversing in the given file path. Because the existing script in `master` branch we have still producing empty `version-name` value in app-center.

So, this was my initial approach by storing the value in a temporary file e.g. `version-name-staging.txt` or in a temporary variable then use it directly here (L125). 

I think, `No such file or directory` is kinda misleading, means its finding the file but there is some other kinda issue.

.circleci/config.yml
khairul-alam-licon: but, I think the previous one was printing the expected value correctly.

.circleci/config.yml
```diff
@@ -120,13 +120,13 @@ workflows:
           file: mapping/staging/mapping.txt
           app: $APP_CENTER_APP_NAME
           token: $APP_CENTER_TOKEN
-          version-name: $(grep "VERSION_NAME" testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d \''";\')
+          version-name: $(grep "VERSION_NAME" ../testapp/build/generated/source/buildConfig/staging/com/rakuten/tech/mobile/miniapp/testapp/BuildConfig.java | awk '{print $7}' | tr -d '";')
```
corycaywood, line 123: @khairul-alam-licon  Is this working? I think it shouldn't be using `../` because circleci runs the scripts from the root of the code directory. 

Also, I don't think that all of the build artifacts are being persisted to the workspace: only the APKs and the mappings currently are persisted (see https://github.com/rakutentech/android-buildconfig/blob/master/circleci/android-sdk/config.yml#L78).

So I think you need to add an additional step to `post-steps:` in the build job above so that `BuildConfig.java` is persisted to the workspace. Alternatively, it looks like the whole `build` directory for `miniapp` is being persisted to the workspace, so maybe you could just use the BuildConfig from that instead?

khairul-alam-licon: Yeah, let me to check. Sorry, I haven't returned to this PR for long time 😅

khairul-alam-licon: @corycaywood Yeah, I have updated the PR with persistence to the workplace. It's not showing the error anymore but looks like we need to merge this PR to see the effect.

corycaywood (approved):
Looks good! 👍

.circleci/config.yml
```diff
@@ -41,6 +41,10 @@ workflows:
                 root: testapp/build/outputs
                 paths:
                   - mapping/
+            - persist_to_workspace:
+                root: testapp/build/generated/source
+                paths:
+                  - buildConfig/
```
munir-rakuten, line 47: Just a suggestion, we can add this persistence in buildConfig/config.yml file, same as apk persist..

munir-rakuten (approved):
LGTM!!👍 I have a suggestion though if it is working...

.circleci/config.yml
khairul-alam-licon: Yeah, that's true.. I have the same intention but I want to confirm the effect before making change in `buildConfig/config.yml`

PR #343, opened 2021-07-13
fix NumberFormatException with Int values in Points

khairul-alam-licon:
# Description
This PR aims to prevent the `NumberFormatException` in demo app's `Points` settings for the values which is out of range.

## Links
- MINI-3470

# Checklist

### DIFF
--- a/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/PointsActivity.kt
+++ b/testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/PointsActivity.kt
@@ -69,9 +69,9 @@ class PointsActivity : BaseActivity() {
         if (pointRakutenCash == "") pointRakutenCash = "0"
         try {
             settings.points = Points(
-                    pointStandard.toLong(),
-                    pointTimeLimited.toLong(),
-                    pointRakutenCash.toLong()
+                    pointStandard.toInt(),
+                    pointTimeLimited.toInt(),
+                    pointRakutenCash.toInt()
             )
         } catch (e: NumberFormatException) {
             e.printStackTrace()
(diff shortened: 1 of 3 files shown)


Climbatize (commented):
Android `Int` is coded on 32 bits, so max value is 2,147,483,647
To avoid using `Long` values you can also limit user's possibilities using a NumberPicker instead of a textfield, or set your max lengths to 9

munir-rakuten (approved):
Looks good, thanks..

corycaywood (approved):
👍

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/PointsActivity.kt
```diff
@@ -67,10 +67,19 @@ class PointsActivity : BaseActivity() {
...
-        )
+        try {
+            settings.points = Points(
+                    pointStandard.toInt(),
+                    pointTimeLimited.toInt(),
+                    pointRakutenCash.toInt()
+            )
+        } catch (e: NumberFormatException) {
+            e.printStackTrace()
+        }
```
Climbatize, line 78: Instead of having one try/catch, you should have method which takes a String as parameter and returns an Int, it could return the int that it passed or `-1` if it caught an exception.

```suggestion
       settings.points = Points(
                extractPointsFrom(pointStandard),
                extractPointsFrom(pointTimeLimited),
                extractPointsFrom(pointRakutenCash)
        )
```
```kotlin
fun extractPointsFrom(pointString: String): Integer {
    try {
          return pointString.toInt()
      } catch (e: NumberFormatException) {
          return -1
      }
}
```

Climbatize (approved):
LGTM

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/userdata/PointsActivity.kt
khairul-alam-licon: @Climbatize Oh, sorry I missed your suggestion before merging 🙇 
But, I think, we can return `0` when there is exception. 
Although, because of `maxLength="9"` in the input UI, we will not have any exception now. So, I'll think about this suggestion in a separate PR. Thank you!

PR #361, opened 2021-09-08
[MINI-3292 + MINI-3320] Show miniapp through qr code

munir-rakuten:
# Description
Added API interface to get miniapp through qr code, Also added optional ssl certificate pinning for api calls from miniapp sdk.

## Links
MINI- 3292, MINI- 3320

# Checklist


Climbatize (approved):
Looks good to me!

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/error/QRErrorWindow.kt
```diff
@@ -44,32 +54,48 @@ class QRErrorWindow {
     }
 
     /** show error screen for miniApp no longer exist. */
-    fun showMiniAppNoLongerExistError() {
-        if (instance != null) renderScreen(QRCodeErrorType.MiniAppNoLongerExist)
+    fun showMiniAppNoLongerExistError(onClosed: (()->Unit)? = null) {
+        if (instance != null) {
+            renderScreen(QRCodeErrorType.MiniAppNoLongerExist)
+            this.onClosed = onClosed
+        }
```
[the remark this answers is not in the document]
munir-rakuten: @Climbatize oh!! my bad, yeah i forgot about that, i wasn't focusing on that class i created may be 2 months ago and there is already a errorType, nice catch man, thanks, i updated the code..

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
```diff
@@ -189,16 +198,17 @@ abstract class MiniApp internal constructor() {
          * @return [MiniApp] instance
          */
         @JvmStatic
-        fun instance(settings: MiniAppSdkConfig = defaultConfig): MiniApp =
-            instance.apply { updateConfiguration(settings) }
+        fun instance(settings: MiniAppSdkConfig = defaultConfig, isTemporaryUpdate: Boolean = false): MiniApp =
```
corycaywood, line 201: What is the `isTemporaryUpdate` property for? Is this because the Host App might want to temporarily set the SSL pin?

Couldn't you instead add the ssl pin to the `key` that is set for each API Client? It uses the [MiniApSdkConfig.key](https://github.com/munir-rakuten/android-miniapp/blob/feature/add_preview_code_api/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniAppSdkConfig.kt#L28) for storing the ApiClient, so you could just append the SSL pin there. Then you still be able to distinguish between ApiClients using the same Host but one has an SSL pin and the other doesn't.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/api/RetrofitCreatorUtils.kt
```diff
@@ -27,13 +33,24 @@ internal fun createRetrofitClient(
...
     headers: RasSdkHeaders
 ): Retrofit {
     @Suppress("SpreadOperator")
-    val httpClient = OkHttpClient.Builder()
+    var httpClientBuilder = OkHttpClient.Builder()
         .addHeaderInterceptor(*headers.asArray())
         .addInterceptor(provideHeaderInterceptor())
-        .build()
+    if (pubKey != "") {
+        Log.e("Pinning key", pubKey)
```
corycaywood, line 44: Probably you can remove this log line.

testapp/build.gradle
```diff
@@ -53,7 +53,8 @@ android {
             subscriptionKey         : property("HOST_APP_SUBSCRIPTION_KEY") ?: "test-subs-key",
             adMobAppId              : property("ADMOB_APP_ID") ?: "",
             appName                 : defaultAppName + " DEBUG",
-            ratEndpoint             : property("ANALYTICS_ENDPOINT") ?: "https://www.example.com/"
+            ratEndpoint             : property("ANALYTICS_ENDPOINT") ?: "https://www.example.com/",
+            sslPinningPublicKey     : property("CERTIFICATE_PUBLIC_KEY") ?: ""
```
corycaywood, line 57: The public key should be different for Staging and Prod. So you need to setup two variables for this: 
`STG_CERTIFICATE_PUBLIC_KEY` and `PROD_CERTIFICATE_PUBLIC_KEY`

corycaywood (commented):
Looking good! I left some comments below.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/MiniApp.kt
munir-rakuten: @corycaywood actually, in demo app now the architecture is like, when we call `MiniApp.instance` it updates the all the api clients with the `SdkConfig` send as a parameter, previously the sdkConfig is always the the what is in `AppSettings`,
But now when we preview a `miniapp` through qr code from default host, we have to create the a `sdkConfig` with the `host id` and `subscriptionKey` but when we open the demo app again we can not use that `sdkConfig`, Because `appSettings` config and the miniapp we are loading from qrcode is different, that's when to make the qrcode preview sdkconfig temporary load the miniapp and the no store that clent so that when we open the demo app again it should use the config in the app settings.. does that make sense?

khairul-alam-licon: If we require this property, is it possible to rename it to understand the usage & also, we need to update the change-log.

munir-rakuten: @khairul-alam-licon @corycaywood or we can create another `instance` method like `instanceForPreviewCode` or `getInstanceForPreviewCode` and use that method without changing anything in the current instance method for that, i am confused which one will be better for this purpose..

corycaywood: @munir-rakuten So basically when you pass a new Config to `instance` it is setting the new Config as the default?

I think this doesn't sound like the right behavior for the SDK, but it's too late to change it since it was already like this. Probably it should instead always use the values from `AndroidManifest` as the default and then calling `instance` with new values just changes the Config for that specific instance. But I think we can't change this now.

Instead, I think maybe just the naming for the property needs to be something that makes it more clear how it works. How about using `setConfigAsDefault: Boolean = true`?

munir-rakuten: @corycaywood `So basically when you pass a new Config to instance it is setting the new Config as the default?` , answer is yes,...

`Instead, I think maybe just the naming for the property needs to be something that makes it more clear how it works. How about using setConfigAsDefault: Boolean = true?`
ok so true will be default.. if it's false then it will not update the default config..thanks

corycaywood (approved):
Looks good! 👍

khairul-alam-licon (approved):
Looks good, nice job 👍 
I left an optional feedback about function renaming in demo app side.

PR #395, opened 2021-11-29
[MINI-4421] Add support for camera capture image

munir-rakuten:
# Description
Capture Image through accessing camera from mini app.

https://user-images.githubusercontent.com/84305007/144017189-ad1a2b67-5bb9-4a1e-913d-1ea5f319de09.mov

## Links
MINI-4421

# Checklist

### DIFF
--- a/miniapp/src/main/AndroidManifest.xml
+++ b/miniapp/src/main/AndroidManifest.xml
@@ -30,6 +30,15 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/miniapp_downloaded_file_provider_paths" />
         </provider>
+        <provider
+            android:name="androidx.core.content.FileProvider"
+            android:authorities="${applicationId}.miniapp.imagefileprovider"
+            android:exported="false"
+            android:grantUriPermissions="true">
+            <meta-data
+                android:name="android.support.FILE_PROVIDER_PATHS"
+                android:resource="@xml/miniapp_image_file_provider_path"></meta-data>
+        </provider>
     </application>
 
     <queries>
--- a/miniapp/src/main/AndroidManifest.xml
+++ b/miniapp/src/main/AndroidManifest.xml
@@ -22,22 +22,13 @@
             tools:replace="android:initOrder" />
 
         <provider
-            android:name="com.rakuten.tech.mobile.miniapp.display.MiniAppDownloadedFileProvider"
-            android:authorities="${applicationId}.miniapp.downloadedfileprovider"
+            android:name="com.rakuten.tech.mobile.miniapp.display.MiniAppDefaultFileProvider"
+            android:authorities="${applicationId}.miniapp.defaultfileprovider"
             android:exported="false"
             android:grantUriPermissions="true">
             <meta-data
                 android:name="android.support.FILE_PROVIDER_PATHS"
-                android:resource="@xml/miniapp_downloaded_file_provider_paths" />
-        </provider>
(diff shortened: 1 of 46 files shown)


miniapp/src/main/AndroidManifest.xml
```diff
@@ -30,6 +30,15 @@
                 android:name="android.support.FILE_PROVIDER_PATHS"
                 android:resource="@xml/miniapp_downloaded_file_provider_paths" />
         </provider>
+        <provider
+            android:name="androidx.core.content.FileProvider"
```
corycaywood, line 34: The default FileProvider shouldn't be used because it could conflict with the Host App if they are also using the default `androidx.core.content.FileProvider`.

I think you could re-use the `com.rakuten.tech.mobile.miniapp.display.MiniAppDownloadedFileProvider` from above instead of creating a new FileProvider. It should already be the same as the default. Maybe it just needs to be renamed to something like `MiniAppDefaultFileProvider`?

corycaywood (commented):
Looking good - just see my comment below.

khairul-alam-licon (commented):
Looks great 👍 
I have left few minor issues.
Also, can we include this feature in the changelog?

miniapp/src/main/AndroidManifest.xml
khairul-alam-licon: I was trying to re-use `com.rakuten.tech.mobile.miniapp.display.MiniAppDownloadedFileProvider`, unfortunately it was not opening the camera :( 
Let's see if @munir-rakuten found something.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileChooser.kt
```diff
@@ -70,6 +82,45 @@ class MiniAppFileChooserDefault(var requestCode: Int) : MiniAppFileChooser {
...
+                takePictureIntent.putExtra(MediaStore.EXTRA_OUTPUT, photoURI)
+                (context as Activity).startActivityForResult(takePictureIntent, requestCode)
+            }
+        }
+    }
+
+    @Throws(IOException::class)
+    private fun createImageFile(context: Context): File {
+        // Create an image file name
+        val timeStamp: String = SimpleDateFormat("yyyyMMdd_HHmmss").format(Date())
```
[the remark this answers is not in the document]
munir-rakuten: no need to catch the exception, as the the whole code block is already in a try block @khairul-alam-licon

khairul-alam-licon (approved):
Looks good 👍

miniapp/src/main/AndroidManifest.xml
```diff
@@ -22,13 +22,13 @@
             tools:replace="android:initOrder" />
 
         <provider
-            android:name="com.rakuten.tech.mobile.miniapp.display.MiniAppDownloadedFileProvider"
-            android:authorities="${applicationId}.miniapp.downloadedfileprovider"
+            android:name="com.rakuten.tech.mobile.miniapp.display.MiniAppDefaultFileProvider"
```
corycaywood, line 25: Could you also add a rule to the `proguard-rules.pro` for this class? I should have done that earlier when I added this 😅 

I think something like this?
```
-keep class com.rakuten.tech.mobile.miniapp.display.MiniAppDefaultFileProvider { *; }
```

corycaywood (approved):
Looks good! Just see my comment below about proguard.

PR #397, opened 2021-12-06
[MINI-4458] bugfix: empty host locale value

khairul-alam-licon:
# Description
Host locale is coming empty when the device language is Chinese. This PR aims to fix it with matching with spec. By default, if HostApp doesn't support any language, then it should return `en-US` to MiniApp.

## Links
- MINI-4458

# Checklist

### DIFF
--- a/miniapp/src/main/res/values-ja/strings.xml
+++ b/miniapp/src/main/res/values-ja/strings.xml
@@ -1,0 +1,3 @@
+<resources>
+    <string name="miniapp.sdk.android.locale">ja-JP</string>
+</resources>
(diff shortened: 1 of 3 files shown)


Climbatize (approved):
LGTM

miniapp/src/main/res/values-ja/strings.xml
```diff
@@ -0,0 +1,3 @@
+<resources>
+    <string name="miniapp.sdk.android.locale">ja-JP</string>
```
corycaywood, line 2: I think this should use `_` to match the other strings.

khairul-alam-licon: Thanks, I have updated it 🙇

munir-rakuten (approved):
LGTM 👍

PR #409, opened 2022-01-18
fix: Unable to launch mini app when device offline

corycaywood:
# Description
This fixes a case where the the mini app is unable to be launched when the device is offline. The check being done on the manifest was failing when offline. We missed this case earlier because it only occurs when testing with multiple mini apps - we cache the manifest for the last launched mini app using ManifestApiCache

## Links
MINI-4635

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## CHANGELOG
 
+### 3.9.1 (2021-12-20)
+
+**SDK**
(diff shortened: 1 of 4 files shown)


Climbatize (approved):
Nice catch

khairul-alam-licon (approved):
Looks good 👍

munir-rakuten (approved):
Looks good to me 👍

PR #426, opened 2022-03-17
[MINI-4671] Download Files to device 

munir-rakuten:
New implementation for downloading files to device passing a url and filename instead of xhr (and base64)

- Download the file in a cache directory
- Share the file through intent
- Clear the cache when miniapp closed

## Links
[MINI-4671]

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileDownloader.kt
```diff
@@ -0,0 +1,146 @@
...
+        if (!cacheDir.exists()) {
+            cacheDir.mkdirs()
+        }
+        cache = cacheDir
+        return File(cacheDir, fileName)
+    }
+
+    @Suppress(" NestedBlockDepth", "MagicNumber")
+    private fun writeInputStreamToFile(inputStream: InputStream, file: File) {
+        inputStream.use { inputStream ->
```
khairul-alam-licon, line 103: We can write `inputStream.use { input ->` to avoid shadowness.

Screen Shot 2022-03-17 at 21 21 44

khairul-alam-licon (commented):
Looking good, besides I have left few suggestions.

khairul-alam-licon (approved):
Looks good 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileDownloader.kt
```diff
@@ -0,0 +1,176 @@
...
+
+    @VisibleForTesting
+    internal fun openShareIntent(mimetype: String?, file: File) {
+        val intent = Intent(Intent.ACTION_SEND)
+        intent.type = mimetype
+        intent.putExtra(
+            Intent.EXTRA_STREAM,
+            DefaultFileProvider(activity).getUriForFile(file)
+        )
+        activity.startActivity(Intent.createChooser(intent, null))
```
corycaywood, line 154: Sorry I think it was missing in the specs. This should use `ACTION_CREATE_DOCUMENT` instead of `ACTION_SEND`. Actually the `WebViewFileDownloader` should maybe be updated to use that as well.

This would actually simplify some things because you don't need to create a cache - you can just save directly to the file. However, this also means we need a way of receiving the Intent result from the Host App. Here's some example code:
```
internal fun openShareIntent(mimetype: String?, fileName: String) {
    val intent = Intent(Intent.ACTION_CREATE_DOCUMENT).apply {
        addCategory(Intent.CATEGORY_OPENABLE)
        type = mimetype
        putExtra(Intent.EXTRA_TITLE, fileName)
    }
    activity.startActivityForResult(intent, MINI_APP_FILE_DOWNLOAD_REQUEST_CODE)
}

fun onActivityResult(requestCode: Int, resultCode: Int, intent: Intent?) {
    if (requestCode == MINI_APP_FILE_DOWNLOAD_REQUEST_CODE) {
        intent?.data?.let { destinationUri ->
            try {
                // Download file...

                activity.contentResolver.openOutputStream(destinationUri)?.let { outputStream ->
                    outputStream.write(responseBody.bytes())
                    outputStream.close()
                } ?: Log.e("Downloader","Failed to download file: could not open OutputStream.")
            } catch (e: IllegalArgumentException) {
                // Error
            }
        }
    }
}
```

corycaywood (commented):
Sorry for a lot of comments - there were some things missing in the specs.

About the `ACTION_CREATE_DOCUMENT` type of Intent: we could keep the current implementation that you have, but this would mean that we will need to change this implementation in the Host App, causing double effort. What do you think?

I was thinking maybe you could create an interface similar to `MiniAppFileChooser` and `MiniAppFileChooserDefault`?

```
interface MiniAppDownloader {
    fun onStartFileDownload(fileName: String, url: String, headers: List<Pair<String, String>>)
}

class MiniAppDownloaderDefault : MiniAppDownloader {
    fun onReceivedResult(file: Uri, url: String,  headers: List<Pair<String, String>>) {
        // ...
    }
}
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/file/MiniAppFileDownloader.kt
```diff
@@ -0,0 +1,128 @@
...
+            startDownloading(destinationUri, client, request)
+        }
+    }
+
+    private fun startDownloading(destinationUri: Uri, client: OkHttpClient, request: Request) {
+        val response = client.newCall(request).execute()
+        if (response.isSuccessful) {
+            response.body?.let { responseBody ->
+                activity.contentResolver.openOutputStream(destinationUri)?.let { outputStream ->
+                    outputStream.write(responseBody.bytes())
```
corycaywood, line 68: Sorry, I used `ResponseBody.bytes` in the example, but I'm not sure if that's the most efficient way to handle large files. I think it can run into out of memory issues.

Maybe it's better to use an `InputStream`? Kotlin has a convenience function for this:
```
responseBody.byteStream().copyTo(outputStream)
```

corycaywood (commented):
Looking good 👍  I left a few more comments.

corycaywood (approved):
Looks good 👍

PR #435, opened 2022-04-12
[MINI-4840] Check whether Location Services is enabled and return error…

gopi-karmakar:
# Description
 Check whether Location Services is enabled and return error

## Links
MINI-4840

# Checklist

### DIFF
--- a/miniapp/src/main/AndroidManifest.xml
+++ b/miniapp/src/main/AndroidManifest.xml
@@ -4,10 +4,6 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
-    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
-
-    <!-- Include only if your app benefits from precise location access. -->
-    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
 
     <application>
         <provider
--- a/miniapp/src/main/AndroidManifest.xml
+++ b/miniapp/src/main/AndroidManifest.xml
@@ -4,6 +4,10 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+
+    <!-- Include only if your app benefits from precise location access. -->
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
 
     <application>
         <provider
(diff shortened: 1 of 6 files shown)


munir-rakuten (commented):
Looks good, left a suggestion though.

miniapp/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,10 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
+    <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION" />
+
+    <!-- Include only if your app benefits from precise location access. -->
+    <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
```
munir-rakuten, line 10: I think you can remove these permissions, we already have those permissions in host app and we generally try not to put permissions in the sdk unless it's unavoidable.

munir-rakuten (commented):
I think you can remove these permissions, we already have those permissions in host app and we generally try not to put permissions in the sdk unless it's unavoidable.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -250,6 +254,15 @@ open class MiniAppMessageBridge {
                 object : TypeToken<DevicePermission>() {}.type
             )
 
+            if (permissionParam.permission == MiniAppDevicePermissionType.LOCATION.name) {
```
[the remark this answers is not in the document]
gopi-karmakar: Actually, When I checked the JS-MiniApp code then I found that the `MiniAppMessageBridge.onRequestDevicePermission()` is the entry point to check for the permissions. So I thought before going any further it should first check whether the location service is enabled or not. Even if the location service is disabled after granting the permission it will still check for location enable case.

Yes, but if in case if it's surpass the `MiniAppMessageBridge.onRequestDevicePermission` and directly call the `WebChromeClient` methods then, in that case it won't check. 

So to address that case I've moved the checking at the `WebChromeClient.onGeolocationPermissionsShowPrompt` so it'll check for all the cases.

corycaywood (approved):
Looks good 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -44,6 +45,7 @@ open class MiniAppMessageBridge {
     private lateinit var downloadedManifestCache: DownloadedManifestCache
     private lateinit var miniAppId: String
     private lateinit var activity: Activity
+    private lateinit var locationManager: LocationManager
```
corycaywood, line 48: Looks like you can remove this one.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -2,13 +2,14 @@ package com.rakuten.tech.mobile.miniapp.js
 
 import android.app.Activity
 import android.content.Intent
+import android.location.LocationManager
```
khairul-alam-licon, line 5: Seems we don't need to import this one.

khairul-alam-licon (approved):
Looks good 👍 
Thank you!

munir-rakuten (approved):

PR #437, opened 2022-04-27
[MINI-4969] Secure Storage Implementation

munir-rakuten:
# Description
Implement Secure storage to mimic Local Storage with encryption.

## Links
MINI-4969

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,249 @@
...
+                } else {
+                    onFailed(MiniAppStorageError.failedDeleteError)
+                }
+            }
+        } else {
+            onFailed(MiniAppStorageError.unavailableStorage)
+        }
+    }
+
+    fun deleteSecureStorageItems(
```
corycaywood, line 145: I think you can simplify the method names in this class to remove `SecureStorage`. i.e. `deleteItems`, etc.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,249 @@
...
+    fun insertSecureStorageItem(
+        miniAppId: String,
+        items: Map<String, String>,
+        onSuccess: (Map<String, String>) -> Unit,
+        onFailed: (MiniAppStorageError) -> Unit
+    ) {
+        scope.launch {
+            storageState.postValue(StorageState.LOCK)
+            if (isStorageAvailable(miniAppId)) {
+                val storedItems = readFromEncryptedFile(miniAppId)
```
corycaywood, line 203: Could you use the already cached contents instead of reading the file again?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,249 @@
...
+                val byteArrayOutputStream = ByteArrayOutputStream()
+                var nextByte: Int = inputStream.read()
+                while (nextByte != -1) {
+                    byteArrayOutputStream.write(nextByte)
+                    nextByte = inputStream.read()
+                }
+
+                val plaintext: ByteArray = byteArrayOutputStream.toByteArray()
+                val jsonToRead = plaintext.toString(Charsets.UTF_8)
+                return deSerializeItems(jsonToRead)
```
corycaywood, line 116: Could you just use `InputStream.readyBytes()` instead? I think it internally handles creating a buffered outputstream.

```suggestion
                val inputStream = encryptedFile.openFileInput()
                val plaintext: ByteArray = inputStream.readBytes()
                val jsonToRead = plaintext.toString(Charsets.UTF_8)
                return deSerializeItems(jsonToRead)
```

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt
```diff
@@ -0,0 +1,169 @@
...
+        } else {
+            bridgeExecutor.postError(callbackId, ERR_WRONG_JSON_FORMAT)
+        }
+    }
+
+    fun onGetItem(callbackId: String, jsonStr: String) = whenReady {
+        val callbackObj: GetItemCallbackObj? =
+            Gson().fromJson(jsonStr, GetItemCallbackObj::class.java)
+        if (callbackObj != null) {
+            if (!cachedItems.isNullOrEmpty() && cachedItems!!.containsKey(callbackObj.param)) {
```
corycaywood, line 94: In the case that the key doesn't exist, I think we should just return `null` instead of an error. So maybe you can add a specific case for `!cachedItems!!.containsKey(callbackObj.param)`.

Also, maybe we should avoid using `!!` and use a null check instead?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,249 @@
...
+            storageState.postValue(StorageState.UNLOCK)
+        }
+    }
+
+    fun loadSecureStorage(
+        miniAppId: String,
+        onSuccess: (Map<String, String>) -> Unit,
+        onFailed: (MiniAppStorageError) -> Unit
+    ) {
+        scope.launch {
```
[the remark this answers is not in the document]
munir-rakuten: yes, this scope is `Dispatchers.IO`
```private var scope: CoroutineScope = CoroutineScope(Dispatchers.IO)```

File: miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt (hunk 4 above)
Comment (corycaywood) [reply miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt]: @munir-rakuten See this comment again - I think iOS updated the behavior to return `null` so Android needs to match.

Comment (munir-rakuten) [reply miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt]: @corycaywood addressed

File: miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt [hunk 6]
```diff
@@ -0,0 +1,257 @@
...
+
+    fun delete(
+        miniAppId: String,
+        onSuccess: () -> Unit,
+        onFailed: (MiniAppStorageError) -> Unit
+    ) {
+        if (isStorageAvailable(miniAppId)) {
+            scope.launch {
+                val file = File(secureStorageBasePath, "$miniAppId.txt")
+                file.delete()
```
corycaywood, line 126: Could deleting the file have any consequences if the mini app tries to use secure storage again later?
For example, probably when they call `MiniApp.clear()`, they'll then call `MiniApp.setItems()` soon afterwards to set new data.

Could we instead replace the file contents with an empty map - basically just re-initialize the file?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,249 @@
...
+                storedItems?.let { items ->
+                    val filterItems = items.filter { !keySet.contains(it.key) }
+                    writeToEncryptedFile(
+                        miniAppId,
+                        serializedItems(filterItems),
+                        onSuccess,
+                        onFailed
+                    )
+                } ?: kotlin.run {
+                    onFailed(MiniAppStorageError.emptyStorage)
```
corycaywood, line 157: I think we shouldn't return an error in this case. We can just return a normal success because the storage is already cleared.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,249 @@
...
+        onFailed: (MiniAppStorageError) -> Unit
+    ) {
+        scope.launch {
+            storageState.postValue(StorageState.LOCK)
+            if (isStorageAvailable(miniAppId)) {
+                val storedItems = readFromEncryptedFile(miniAppId)
+                storedItems?.let {
+                    onSuccess(storedItems)
+                } ?: kotlin.run {
+                    onFailed(MiniAppStorageError.emptyStorage)
```
corycaywood, line 178: What is the case where this error will happen? Is it when there is some error loading the file?
I think it should also return `unavailableStorage` in that case.

gopi-karmakar (approved):
Looks Good. 
Just a suggestion if we are not handling that. As per the Storage State if we could return the state to the host app so that if host app tries to access the storage then it could get the something like:  `Cache is still loading` or `Cache is ready` or `Cache is full` etc. and put conditions in the implementation as per the state that if it's not ready don't write or something.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Will be invoked by MiniApp.clearSecureStorage(miniAppId: String).
+     * @param miniAppId will be used to find the file to be deleted.
+     */
+    fun clearSecureStorage(miniAppId: String) {
+        if (isStorageAvailable(miniAppId)) {
+            scope.launch {
+                val file = File(secureStorageBasePath, "$miniAppId.txt")
+                file.delete()
```
corycaywood, line 241: It's possible that the Host App could call `clearSecureStorage` while the Mini App is still open. What would happen in this case?

corycaywood (approved):

khairul-alam-licon (approved):
Looks good 👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
munir-rakuten: @corycaywood we always create a new encrypted file when `miniapp` try to set new items, because encrypted files data can not be overwritten. if host app clear the storage while the `miniapp` is open, it will just simply create a new file and store the data.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
munir-rakuten: @corycaywood addressed.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
munir-rakuten: @corycaywood we always check the storage is available or not before performing any operation, so if it can not find the storage it just return an error or create a new one if `miniapp` wants to set a new item.

PR #439, opened 2022-05-09
[MINI-5008] Show null when get item failed After clear storage

munir-rakuten:
# Description
Able to fetch null value for key's post clear storage from inside miniapp.

## Links
MINI-5007, MINI-5008,  MINI-5009

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppAccessTokenError.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/errors/MiniAppAccessTokenError.kt
@@ -1,8 +1,11 @@
 package com.rakuten.tech.mobile.miniapp.errors
 
+import androidx.annotation.Keep
+
 /**
  * A class to provide the custom errors specific for access token.
  */
+@Keep
 class MiniAppAccessTokenError(val type: String? = null, val message: String? = null) :
     MiniAppBridgeError(type, message) {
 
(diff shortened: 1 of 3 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -233,13 +232,13 @@ internal class MiniAppSecureStorage(private val activity: Activity) {
                     if (storedItems.containsKey(key)) {
                         onSuccess(storedItems[key] ?: "")
                     } else {
-                        onFailed(MiniAppSecureStorageError.secureStorageIOError)
+                        onSuccess("null")
```
rleojoseph, line 235: Thanks for PR @munir-rakuten , I thinking to make changes in iOS to return `secureStorageIOError` if no key is found. @corycaywood your thoughts ?

corycaywood: @rleojoseph I think it should return `null` in this case. That matches how the browser's LocalStorage is working. 
The user will know that the key doesn't exist if `null` is returned, so I think there's no need to return an error.

khairul-alam-licon (approved):
Looks good 👍

gopi-karmakar (approved):

corycaywood (approved):

PR #440, opened 2022-05-10
[MINI-5008] Display null if item is not available

munir-rakuten:
# Description
Display null if item is not available.

## Links
MINI-5008

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
@@ -223,7 +223,6 @@ internal class MiniAppSecureStorage(private val activity: Activity) {
         miniAppId: String,
         key: String,
         onSuccess: (String) -> Unit,
-        onFailed: (MiniAppSecureStorageError) -> Unit
     ) {
         scope.launch {
             if (isStorageAvailable(miniAppId)) {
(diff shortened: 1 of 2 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -232,13 +231,13 @@ internal class MiniAppSecureStorage(private val activity: Activity) {
...
-                        onFailed(MiniAppSecureStorageError.secureStorageIOError)
+                        onSuccess("null")
                     }
                 } ?: kotlin.run {
-                    onFailed(MiniAppSecureStorageError.secureStorageIOError)
+                    onSuccess("null")
                 }
             } else {
-                onFailed(MiniAppSecureStorageError.secureStorageUnavailableError)
+                onSuccess("null")
```
rleojoseph, line 240: When storage is not available we should return an error. But remember whenever mini-app is opened secure storage will be available always unless the device storage is full already.

munir-rakuten: @rleojoseph  in our case thats the case we get into after clear storage, we delete the file, so when you try to get the item the file is not there so it will throw error and show the error currently showing.

corycaywood (approved):

khairul-alam-licon (approved):
Looks good 👍

PR #447, opened 2022-06-01
[MINI-5046] Close confirmation alert for Mini App

khairul-alam-licon:
# Description
A MiniApp can be developed to display a confirmation alert dialog when the users try to close the MiniApp. This PR aims to show the alert dialog inside HostApp using the JS interface from message bridge.

## Links
- MINI-5046

## Screenshots
Screen Shot 2022-06-02 at 13 50 18

# Checklist


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
```diff
@@ -278,6 +283,10 @@ class MiniAppDisplayActivity : BaseActivity() {
                     AppPermission.getDeviceRequestCode(miniAppPermissionType)
                 )
             }
+
+            override fun miniAppShouldClose(alertInfo: MiniAppCloseAlertInfo) {
+                closeAlertInfo = alertInfo
```
corycaywood, line 288: I wonder if it would be better for the Host App if we store the `MiniAppCloseAlertInfo` within the SDK, and then the Host App doesn't need to override any methods for this.

For example, when the Mini App sets the Close Alert info, we would store it in the SDK, and then we return it to the Host App when they call `MiniAppMessageBridge.miniAppShouldClose`. So the Host App would just do something like this:
```
 private fun checkCloseAlert() {
  val closeAlert = miniAppMessageBrdige.miniAppShouldClose
  if (closeAlert.shouldDisplay) {
    // ...
  }
}
```

@khairul-alam-licon @rleojoseph What do you think?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/closealert/MiniAppCloseAlertInfo.kt
```diff
@@ -0,0 +1,11 @@
+package com.rakuten.tech.mobile.miniapp.closealert
+
+import androidx.annotation.Keep
+
+/** An object to prepare information of close alert popup. */
+@Keep
+data class MiniAppCloseAlertInfo(
+    var shouldDisplay: Boolean = false,
+    var title: String? = "",
+    var description: String? = ""
```
corycaywood, line 10: I think these don't need to be nullable since you're setting a default, right? We can always set at least an empty string.

corycaywood (commented):
Looking good, but I left a couple of comments below

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/display/MiniAppDisplayActivity.kt
khairul-alam-licon: Yeah, I have changed the implementation accordingly by this commit: https://github.com/rakutentech/android-miniapp/pull/447/commits/7dd88ea425dfc97a218c1595ee18c4080de2abcf, please review when you'll have available time 🙇

rleojoseph (approved):
Looks good to me :+1: 
Could you please update the changelog?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -444,6 +450,19 @@ open class MiniAppMessageBridge {
         if (this::screenBridgeDispatcher.isInitialized)
             screenBridgeDispatcher.allowScreenOrientation = allowScreenOrientation
     }
+
+    @SuppressWarnings("SwallowedException")
+    private fun onSetCloseAlert(callbackId: String, jsonStr: String) = try {
```
munir-rakuten: Please check my other comment about changing the function name.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -444,6 +450,19 @@ open class MiniAppMessageBridge {
...
+        val callbackObj = Gson().fromJson(jsonStr, CloseAlertInfoCallbackObj::class.java)
+        val alertInfo = callbackObj.param.closeAlertInfo
+        onMiniAppShouldClose(alertInfo)
```
munir-rakuten, line 459: @khairul-alam-licon I think we can remove this `onMIniAppShouldClose` and set the value here and change the function name `onSetCloseAlert` to ``onMIniAppShouldClose`.

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
```diff
@@ -55,6 +57,9 @@ open class MiniAppMessageBridge {
     private var allowScreenOrientation = false
     private lateinit var miniAppSecureStorageDispatcher: MiniAppSecureStorageDispatcher
 
+    /** provide MiniAppCloseAlertInfo to HostApp to show close alert popup. */
+    var miniAppShouldClose: MiniAppCloseAlertInfo? = null
```
munir-rakuten, line 61: I think we can make this private and set a getter function for host app to get this value, so that host app can't manipulate the actual value comes from js. what do you think @khairul-alam-licon ?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppMessageBridge.kt
khairul-alam-licon: Yeah, I have updated the changes.

munir-rakuten (approved):
Looks good to me, thanks 👍

PR #477, opened 2022-09-12
[MINI-5417] Bulk delete should not show error message.

gopi-karmakar:
# Description
Describe the changes in this pull request.
Explain your rationale of technical decisions you made (unless discussed before).

## Links
Add links to github/jira issues, design documents and other relevant resources (e.g. previous discussions, platform/tool documentation etc.)

# Checklist
- [ ] I have read the [contributing guidelines](https://github.com/rakutentech/android-miniapp/blob/master/.github/CONTRIBUTING.md).
- [ ] I have completed the [SDK Development Learning Path](https://github.com/rakutentech/android-miniapp/blob/master/.github/CONTRIBUTING.md#sdk-development-learning-path) (if submitting a feature PR)
- [ ] I removed all sensitive data before every commit, including API endpoints and keys
- [ ] I ran `./gradlew check` without errors


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/database/MiniAppSecureDatabase.kt
```diff
@@ -435,7 +435,10 @@ internal class MiniAppSecureDatabase(
             throw e
         } finally {
             finalize()
+            if (!isDeleted && items.size > 1)
+                isDeleted = !isDeleted
         }
+
         return isDeleted
```
laubengaier, line 442: @gopi-karmakar why not just like below 🤔 ?

```
if (items.size > 1)
    isDeleted = true
```

gopi-karmakar: Yeah That is an obvious one but we will enter in to this condition only if the `isDeleted` is `false` so assigning with `!isDeleted`. It's something what I have noticed in the professional open source coding. Avoid hard coding as much as possible and rather taking the variables value in the runtime to be precisely sure that if the logic is working fine then the variable value should be what is expected.

munir-rakuten: @gopi-karmakar what is the reason of this check, `items.size > 0` might always be true right? as i can not see any remove item operation. can you please explain for which scenario is condition will happen?

so what i understood  `isDeleted` is true if it can delete an item, after the delete operation is done and `isDeleted` is still false may be for some exceptions, you just simply made it true and return it.it looks like even it fails to delete items it still returns true.

munir-rakuten: i understood after analyzing the code, both @laubengaier and @gopi-karmakar suggestion will work, but in this way there is one assertion less if `isDeleted` is already true, thanks.

munir-rakuten (approved):
Looks good to me.

laubengaier (approved):

PR #478, opened 2022-09-16
[MINI-5359] MiniApp Demo Revamp 

munir-rakuten:
# Description
 Demo App Revamp

- MiniApp Demo App will now use a TabBar instead of a single screen
- Adding List I and List II for displaying multiple MiniApps at the same time (configurable through Settings)
- Features will start with URL feature for testing with locally run MiniApps
- Settings is moved to it's own tab

#Walkthrough
https://rak.box.com/s/t0vyu05xr8mixe3ta3cva2ntmyfc0h6a

## Links
MINI-5359

# Checklist


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/RealMiniApp.kt
```diff
@@ -73,10 +75,37 @@ internal class RealMiniApp(
         }
     }
 
-    override fun clearSecureStorage() = secureStorageDispatcher.clearSecureStorage()
+    @Suppress("TooGenericExceptionCaught", "SwallowedException")
+    override fun clearSecureStorage(context: Context) {
```
laubengaier, line 79: I think this could be named `clearSecureStorage**s**` to differentiate with `clearSecureStorage(context: Context, miniAppId: String)`

munir-rakuten: @laubengaier addressed.

laubengaier (changes requested):
Good job looks great 👍 !

Think it's almost good to approve from my side although I've found one thing that needs to be fixed

MiniApp events (`NativeEventType`) should be extended with `appId` and `version` so the events will not be sent to all active MiniApps. (see for reference https://github.com/rakutentech/ios-miniapp/blob/master/Sources/Classes/core/JavascriptBridge/MiniAppJavaScriptEnumerations.swift#L126 https://github.com/rakutentech/ios-miniapp/blob/master/Sources/Classes/core/Display/RealMiniAppView.swift#L112). Can be reproduced with opening the event listener / external webview feature and checking between MiniApps they will receive multiple events not just their own.

Otherwise I could see some glitches with the second list not loading sometimes probably will be raised by QA as well. And another thing when opening `Features` the url feature should be titled with "URL" or "Local Testing" just to give an easy understanding of what it does.

laubengaier (approved):
👍

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt
```diff
@@ -43,7 +43,7 @@ internal class MiniAppSecureStorageDispatcher(
         this.miniAppSecureStorage = MiniAppSecureStorage(
             context,
             databaseVersion,
-            maxStorageSizeLimit
+            (maxStorageSizeLimit * CONVERT_TO_KB)
```
rleojoseph, line 46: Any reason why we are converting to KB here?

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -92,7 +92,8 @@ internal class MiniAppSecureStorage(
     }
 
     internal fun closeDatabase() {
-        miniAppSecureDatabase.closeDatabase()
+        if (this::miniAppSecureDatabase.isInitialized)
+            miniAppSecureDatabase.closeDatabase()
```
rleojoseph, line 96: 👍🏻

rleojoseph (approved):
Overall Looks good to me :+1: Thank you so much for the changes. I just have one comment.
Also, please update your changelog file about the Demo app changes

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt
munir-rakuten: @rleojoseph yes, in previous implementation this initialization has not been used, it uses this function [link](https://github.com/rakutentech/android-miniapp/blob/1429479cdefbfc04997c461018c40214202f14ce/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/js/MiniAppSecureStorageDispatcher.kt#L50-L52) so we didnt face any issue previously. But now we use the initial intialization so if we dont convert it in kb, then db will assume the value 5 MB as 5 kb which is very less and memory will be full very quickly.

PR #499, opened 2022-10-28
[MINI-5613] Set Default values settings

munir-rakuten:
# Description
Make `Production` mode default and set default values for

- profile name
- profile pic
- points

## Links
MINI-5613

# Checklist
- [x] I have read the [contributing guidelines](https://github.com/rakutentech/android-miniapp/blob/master/.github/CONTRIBUTING.md).
- [x] I have completed the [SDK Development Learning Path](https://github.com/rakutentech/android-miniapp/blob/master/.github/CONTRIBUTING.md#sdk-development-learning-path) (if submitting a feature PR)
- [x] I removed all sensitive data before every commit, including API endpoints and keys
- [x] I ran `./gradlew check` without errors

PR #502, opened 2022-10-31
[MINI-5665] [MINI-5522] Move preview mode based on list app

shohiebsense:
# Description
Environments and Preview Mode based on tabs. 
Also save the button in the enabled state after moving to tabs and back again, even if it's unchanged.

## Links
[MINI-5665](https://jira.rakuten-it.com/jira/browse/MINI-5665)
[MINI-5522](https://jira.rakuten-it.com/jira/browse/MINI-5522)

## Videos
https://rak.box.com/s/jy7y7uwzp7n6zg74kp3tarmo0k6k9jtw 
https://rak.box.com/s/jhgndfkuz27l8pmaw6ifzdojp3xwwrhk 

## PR Follow up Videos
https://rak.box.com/s/2neb57rgiz4ntimcipb3wjxaz7sx0ir6 (Obsolete)
https://rak.box.com/s/f0io5ussg7vwpwa9ubdr7hsoj9qhcbe1 (Obsolete)
https://rak.box.com/s/9oxzho09fwq7gji1ilvoxkie6qudct6n

# Checklist


testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
```diff
@@ -4,53 +4,30 @@ import android.content.Context
...
-internal class Cache(context: Context) {
+internal class Cache(context: Context, manifestConfig: AppManifestConfig) {
 
     private val gson = Gson()
     private val prefs: SharedPreferences = context.getSharedPreferences(
         "com.rakuten.tech.mobile.miniapp.sample.settings",
         Context.MODE_PRIVATE
     )
+    val productionBaseUrl = context.getString(R.string.prodBaseUrl)
+    val stagingBBaseUrl = context.getString(R.string.stagingBaseUrl)
```
rleojoseph, line 24:
```suggestion
    val stagingBaseUrl = context.getString(R.string.stagingBaseUrl)
```

testapp/src/main/res/values/strings.xml
```diff
@@ -117,4 +117,5 @@
     <string name="deeplink_activity_input_hint">Input here</string>
     <string name="deeplink_activity_no_deeplink">No deeplink available!</string>
     <string name="deeplink_activity_not_saved">No deeplink has been saved yet.</string>
+    <string name="miniapp_by_input">MiniApp By Input</string>
```
rleojoseph, line 120: Do we have this change done in iOS as well? @laubengaier

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/MiniAppCredentialCache.kt
```diff
@@ -0,0 +1,81 @@
+package com.rakuten.tech.mobile.testapp.ui.settings
+
+import android.content.SharedPreferences
+import android.content.SharedPreferences.Editor
+
+class MiniAppCredentialCache(
```
rleojoseph, line 6:
```suggestion
class MiniAppConfigCache(
```

rleojoseph (approved):
Good job. Overall looks ok to me, just a few suggestions. I leave it up to you.
Please update the Changelog file about the Sample app changes about this PR and the other one about the permission settings.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/MiniAppCredentialCache.kt
shohiebsense: Updated 🔔

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
shohiebsense: Updated 🔔

testapp/src/main/res/values/strings.xml
laubengaier: actually this should be called `URL` I think @munir-rakuten already updated the name? 🤔

shohiebsense: I see, I made an update, please check

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/input/MiniAppByUrlActivity.kt
```diff
@@ -25,9 +25,9 @@ class MiniAppInputActivity : BaseActivity(), PreloadMiniAppWindow.PreloadMiniApp
     private val preloadMiniAppWindow by lazy { PreloadMiniAppWindow(this, this) }
 
     sealed class InputDisplay(val input: String) {
-        class AppId(input: String): InputDisplay(input)
-        class Url(input: String): InputDisplay(input)
-        class None: InputDisplay("")
+        class AppId(input: String) : InputDisplay(input)
```
laubengaier, line 28: This one (speaking of functionality) should probably also be removed (maybe here's the confusion) the MiniAppByUrlActivity should be used for local testing with eg `http://localhost:1337` but not load a miniapp by it's id. The list search should already provide a way to input an id to find the MiniApp. Can you confirm this is the intention behind `InputDisplay.AppId`?

shohiebsense: Can confirm, the page is intended to have not by the **URL** but also the **appId**, to begin with. @laubengaier.

shohiebsense:
```
if (URLUtil.isValidUrl(input)) {
    onValidUI()
    InputDisplay.Url(input)
} else if (!AppSettings.instance.isDisplayUrlPreviewMode && !input.isInvalidUuid()) {
    onValidUI()
    InputDisplay.AppId(input)
}
```

which can be toggled in the settings page/

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
```diff
@@ -170,92 +149,174 @@ internal class Cache(context: Context) {
...
 
-        val defaultProdPair = Pair(
-            context.getString(R.string.prodProjectId),
-            context.getString(R.string.prodSubscriptionKey)
+        private val tab1TempMiniAppConfigCache = MiniAppConfigCache(
+            TEMP_IS_PROD_VERSION_ENABLED,
+            TEMP_IS_PREVIEW_MODE,
+            TEMP_REQUIRE_SIGNATURE_VERIFICATION,
+            TEMP_APP_ID,
+            TEMP_SUBSCRIPTION_KEY
```
laubengaier, line 176: Just a question regarding this is this an android specific implementation choice to use extra properties for the temporary config? couldn't this be implemented in a way such as `var listConfigI = MiniAppConfigCache(...)` to make it mutable? I feel this is bit of an unnecessary overhead.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/input/MiniAppByUrlActivity.kt
laubengaier: oh yeah then please remove it (also the settings option) we only need the url functionality here as you can see here
https://rak.box.com/s/noqj511lvpq29rgg5qouc0jjvp2dmush
(normally it should load the miniapp if it's running locally instead of the charles warning)

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
shohiebsense: 1. There needs a data to make it temporary when the user havent clicked the save button yet.
2. There needs a data to make it persistent when user switched to another tab and comeback. This one.
3. Should the save button functionality removed,  It wouldnt need these, agree.
4. When save button is clicked. It would clear these temporaries (both 1 and 2)

munir-rakuten (changes requested):
When the demo app settings page will open, the `Production` switch turned on and prod values will be the default, But i have checked this is not the case here.

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
laubengaier: ok but you could use a getter that uses a fallback to temporary data (from prefs) and otherwise use the actual data right? just as an example in iOS `Config.string(listType, key: .projectId, fallbackKey: .projectId) ?? ""` which basically falls back to the fallBackKey in case no persisted data exists

shohiebsense:
> When the demo app settings page will open, the `Production` switch turned on and prod values will be the default, But i have checked this is not the case here.

It would be contrary to this ticket https://jira.rakuten-it.com/jira/browse/MINI-5538

kindly think about these and confirm with the team

cc: @rleojoseph @laubengaier

munir-rakuten:
> > When the demo app settings page will open, the `Production` switch turned on and prod values will be the default, But i have checked this is not the case here.
> 
> It would be contrary to this ticket https://jira.rakuten-it.com/jira/browse/MINI-5538
> 
> kindly think about these and confirm with the team
> 
> cc: @rleojoseph @laubengaier



> > When the demo app settings page will open, the `Production` switch turned on and prod values will be the default, But i have checked this is not the case here.
> 
> It would be contrary to this ticket https://jira.rakuten-it.com/jira/browse/MINI-5538
> 
> kindly think about these and confirm with the team
> 
> cc: @rleojoseph @laubengaier

@shohiebsense i dont understand why it is contradictory? at the first run the production switch should be on as a default and it's already done in this [PR](https://github.com/rakutentech/android-miniapp/pull/499)

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
shohiebsense: As far as I know in Android the fallback should directly assign the primitive nullable value, except if it calls another key and acts as a fallback, which would be in a call loop since it ends to call primitive value as default. Also, this can also have the same concept as the fallback key...

Besides, the layout here is using the same view components for both tabs. each one view holds different states (previous saved data, modified unsaved data, modified saved data).

shohiebsense:
> > > When the demo app settings page will open, the `Production` switch turned on and prod values will be the default, But i have checked this is not the case here.
> > 
> > 
> > It would be contrary to this ticket https://jira.rakuten-it.com/jira/browse/MINI-5538
> > kindly think about these and confirm with the team
> > cc: @rleojoseph @laubengaier
> 
> > > When the demo app settings page will open, the `Production` switch turned on and prod values will be the default, But i have checked this is not the case here.
> > 
> > 
> > It would be contrary to this ticket https://jira.rakuten-it.com/jira/browse/MINI-5538
> > kindly think about these and confirm with the team
> > cc: @rleojoseph @laubengaier
> 
> @shohiebsense i dont understand why it is contradictory? at the first run the production switch should be on as a default and it's already done in this [PR](https://github.com/rakutentech/android-miniapp/pull/499)

You might think about the case when a user has already saved the configuration once, then come back then toggled the production mode. It would confuse between the default and the last saved data, and those kind of related experience issues.  

Nevertheless, let me try some workaround to get what you expect.

Production mode should be true by default regardless BuildConfig (debug and release), ok noted.

shohiebsense:
Updated. Check the Video on the Description (Follow-Up) @munir-rakuten

laubengaier:
> Updated. Check the Video on the Description (Follow-Up) @munir-rakuten

@shohiebsense @munir-rakuten I've just checked the video and the prod toggle did change without user interaction. Seems like the wrong behavior. Actual should be quite simple. Default values should be loaded when opening the settings page for the first time but any changes the user need to be maintained even when switching the tabs. 

See the video (first time launch)
https://rak.box.com/s/7b0sh6w4ui60lezvwo0s7z9mrgn3nzs3

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/Cache.kt
laubengaier: gotcha seems platform related

shohiebsense:
> g behavior. Actual should be quite simple. Default values should be loaded when opening the settings page for the first time but any changes the user need to be maintained even when switching the tabs.
> 
> See the video (first time launch) https://rak.box.com/s/7b0sh6w4ui60lezvwo0s7z9mrgn3nzs3

Thanks for the notice, It was from SDK API that it recreate whenever to come back to the tab, not resumed, thus the data displayed is wrong. The actual data is fine. Please wait.

shohiebsense:
Updated again, please check @laubengaier @munir-rakuten 

https://rak.box.com/s/9oxzho09fwq7gji1ilvoxkie6qudct6n

CHANGELOG.md
```diff
@@ -11,9 +11,11 @@
 
 **Sample App**
 - **Feature:** Demo App will now use a tabBar instead of a single screen.
-- **Feature:** Adding List I and List II for displaying multiple MiniApps at the same time and configurable through settings.
-- **Feature:** Features tab will start with URL feature to run MiniApps locally.
+- **Feature:** Added List I and List II for displaying multiple MiniApps at the same time and configurable through settings.
+- **Feature:** Features tab started with URL feature to run MiniApps locally.
 - **Feature:** Settings is moved to it's own tab.
+- **Feature:** Multiple MiniAppSdk configurations
+- **Feature:** Enabled PermissionSettings on MiniApp top bar
```
munir-rakuten, line 18: Please add a new section in the change log like this
```
### 5.1.0 (xxxx-xx-xx)
**Sample App**
- **Feature:** Added List I and List II for displaying multiple MiniApps at the same time and configurable through settings.
- **Feature:** Features tab started with URL feature to run MiniApps locally.
- **Feature:** Settings is moved to it's own tab.
- **Feature:** Multiple MiniAppSdk configurations
- **Feature:** Enabled PermissionSettings on MiniApp top bar
```

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/AppSettings.kt
```diff
@@ -9,72 +10,24 @@ import com.rakuten.tech.mobile.miniapp.js.userinfo.Contact
...
 import com.rakuten.tech.mobile.miniapp.testapp.BuildConfig
-import com.rakuten.tech.mobile.testapp.BuildVariant
 import java.util.*
 
+@Suppress("TooManyFunctions")
 class AppSettings private constructor(context: Context) {
 
     private val manifestConfig = AppManifestConfig(context)
-    private val cache = Cache(context)
+    private val cache = Cache(context, manifestConfig)
```
munir-rakuten, line 18: I think you dont need to pass `AppManifestConfig` here, you can set the values depending on settings page, `AppManifestConfig` is being used for sdk i guess,

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/MiniAppConfigCache.kt
```diff
@@ -0,0 +1,86 @@
...
+        editor.apply {
+            putBoolean(isproductionKey, configData.isProduction).commit()
+            putBoolean(
+                isSignatureVerificationRequiredKey,
+                configData.isVerificationRequired
+            ).commit()
+            putBoolean(isPreviewModeKey, configData.isPreviewMode).commit()
+            putString(projectidKey, configData.projectId).commit()
+            putString(subscriptionKey, configData.subscriptionId).commit()
+        }
```
munir-rakuten, line 73: you can create a json string from that object and save that in prefs instead of save individually.

CHANGELOG.md
shohiebsense: Updated, how now? @munir-rakuten

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/MiniAppConfigCache.kt
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    fun clear(editor: Editor) {
+        editor.apply {
+            remove(isproductionKey).commit()
+            remove(isPreviewModeKey).commit()
+            remove(isSignatureVerificationRequiredKey).commit()
+            remove(projectidKey).commit()
+            remove(subscriptionKey).commit()
```
munir-rakuten, line 82: check the previous comment, you can just remove one prefs value.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## CHANGELOG
 
+### 5.1.0 (2022-11-02)
+
```
munir-rakuten, line 4: remove this space

munir-rakuten: Alos we dont have a release date yet so please change it to  `### 5.1.0 (xxxx-xx-xx)`

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/MiniAppConfigCache.kt
shohiebsense: I followed the previous approach.... ok then

testapp/src/main/java/com/rakuten/tech/mobile/testapp/ui/settings/AppSettings.kt
shohiebsense: AppManifestConfig itself along with its AppSettings runs on the runtime even previously. 

Alright.

shohiebsense:
Updated using json format cache @munir-rakuten

munir-rakuten (approved):
Looks good to me, thanks 👍

shohiebsense:
Thank you

PR #518, opened 2022-11-11
[MINI-5737] Error on clear storage data once the MiniApp is closed and reopened.

gopi-karmakar:
# Description
It was a side effect of [MINI-5711](https://jira.rakuten-it.com/jira/browse/MINI-5711) and was a genuine issue. There are many cases to DB need to take care and since the DB load time was changed to insert time only from [MINI-5711](https://jira.rakuten-it.com/jira/browse/MINI-5711) so due to the previous flow DB was loaded and opened only once during MiniApp Launch time but now it's loaded and opened only during insert time, So because if that the DB was unavailable for the other functionalities. However I already found the issue and fixed it. Now the Db will be loaded and opened for other functionalities if the DB is available for a MiniApp in the device. Tested the fix with many various scenarios and it seems to be working fine now as expected.

## Links
[MINI-5711](https://jira.rakuten-it.com/jira/browse/MINI-5711)

# Checklist

### DIFF
--- a/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
+++ b/miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
@@ -59,17 +59,26 @@ internal class MiniAppSecureStorage(
         return status
     }
 
-    private fun validatePreCheck(onFailed: (MiniAppSecureStorageError) -> Unit): Boolean {
-        val status: Boolean = if (!miniAppSecureDatabase.isDatabaseAvailable(databaseName)) {
+    private fun isPreCheckPasses(onFailed: (MiniAppSecureStorageError) -> Unit): Boolean {
+        return if (!miniAppSecureDatabase.isDatabaseAvailable(databaseName)) {
             onFailed(MiniAppSecureStorageError.secureStorageUnavailableError)
             false
+        } else if (!miniAppSecureDatabase.isDatabaseReady()) {
+            createOrOpenAndUnlockDatabase()
+            true
         } else if (miniAppSecureDatabase.isDatabaseBusy()) {
             onFailed(MiniAppSecureStorageError.secureStorageBusyError)
             false
         } else {
             true
         }
-        return status
+    }
+
+    @VisibleForTesting
+    internal fun loadDatabase() {
+        if (!miniAppSecureDatabase.isDatabaseReady()) {
+            createOrOpenAndUnlockDatabase()
+        }
     }
 
     @Suppress("SwallowedException", "TooGenericExceptionCaught")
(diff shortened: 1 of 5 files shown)


miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
```diff
@@ -59,17 +59,26 @@ internal class MiniAppSecureStorage(
         return status
     }
 
-    private fun validatePreCheck(onFailed: (MiniAppSecureStorageError) -> Unit): Boolean {
-        val status: Boolean = if (!miniAppSecureDatabase.isDatabaseAvailable(databaseName)) {
+    private fun isPreCheckPasses(onFailed: (MiniAppSecureStorageError) -> Unit): Boolean {
```
shohiebsense, line 62:
```suggestion
    private fun isPreCheckPassed(onFailed: (MiniAppSecureStorageError) -> Unit): Boolean {
```

rleojoseph (changes requested):
Overall looks fine to me. Please update the description and links for the ticket. Also, update the Changelog to mention the Hotfix changes version and details

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
Climbatize: we are not judging grammar here @shohiebsense 🤣 , the `is` is a convention to add as a prefix to the boolean method `preCheckPasses`. But honestly I'd rather keep the original name. if you want to go this way, you can use `isPrecheckValid` or `hasPrecheckPassed`

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/database/MiniAppSecureDatabase.kt
```diff
@@ -197,6 +197,8 @@ internal class MiniAppSecureDatabase(
         }
     }
 
+    override fun  isDatabaseReady(): Boolean = this::database.isInitialized
+
     override fun isDatabaseOpen(): Boolean = database.isOpen
```
Climbatize, line 202: why not having readonly boolean vars instead of methods?

Climbatize (approved):
LGTM, some trivial comments

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/database/MiniAppSecureDatabase.kt
Climbatize: nevermind, I just noticed the override

rleojoseph (approved):
Looks good to me :+1:

miniapp/src/main/java/com/rakuten/tech/mobile/miniapp/storage/MiniAppSecureStorage.kt
shohiebsense: haha was not being grammar police at all sorry! just a bit strange, agree, it's the preference after all, hence the suggestion, thanks for the thoughts!

